PR #57, opened 2012-10-11
Give values for key attributes v2.

haeal:
Fixed up the line endings problem (hopefully).
Also added another bugfix for postgres using the Get<T> extension.


NickCraver:
Obviously *just a little* late reviewing this (sorry!), but I'm thinking this may not be effectual after the v3.0 driver changes. I wouldn't merge this as-is because of the breaking change on `Insert<T>` though - that needs to be a separate discussion and PR.

NickCraver:
This is now being implemented via #352, closing this one out. Thanks for the push to look at this and the work!

PR #65, opened 2012-12-03
Simple Composite Entity Mapping Support

jdaigle:
I've added support for Composite Entity Mapping. For reference, look at my test named TestCompositeEntity().

The idea is that to map the column names onto a simple entity graph. Using quoted column name identifiers (such as "Category.Name"), the mapper will get a property named "Category" and create it if necessary using the type's default ctor. It then maps the column onto the "Name" property of the "Category" instance.

This graph can be nested.

This was implemented primarily by modifying DefaultTypeMap and SimpleMemberMap to include the nested property graph detection code. A new property was added to IMemberMap to expose the nested property graph to IL generating method.

A new subroutine was added during IL generation to handle the nested property graph (recursively creates the necessary properties and pushes it onto the stack to the receive the value).

It references this: http://code.google.com/p/dapper-dot-net/issues/detail?id=126

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -2705,28 +2705,31 @@ this IDbConnection cnn, string sql, Func<TFirst, TSecond, TThird, TFourth, TRetu
                 var count = 0;
                 bool isString = value is IEnumerable<string>;
                 bool isDbString = value is IEnumerable<DbString>;
-                foreach (var item in list)
+                if (list != null)
                 {
-                    count++;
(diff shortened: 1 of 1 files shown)


adaam2:
I know that this was a while ago, but is there way to acheive this using the most up to date version of Dapper? I am querying a single table, however in my POCOs they follow a nested (composite) structure with some columns in the original table being represented by child classes (e.g. a single column "LanguageISO" is represented by the "Language" class)

Had a quick look around and couldn't find anything obvious relating to this and didn't know whether the mapping functionality was the right thing to be looking at.

PR #70, opened 2012-12-15
Support a default value for DbString.IsAnsi

kppullin:
It is painful and easy to forget to set `DbString.IsAnsi = true` when dealing with databases that mainly use char/varchar data types (as opposed to nchar/nvarchar).

This patch allows a default value for `DbString.IsAnsi` to be configured via the `DefaultStringDbType` static property.


mgravell:
Merged in essence; I tweaked the implementation a bit, so it is just a bool flag `IsAnsiDefault`, to match the `IsAnsi` (rather than exposing the `DbType`) - but same concept; thanks.

PR #75, opened 2013-02-03
Table reflection does not work for table with Id that is not int

ctstone:
The problem originated from trying to use Activator on typeof(Table<>) instead of typeof(Table<,>) e.g. for Table<Product, Guid>.

However, since table construction is handled the same for all tables in a database, I pulled the Table Id generic typing from the Table class and put it into the Database class. The idea is to specify Id typing once in the Database instead of in each Table.

I also added a Database implementation that assumes Id=typeof(int) to prevent breaking existing usage. I don't think Table was working with non-int Ids before this, so these changes should not negatively affect that use case.

Maybe the ideal situation would be to examine each Table property in the Database, and use Activator directly on its GetType().GetGenericTypeDefinition(). Though this might be a bit of an overhaul and might affect performance.

DapperTests all passed, and I successfully tested Tables of type int and Guid. Hope this is helpful.

### DIFF
--- a/Dapper.Rainbow/Database.cs
+++ b/Dapper.Rainbow/Database.cs
@@ -19,18 +19,28 @@ using System.Reflection.Emit;
 namespace Dapper
 {
     /// <summary>
+    /// A container for a database, assumes all the tables have an Id column of type int named Id
+    /// </summary>
+    /// <typeparam name="TDatabase">The Dapper.Database implementation type</typeparam>
+    public abstract class Database<TDatabase> : Database<TDatabase, int>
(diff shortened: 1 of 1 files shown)


NickCraver:
I'm not sure this makes sense, because it contains the assumption that *all* tables in the database have the same key type - or am I mis-reading things here? In a data warehouse for example, I would have `int` (or `tinyint`) columns for most IDs and `long` ID columns on larger tables. 

I'm all for breaking the `int` assumption (still true in current code) - I think we need a different approach though. It could be attribute-based as well.

PR #84, opened 2013-06-18
For Query method, adding overload which takes Type as a regular input parameter and return IEnumerable<object>

SolutionYogi:
In my current project, we are using Dapper and are extremely happy with it.

However, in our system, we create the database tables and types dynamically based on user input. 

For us, Query<T> method doesn't work because we don't know the T at compile time.

We wanted to have a Query method which takes in a type and returns ```IEnumerable<object>```. 

It would be really great if this could be merged in to mainline.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -610,6 +610,50 @@ namespace Dapper
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


NickCraver:
It looks like this overload pattern was added in 1a41f12c - @SolutionYogi, @mgravell can we close this out?

PR #87, opened 2013-06-24
Added optional property ignore during insert.

AdaptEngine:
Regarding pull request #66.  I agree that Id should not always be removed.  This will allow for an optional id column, as well as any other column for that matter.  It ignores the cache, if there exists some prop to ignore.

### DIFF
--- a/Dapper.Rainbow/Database.cs
+++ b/Dapper.Rainbow/Database.cs
@@ -50,11 +50,24 @@ namespace Dapper
             /// </summary>
             /// <param name="data">Either DynamicParameters or an anonymous type or concrete type</param>
             /// <returns></returns>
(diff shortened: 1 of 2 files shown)


NickCraver:
There are a few issues here: Each insert does reflection to duplicate the exact work that `GetParamNames()` is already doing (and caching). This is prohibitively expensive, especially in a loop. The interface on `.Insert()` is changed in a matter than permanently reduces the ability to expand it (a single collection parameter would be far better than `params` here).

My inclination honestly would be to close this PR since an applicable PR would bear little resemblance to this one. I'll wait a few days before doing anything though, to see if you want to fix things up @AdaptEngine. Note: it's unlikely this would be merged before a 2.0 in either case, due to the interface changes.

PR #101, opened 2013-10-30
Adding TVP (Table Valued Parameter) support.

nliao:
We downloaded your repository a few months ago, but found out that TVP was not supported, so I implemented this simple fix and it works pretty well in the enterprise application I'm participating in. Can you take a look and see if it is a reasonable fix, and maybe merge to your repository?


super-rain:
good

PR #110, opened 2014-01-14
Simple Func to allow provision of TypeMaps automatically

garyhuntddn:
Provision of a TypeMapProvider property to allow lazy creation of type maps only when they are needed

The Func simply returns a new DefaultTypeMap instance by default but you can replace the Func with something more useful - for example I am using a custom class that implements ITypeMap that uses the DataAnnotations attributes for column names.


NickCraver:
+1

NickCraver:
Note: need to address the IL issues.

NickCraver:
Thanks!

PR #113, opened 2014-02-13
Added ExecuteWithOutput method

arootbeer:
I ran into a situation recently where I need to retrieve more than just the ID of the new row that was created, and I wanted to add code that would behave the same as the existing `Execute()` method when provided with an `IEnumerable` parameter.

I noted that the `TestSplitWithMissingMembers()` test is failing, but I did not touch that code, so I assume it's either something specific to my environment or would also be failing in master.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -729,6 +729,56 @@ this IDbConnection cnn, string sql, dynamic param = null, IDbTransaction transac
             }
             return ExecuteCommand(cnn, transaction, sql, (object)param == null ? null : info.ParamReader, (object)param, commandTimeout, commandType);
         }
+
+        /// <summary>
+        /// Execute parameterized SQL and return selected or output records
+        /// </summary>
+        /// <returns>Selected or Output records</returns>
+        public static IEnumerable<T> ExecuteWithOutput<T>(
+#if CSHARP30
+this IDbConnection cnn, string sql, object param, IDbTransaction transaction, int? commandTimeout, CommandType? commandType
+#else
+this IDbConnection cnn, string sql, dynamic param = null, IDbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null
+#endif
+)
+        {
+            var multiExec = (object)param as IEnumerable;
+            if (multiExec == null || multiExec is string)
+            {
+                multiExec = new List<object>{param};
+            }
+            Identity identity = null;
+            CacheInfo info = null;
+            var results = new List<T>();
+
+            using (var cmd = SetupCommand(cnn, transaction, sql, null, null, commandTimeout, commandType))
+            {
+                string masterSql = null;
+                var isFirst = true;
+                foreach (var obj in multiExec)
+                {
+                    if (isFirst)
+                    {
+                        masterSql = cmd.CommandText;
+                        isFirst = false;
+                        identity = new Identity(sql, cmd.CommandType, cnn, null, obj.GetType(), null);
+                        info = GetCacheInfo(identity);
+                    }
+                    else
+                    {
+                        cmd.CommandText = masterSql; // because we do magic replaces on "in" etc
+                        cmd.Parameters.Clear(); // current code is Add-tastic
+                    }
+                    info.ParamReader(cmd, obj);
+                    using (var reader = cmd.ExecuteReader())
+                    {
+                        results.AddRange(ExecuteReader<T>(reader, identity, info));
+                    }
+                }
+            }
+            return results;
+        }
+
 #if !CSHARP30
         /// <summary>
         /// Return a list of dynamic objects, reader is closed after the call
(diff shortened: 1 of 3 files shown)


mgravell:
Sorry for extended delay; my bad - conflicting priorities etc. I am a bit against this at a philosophical level, simply for the reason that it confuses the expected terminology a bit. `Execute*` is something that doesn't have an output; if it has results in the way described, that *is* a `Query*` - which is underlined by the fact that the method you've implemented is basically `Query<T>` - which is shown readily by the fact that all of the examples added (with the exception of `TestExecuteWithOutputOnTypedIEnumerable`) would work *today* with `Query<T>`. Perhaps the real feature we should be pursuing here is "make `Query<T>` support `IEnumerable*` parameters by unrolling them". There's a *slight* complication there in that this could *technically* be a breaking change to consumers - that is something we could work on. But if we added a new API, it kinda feels like it would be a new `Query*` method, such as `QueryUnrollMultipleNamingIsHard<T>(string cmd, IEnumerable<T> args)`.

I suspect you're coming at the name from a "command / query separation" kinda perspective: "execute is command, query === query", but ultimately the database doesn't care about such distinctions, and I'm not sure we need a method called `Execute*` that looks like `Query*` just because the caller is thinking "this is a mutation command".

So: long story short - I'm going to decline it for the reasons above, but I think there *is* probably something we can do here to support `Query<T>` around a loop. I just don't think this is the way to do it.

arootbeer:
I agree with you on the naming: `ExecuteWithOutput` is pretty much self-contradictory in ADO.Net parlance. But I'd counter that `QueryUnrollMultipleNamingIsHard` is possibly less _explanatory_ :P

To be clear, the actual intent of my usage is to upsert multiple objects, which after any transformations performed during the upsert will continue to be used as part of a larger workflow, *while only sending a single command to the database*. AFAIK that's not possible (although obviously it may have been added in the intervening time period here).

mgravell:
You can do that with TVPs, which have direct support (and I'm going to
merge a second way to do it shortly). The only problem then is that you
need to define the custom type in the server, but that's easy enough.

mgravell:
Totally agree that QueryUnrollMultipleNamingIsHard is a terrible name :)

arootbeer:
The data structures are already shaped like the table in question, so it seems that adding a TVP just introduces a third layer that has to be kept in sync with changes to the table and the data structure. Worse, if nullable columns are added to the underlying table and the TVP isn't kept in sync, there would be silent corruption of the incoming data for those columns.

(I have only a passing familiarity with TVPs - I know what they are and could put one together - so if that second point badly misrepresents how they work in practice please let me know.)

mgravell:
I totally agree with you, but a TVP is the only *simple* way of doing a
complex multi-record upsert as a single command. There is another way that
would involve a single round trip and command cloning inside that, but that
has a range of complicating factors.

arootbeer:
> a TVP is the only *simple* way of doing a complex multi-record upsert as a single command

Do you consider the code in this PR to be a non-simple way of performing that task? I'm asking honestly, not trying to be snarky. I felt that your original rejection was based more on the cognitive dissonance introduced by the name of the method, which is something I can think about and attempt to remedy on my own based on our discussion so far, and less on the actual implementation, which is something that would take (probably much) more discussion. If that's not the case (i.e. if the code is also problematic), and you'd prefer that I just create an issue to request similar non-TVP functionality, I can do that as well.

mgravell:
No, I'm saying the method in this PR doesn't achieve that - it is multiple
found trips. There are two ways of doing it in one round trip: TVPs (the
simple way), or command cloning - by the latter I mean:

insert Table(col) values (@p);

Becomes (for 3 rows):

insert Table(col) values (@p0);
insert Table(col) values (@p1);
insert Table(col) values (@p2);

That is non-trivial in the general case, for many reasons. The naming is
important, but even with the rename, as far as I can see this could
actually be implemented simply by using Query in a loop. A complex question
is whether this should be made to work for the existing Query method.

Does that help?

arootbeer:
Geez...I'm not even reading my own code correctly :-1: When I re-read it yesterday, I thought it was essentially doing the command cloning. I can certainly see how that would be non-trivial in the general case. 

We've discussed previously in the comments that the general goal can be accomplished quite simply using the existing `Query` method. But to return to the question of naming, just as `Execute` implies no output, `Query` implies no side effects, which is obviously not _my_ intent with this code. As such, I'd suggest that it would be wise to find a more appropriate use case before heading down that path.

mgravell:
Purism aside, I'm just not sure that it is useful to try to enforce that
distinction. SQL does whatever the SQL says to do. Perhaps Query isn't the
perfect name, but it is what it is. Ultimately I'm just saying "that is
what the method called Foo does; we don't need Bar to do the same thing".
Naming is hard.

arootbeer:
I believe we're vehemently agreeing with each other at this point :+1:

PR #122, opened 2014-06-02
FIX: Properties of type Table<T, TId> are not initialized

MartinDevillers:
Table properties that implement ```Table<T,TId>``` instead of ```Table<T>``` are not recognized during the initialization of the ```Database<T>``` class. As a result, these properties remain uninitialized after the call to ```Database<T>.Init()```.

```
// Works fine
public Table<Person> Persons { get; set; }

// Remains 'null' after Init
public Table<Car, Guid> Cars { get; set; }
```

### DIFF
--- a/Dapper.Rainbow/Database.cs
+++ b/Dapper.Rainbow/Database.cs
@@ -177,7 +177,7 @@ namespace Dapper
 
         internal virtual Action<TDatabase> CreateTableConstructorForTable()
         {
-            return CreateTableConstructor(typeof(Table<>));
+            return CreateTableConstructor(typeof(Table<>), typeof(Table<,>));
         }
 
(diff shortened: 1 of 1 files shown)


NickCraver:
This is the simple version I believe is safe and we'd need to close out #184 on merge.

NickCraver:
This is now pulled in via https://github.com/StackExchange/dapper-dot-net/commit/1e8011ce916d610ec4623c8b0bf37a2491f59bee with a slight tweak to avoid breaks in https://github.com/StackExchange/dapper-dot-net/commit/dfe90c40c46c43e7d4e9902520ed21aad3777ac4. I'm not sure why it didn't auto-close this, but your fixes are in, thanks!

PR #132, opened 2014-07-19
Allow Dapper users to specify custom default type maps

ragnarok111:
Edit: this pull request originated as a way to allow SNAKE_CASE in column names but has evolved into a way to allow users to specify custom default mappers.

This allows DefaultTypeMap to search for columns that use "_" to separate words as an alternative to camel case. Some older database systems, such as DB2 for iSeries only return columns in all caps. This make camel case difficult to deal with during development.


ragnarok111:
@mgravell, if I rewrite this for this current version of Dapper, would there be a chance it would get merged?

NickCraver:
This is going in, but via a simpler implementation in #110. I really do appreciate the commit, the public/virtual bits especially are something we need to discuss. I think exposing the `DefaultTypeMap` as a base may be something that's reusable. However, the IL bits still need addressing, and I left a note on this in the other PR - we'll take a look before merging in and introducing unexpected behavior.

PR #172, opened 2014-09-17
Added support for using template with WHERE already in it

programcsharp:
Many times, I have a template with a lot of WHERE already there, and want to add stuff dynamically.

This adds a flag so you can specify that the template already has a WHERE clause, so the dynamic where's are connected with AND instead of WHERE.


NickCraver:
I think rather than a breaking change here we probably want to go with the direction that @tms laid out - it's a good suggestion about how to do this very simply and contained within the template. This allows someone to modify the query pretty autonomously without worrying about switches to methods (that we can otherwise avoid) downstream. I'll leave this open until I have an alternative in place - we're working on some code maintenance first though.

PR #174, opened 2014-09-17
Add mysql/mariadb sql adapter.

majorsilence:
Mysql/mariadb support in dapper.contrib.

### DIFF
--- a/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
+++ b/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
@@ -395,3 +395,20 @@ public partial class SQLiteAdapter
         return id;
     }
 }
+
+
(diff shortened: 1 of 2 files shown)


NickCraver:
I think this one's pretty straight-forward, we'll take a look this week. @majorsilence can you please fixing the spelling of Quote in the params? I have no doubt that intellisense helped replicate the hell out of that one :)

NickCraver:
We want to pull this in, but a little different direction on the actual implementation bits. I'm going to close this out but don't want to throw away the work. We have to pick and choose merge orders with the current code setup before doing a split - so we're going to go with #352 which include this and other changes but the quoting and such a bit more flexible. There are additional changes needed to #352 but it has other stuff we want to merge on in.

I really do appreciate the contrib, and sorry the *huge* file creates problems like this with competing PR merge conflicts...we're working to address that right now.

PR #177, opened 2014-09-22
Always use TypeHandler for parameters if available

matwilko:
Previously TypeHandlers could override the logic for built-in SQL data
types (int, byte, DateTime, etc.) when they were return values via
ITypeHandler.Parse, but ITypeHandler.SetValue was never called for these
types when used as parameters. Moved the typeHandlers.TryGetValue() call
up to make sure a type handler is always used if available.

Added tests to confirm standard behaviour is still working, and that ITypeHandler.SetValue is now called for built-in types.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -849,7 +849,7 @@ namespace Dapper
         internal static DbType LookupDbType(Type type, string name, out ITypeHandler handler)
         {
             DbType dbType;
-            handler = null;
+            typeHandlers.TryGetValue(type, out handler);
             var nullUnderlyingType = Nullable.GetUnderlyingType(type);
             if (nullUnderlyingType != null) type = nullUnderlyingType;
             if (type.IsEnum && !typeMap.ContainsKey(type))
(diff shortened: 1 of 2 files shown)


kartofeell:
I have applied changes from this PR on the latest master, and it works like a charm for me. TypeHandler.Parse and TypeHandler.Set

kartofeell:
It would be great if this fix could get in. I guess there is not many conflicts to resolve, right?

matwilko:
I would imagine not if you were able to simply apply the changes to the current master, however even then it is still a matter of Marc's time, since he needs to verify the PR before merging.
Functionality within Dapper may well have changed around this in the meantime though, it has been 11 months after all!

NickCraver:
Do you mind rebasing this to current master so we can take a proper look and merge it in?

matwilko:
Sure, I'll have a look this evening.

NickCraver:
It looks like we can't actually pull this in due to comments mentioned on #339. It'll require some more digging to see if we can do this in a workable way - punting for the moment while we do some maintenance to make all future PRs easier though.

nil4:
In https://github.com/StackExchange/dapper-dot-net/pull/339#issuecomment-148461888:

> Since we can't merge either - I'll consolidate this into #177 so we can track the issue in one place.

However, both issues are now closed despite not being resolved; @NickCraver consider reopening.

PR #184, opened 2014-09-30
Properties of type Table<TDatabase, TId> results in a null value after Database.Init

ryanrodemoyer:
For Dapper.Rainbow, declaring a property as Table<TDatabase, TId> will result in a null value for the property after calling Database.Init because the method to generate the constructor does not account for the type Table<,>. This pull request fixes that problem.

The workaround for this bug would be to subclass Table like "StringTable : Table<TDatabase, string>" and then declare the properties as StringTable versus Table<TDatabase, string>.

### DIFF
--- a/Dapper.Rainbow/Database.cs
+++ b/Dapper.Rainbow/Database.cs
@@ -177,7 +177,7 @@ namespace Dapper
 
         internal virtual Action<TDatabase> CreateTableConstructorForTable()
         {
-            return CreateTableConstructor(typeof(Table<>));
+            return CreateTableConstructor(typeof (Table<>), typeof (Table<,>));
         }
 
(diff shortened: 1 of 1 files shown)


NickCraver:
To prevent a breaking change here - we may need to restructure the call a bit. I would do `CreateTableConstructor(Type tableType)` which calls another multi-method underneath (the constructor could call this directly). This way all inheritors do not break without a recomplication.

This *may* not be needed though, @mgravell do you remember the resolution rules without re-compile if we just leave the `params` version in place on the current single arg? If that's 100% safe, I'd remove the `List` intermediate altogether.

NickCraver:
Fixed via https://github.com/StackExchange/dapper-dot-net/commit/1e8011ce916d610ec4623c8b0bf37a2491f59bee, thanks!

PR #188, opened 2014-10-12
added a option whether packing list parameters round brackets.

nabehiro:
When I develop database query with Dapper,
at first I write SQL statement on SSMS(SQL Server Management Stduio) 
ane next I add SQL statement to C# code.

But In clause  indication of Dapper " where Id in @Ids" is syntax error on SSMS.
I want to write  " where Id in ( @Ids )" .


NickCraver:
I'm with Marc on this one - I use SSMS all the time back and forth. For me, this would be a net-worse user experience. You're either using the expanded version (which already works), or you have to modify the current one for multiple values *anyway*. If multiple values wasn't the common case, it wouldn't be an  `IN` clause - I think very often it is, and that has held true across a wide range of projects for me.

PR #194, opened 2014-10-22
Adding support for TVP using List<SqlDataRecord>

sqmgh:
Since it's measurably faster to create a List than it is a one column DataTable, support for this method of using TVPs should be included as well.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -715,7 +715,7 @@ namespace Dapper
             typeMap[typeof(object)] = DbType.Object;
 
             AddTypeHandlerImpl(typeof(DataTable), new DataTableHandler(), false);
-            AddTypeHandlerImpl(typeof(List<Microsoft.SqlServer.Server.SqlDataRecord>), new SqlDataRecordHandler(), false);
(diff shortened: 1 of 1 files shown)


Dapper NET40/SqlMapper.cs
```diff
@@ -4872,6 +4885,63 @@ public void SetValue(IDbDataParameter parameter, object value)
...
+        private readonly string typeName;
+        /// <summary>
+        /// Create a new instance of SqlDataRecordListTVPParameter
+        /// </summary>
+        public SqlDataRecordListTVPParameter(IEnumerable<Microsoft.SqlServer.Server.SqlDataRecord> data, string typeName)
+        {
+            this.data = data;
+            this.typeName = typeName;
+        }
+        static readonly Action<System.Data.SqlClient.SqlParameter, string> setTypeName;
```
rcollette: @sqmgh  - I'm curious what you mean by "stock 'in' support".

sqmgh: Being able to write 'WHERE foo IN @Params' into your query, and having dapper expand that into a giant list of parameters. With very large parameter lists it's quite slow, and puts a noticeable CPU load on SQL Server parsing it all.

mgravell:
Much appreciated; sorry for the delay

PR #226, opened 2014-12-24
Cancel the command in unbuffered QueryAsync

mburbea:
This is an optimization for SQL Server and potentially other providers.
Currently, if you Dispose an IEnumerable produced from QueryAsync .NET
will still run the SqlCommand through to calculate statistics and assign
out params.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -4327,6 +4327,10 @@ Type type, IDataReader reader, int startBound = 0, int length = -1, bool returnN
             }
             private void NextResult()
             {
+                // this seems insane but its to allow cancelling out of deferred executions.
+                // if you try to read again you'll get the object disposed exception.
+                // however, it seems reasonable to want to dispose of the enumerator.
+                if (reader == null) return;
                 if (reader.NextResult())
                 {
                     readCount++;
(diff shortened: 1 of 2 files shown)


NickCraver:
I'm not sure this is a common use case enough to warrant the change. I argue against this one specifically because of #292 which will likely change up the implementation here and allow for much more foot-shooting as a result if we supported cancellation in this manner.

I think a better approach here is probably on the async path, via cancellation tokens and a few more checks possibly added in the pipe.

mburbea:
I found myself needing this because I was working on a database project that used stored procs with multiple (unwanted) record sets. It's very annoying to have to avoid disposing the gridreader.

mgravell:
Cancelling a command here feels inherently *just the wrong thing to be doing*. In virtually all scenarios, this would be: bad. It certainly shouldn't be a default. The fact that  *in your particular case* it does no harm if cancelled is incidental. It kinda feels like the real solution here is: don't `select` additional grids that you don't want. As implemented, I think it is best to decline this. There *may* be more acceptable ways of doing it, but I kinda think this is "solving" the wrong problem. The real problem is your sproc.

mgravell:
As a thought: this *might* be acceptable as an explicit `Cancel()` / `CancelAsync()` method on `GridReader`, but... that's abut it. In most other cases, it is *expected* that you will consume the data.

mburbea:
I wish I had control over the sproc but it was not in scope of the project. My task was merely a viewer not to touch the original data. 

As for cancelling, I'd like that very much. I understand that's the intent, but when you're dealing with a legacy db not everything works as expected.

PR #235, opened 2015-01-14
QueryAsync<Dynamic> should return an IEnumerable<DapperRow>

mburbea:
In `QueryAsync<Dynamic>` you will get back an `IEnumerable<object>` requiring you to use `.Cast<IDictionary<String,Object>>()` to get the desired behavior.
However, `Query` returns an `IEnumerable<DapperRow>` so you can use a simple cast or as to  `IEnumerable<IDictionary<string,object>>`

I propose this change to allow this code to just work::
```csharp
var rows = await QueryAsync(cnn, sql) as IEnumerable<IDictionary<string,object>>;
```

### DIFF
--- a/Dapper NET45/SqlMapperAsync.cs
+++ b/Dapper NET45/SqlMapperAsync.cs
@@ -18,18 +18,18 @@ namespace Dapper
         /// Execute a query asynchronously using .NET 4.5 Task.
         /// </summary>
         /// <remarks>Note: each row can be accessed via "dynamic", or by casting to an IDictionary&lt;string,object&gt;</remarks>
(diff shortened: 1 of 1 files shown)


mgravell:
That is almost certainly a breaking change. Does it actually return
IEnumerable<object>? Or does it return IEnumerable<dynamic> ?

mburbea:
Currently, QueryAsync dynamic version returns `IEnumerable<Object>` Query returns `IEnumerable<DapperRow>` both return it as an `IEnumerable<dynamic>`

mgravell:
Are you sure? Looking at the code:

public static IEnumerable<dynamic> Query(...

Vs

public static Task<IEnumerable<dynamic>> QueryAsync(...

It is unclear what needs fixing... Can you clarify?

mgravell:
Oh right, I see; I misread. My bad. Will read the PR tomorrow.

PR #245, opened 2015-02-26
Additional QueryAsync method overrloads

gedbac:
I have faced with some problems in real projects. One of them is parameter count is limited to 7. I have increased to to 16 and have included method very you can pass types as array. Its solves problem when you have 16 or more subtypes. Other issue was that cancellation tokens very hidden while invoking async methods.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -1600,10 +1600,10 @@ this IDbConnection cnn, string sql, object param, IDbTransaction transaction, in
         /// <summary>
         /// Maps a query to objects
         /// </summary>
-        /// <typeparam name="TFirst">The first type in the record set</typeparam>
-        /// <typeparam name="TSecond">The second type in the record set</typeparam>
+        /// <typeparam name="T1">The first type in the record set</typeparam>
+        /// <typeparam name="T2">The second type in the record set</typeparam>
         /// <typeparam name="TReturn">The return type</typeparam>
-        /// <param name="cnn"></param>
+        /// <param name="connection"></param>
         /// <param name="sql"></param>
         /// <param name="map"></param>
         /// <param name="param"></param>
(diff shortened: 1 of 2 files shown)


bartread:
Is there a reason (apart from the now outstanding merge conflicts) that this wasn't merged in or rejected?

I ask only because we've just run into the exact same issue: 7 type parameters for multi-mapping just isn't enough.

Thanks!

johandanforth:
I'll look into it as soon as I can

bartread:
Thanks, although on further inspection I'm not sure it's necessary: there's the version of Query that takes a Type [] and a Func<object [], TResult>, which looks like it's sufficient for scenarios where you need to map back to more than 7 objects.

johandanforth:
Yes and the exposing of the cancellation token will change the interface...

gedbac:
Why there is no option to set cancellation token for async methods and why it's hidden? Using async methods without option to cancel it doesn't make sense.

NickCraver:
Please take this as constructive feedback, not a flat rejection of your work. Contributions are *always* appreciated, even though we don't have time to absorb them in nearly as often as we'd like.

When doing a pull request like this on a major project, it's important to scope the changes to what's *actually* changing. For example by changing `TFirst` to `T1` and `cnn` to `connection` and similar, you have created hundreds of lines of additional changes that aren't relevant to what's happening. Doing this has a few major downsides, most importantly:
 - It makes reviewing and merging the commit much more difficult and time consuming
 - It significantly increases the number of conflicts with *other* pull requests that are touching the same code

The reason I wouldn't merge this isn't based solely on that though, there are 2 main issues in play here:
 - Breaking the interface to the QueryAsync method, this *must* be reserved for a major release
 - The addition of MultiMap overloads

The second is of course subjective, but I think the majority of people aren't returning more than 7 *types*, which is where we stopped adding overloads. There is already a `Func<object [], TResult>` for the `n` result use case as @bartread notes above. Why 7? Simply because we thought it was "reasonable" early on. Why not 10, 20, 100? There can't be a hard and fast rule here, but adding each creates more maintenance overhead when there is a change and again increases the chances of conflict for other pull requests. In my opinion, 15 is over the top here.

The cancellation token addition **is a good change**, and I'd welcome a PR for that - it just has to wait for a major release to be taken in so we don't unexpectedly break people and other dependent libraries.

NickCraver:
I'm closing this one out for the reasoning above, but I'll make sure we look at the Cancellation token for the next release. The overloads missing the token passing were simply missing from #119 and are still an open issue in #313.

PR #248, opened 2015-02-28
Dapper.contrib Get methods state that "Returns a single entity". However...

paulomorgado:
..., they use FirstOrDefault instead of SingleOrDefault, which are sematically different.

### DIFF
--- a/Dapper.Contrib NET45/SqlMapperExtensions.cs
+++ b/Dapper.Contrib NET45/SqlMapperExtensions.cs
@@ -134,7 +134,7 @@ namespace Dapper.Contrib.Extensions
 
             if (type.IsInterface)
             {
-                var res = connection.Query(sql, dynParms).FirstOrDefault() as IDictionary<string, object>;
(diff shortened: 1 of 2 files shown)


johandanforth:
You have a point, but this would change the behaviour too much for projects already used to the FirstOrDefault I'm afraid.

paulomorgado:
@johandanforth, can you think of a use case where the behavior would change?

Isn't performance a requirement for StackExchange?

johandanforth:
I'm thinking about code that could potentially throw exception if we change to SingleOrDefault. Shouldn't happen when you get an object by id, I know, but I'm still not comfortable about it. 

What do you mean with performance @paulomorgado? Could you explain a bit more please?

Thanks for caring about the behavior of contrib, I appreciate it!

paulomorgado:
As you say, the expected result can only have one or no records. And for that you use `SingleOrDefault`.

As to performance; sorry, my mistake. It's actually the other way around. But that's no excuse for lack of correctness.

PR #256, opened 2015-03-20
SqlBuilder for .NET 3.5

henkmollema:
According to: http://stackoverflow.com/questions/29151282/does-dapper-sqlbuilder-exist-for-net-3-5.

- New project which targets .NET 3.5.
- `CSHARP30` debug symbol.
- `#if` statements for dynamic parameters.
- Used `.ToArray()` with `string.Join()`.
- Removed named arguments.


NickCraver:
Are you still on .Net 3.5? There's going to be a point at which we stop deploying new versions for older frameworks but I'm not sure what that line is yet. I think it's safe to say we won't support .Net 2.0 on most things going forward, but I don't have a good handle on the .Net 3.5 usage.

I'm not really opposed to this pull request at all - it seems pretty useful, but it needs a rebase before I can yank it in. Can you get a clean merge happening? Then I'm happy to pull it in.

henkmollema:
No, I'm not using .NET 3.5 myself. Dapper supports .NET 3.5, so I figured why shouldn't the SqlBuilder? I'll see if I can rebase it. Otherwise I might create a new PR for this.

henkmollema:
@NickCraver squashed and rebased.

NickCraver:
Thanks @henkmollema! Merged into master, it'll be in the next release.

henkmollema:
Cool! :smile:

PR #257, opened 2015-03-21
fix  value type bug

zeusro:
during call the method “ QueryAsync<T>”，I found that if the T is value type ,it will cause the  System.NullReferenceException.I find out that the reason is that  the it add null to the List< value type >
                             buffer.Add((T)func(reader));
SqlMapperAsync--Line98
So I add a Condition to judge  the database return

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -3432,6 +3432,10 @@ this IDbConnection cnn, string sql, Func<TFirst, TSecond, TThird, TFourth, TRetu
             return r =>
             {
                 var val = r.GetValue(index);
+                if (type.IsValueType)
+                {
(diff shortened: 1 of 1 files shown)


NickCraver:
Do you have a stack of the exception? Is this caused by querying for something with a `int` vs a `int?`? I'm not sure of the context here.

NickCraver:
Looking at your info here, it looks like you're doing something like `.QueryAsync<int>` against a nullable column where `.QueryAsync<int?>` is appropriate for a nullable column.

We *want* this to fail, it's much better to crash on an incorrect cast that to silently return bad data. I'm closing this out hoping it's just a misunderstanding on the API side (and if so, please let me know how we might improve the messaging there, maybe we should add notes specifically on nullable value types?). If I've totally misunderstood, ping me and I'll re-open it.

PR #258, opened 2015-03-24
Fix #253, use user defined TypeHandlers over internal implementations.

### LINKED ISSUE #253: Insert statement not working with custom TypeHandler
I'm trying to create a custom TypeHandler to handle the following:

* Convert text[] fields from Postgres into List<string> in my object.
* Convert json fields from Postgres into JObject field in my object.

Here is my test class, and the two basic TypeHandler's I've created:

```c#
public class Test
    {
        public List<string> test_array { get; set; }
        public JObject test_json { get; set; }
    }

public class StringListHandler : SqlMapper.TypeHandler<List<string>>
    {
        public override void SetValue(IDbDataParameter parameter, List<string> value)
        {
            string stringValue = "{";

            int i = 0;
            foreach (string s in value)
            {
                if (i > 0) stringValue += ",";

                stringValue += s;
                i++;
            }
           stringValue += "}";

            parameter.Value = stringValue;
        }

        public override List<string> Parse(object value)
        {
            List<String> strings = new List<String>();
            string stringValue = (string)value;
            stringValue = stringValue.Substring(1, stringValue.Length - 2);
            char[] delim = { ',' };
            string[] words = stringValue.Split(delim);

            foreach (string word in words) {
                strings.Add(word);
            }

            return strings;
        }
}

public class JObjectHandler : SqlMapper.TypeHandler<JObject>
    {
        public override void SetValue(IDbDataParameter parameter, JObject value)
        {
        }

        public override JObject Parse(object value)
        {
            string stringValue = (string)value;
            JObject obj = JObject.Parse(stringValue);

            return obj;
        }
}
```

I register the TypeHandlers in by WebApiConfig:

```c#
SqlMapper.AddTypeHandler(typeof(List<string>), new StringListHandler());
SqlMapper.AddTypeHandler(typeof(JObject), new JObjectHandler());
```

When I do the SELECT statement, the Parse method is getting called from my type handlers correctly and I am able to get a List<Test> objects back using below:

```c#
PgSqlConnection connection = ConnectionFactory.GetOpenConnection();

PgSqlCommand cmd = new PgSqlCommand();
cmd.Connection = connection;
cmd.CommandText = "SELECT test_array FROM test";

var tests = connection.Query<Test>("select * from test").ToList();
```

But when I try to do an INSERT, the SetValue method is never called, and I get an error back (a1 is not a column):

```c#
Test t = new Test();
t.test_array = new List<string>();
t.test_array.Add("hello");

t.test_json = new JObject();
t.test_json.Add(new JProperty("hi", "bye"));

var operation = "insert into test (test_array, test_json) VALUES (@a,@b)";

PgSqlConnection conn = ConnectionFactory.GetOpenConnection();
conn.Execute(operation, new { a=t.test_array, b=t.test_json });
```

I saw similar questions being asked on StackOverflow: http://stackoverflow.com/questions/25568657/dapper-typehandler-setvalue-not-being-called with no resolution, so I figured I would check here if it was an issue in Dapper or with the way I am using it.

adminnz:
basically SqlMapper was using its own internal IEnumerable type handling even if the user has defined their own type handler for IEnumerable types (e.g. List<String>).

I've switched the order so it will check if the user has defined a type handler for this, and then use the IEnumerable implementation.
Also I added two tests, to DapperTests NET40 with a very simple List<String> TypeHandler implementation that will read Strings into a List<String> by splitting the string comma. And also will use join a list by comma and set a parameter from a List<String>.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -885,15 +885,16 @@ namespace Dapper
             {
                 return DbType.Binary;
             }
+            if (typeHandlers.TryGetValue(type, out handler))
+            {
+                return DbType.Object;
+            }
             if (typeof(IEnumerable).IsAssignableFrom(type))
             {
                 return DynamicParameters.EnumerableMultiParameter;
             }
 
-            if (typeHandlers.TryGetValue(type, out handler))
-            {
-                return DbType.Object;
-            }
+
             switch (type.FullName)
             {
                 case "Microsoft.SqlServer.Types.SqlGeography":
--- a/Tests/Tests.cs
+++ b/Tests/Tests.cs
@@ -1571,10 +1571,10 @@ end");
                             2 as BlogId, 'Blog' as Title";
             var postWithBlog = connection.Query<Post_DupeProp, Blog_DupeProp, Post_DupeProp>(sql,
                 (p, b) =>
-            {
-                p.Blog = b;
-                return p;
-            }, splitOn: "BlogId").First();
+                {
+                    p.Blog = b;
+                    return p;
+                }, splitOn: "BlogId").First();
 
             postWithBlog.PostId.IsEqualTo(1);
             postWithBlog.Title.IsEqualTo("Title");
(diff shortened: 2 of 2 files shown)


espenekvang:
Any indications on when this pullrequest can be merged in? We have implemented a custom typehandler but SetValue is never called.

jhovgaard:
Spent time "bug fixing" this too. +1 for merge.

RufusJWB:
I have the same problem and it would be great, if it could be fixed.

klesh:
Support +1
It would be great to fix this problem.

johandanforth:
I merged the fix and the tests provided by @adminnz . Would you all please test things a bit and give feedback before I update Nuget as well? I normally only work with the Contrib parts but this one seems pretty obvious to merge in so...Thanks for your patience. ping @espenekvang  @jhovgaard @RufusJWB @klesh

klesh:
Hi, Johan, thanks for your help.
I checked the test case and ran it, everything seems alright...
All two test cases are passed. @johandanforth

johandanforth:
Great. I'm closing this one then.

RufusJWB:
I couldn't help testing, but I highly appreciate your quick response! Thank you very much!

jhovgaard:
Thanks for the merge! Works here.

PR #269, opened 2015-04-15
Added extension method to simplify using SqlBuilder.Templates with Query method

Obalix:
Added the static DapperSqlBuilderMixin class inside the SqlBuilder project to provide SqlSpecific extensions to Dapper.

The first extension method implemented here is a method simplifying the use of SqlBuilder.Template with the Query method. So instead of writing

`cnn.Query(template.RawSql, template.Parameters, ...);`

the user can write

`cnn.Query(template, ...);`

### DIFF
--- a/Dapper.SqlBuilder NET45/Dapper.SqlBuilder NET45.csproj
+++ b/Dapper.SqlBuilder NET45/Dapper.SqlBuilder NET45.csproj
@@ -39,6 +39,9 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


NickCraver:
We'd like to pull this in, but can you please add a unit test for the actual execution? This also illustrates the usage to someone finding it. In addition, for public methods we really need XML comments so someone using intellisense can see how to use this method. It's a good addition, and one I'd use personally - just needs a bit more love :)

NickCraver:
Closing due to inactivity and inability to merge - ping to reopen if anything changes!

PR #271, opened 2015-04-16
Add convert logic code to `QueryAsync` method

harrywong:
In current code, if we use QueryAsync<int>, when the db returns a long value, the code will break. Because long cannot be convert to int directly.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -3443,6 +3443,10 @@ this IDbConnection cnn, string sql, Func<TFirst, TSecond, TThird, TFourth, TRetu
             return r =>
             {
                 var val = r.GetValue(index);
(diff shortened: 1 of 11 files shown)


NickCraver:
Conflicts because of #338, need to sanity check the null situation though.

NickCraver:
I appreciate the fixes here, there's a lot of overlap with #338 which I already pulled in the difference being the null checks. I wanted to quickly explain that we *want* to fail on a improper case due to nullability if the type doesn't support it. For example, if you have a null in a nullable int column then you should be using a `int?` (`Nullable<int>`) rather than `int`. Silently returning a 0 in place of nulls would be a bad thing in most situations - so we actually want the casts to fail here and alert the user to the problem. I hope that's clear for anyone finding this later - if not just comment and I'll try to clarify.

Issue #282, opened 2015-05-12
Dapper-level global/default CommandTimeout


Irrational86:
Hello all,

I've searched a bit, but haven't found anything on the subject.

Is there any opposition to implementing a global (or default) `CommandTimeout` property within Dapper? I am opening this Issue in hopes someone can provide positive or negative feedback on the suggestion.

The way this new property would work is by looking at the nullable `commandTimeout` parameter passed to the `CommandDefinition` class, and if null, use the new CommandTimeout property.

I realize that Dapper works via Extension Methods, which are static, and therefore are not instantiated, which in turn helps in maintaining differing property values (if necessary). I find that this is OK, given that the option to specify the Command Timeout is provided on each call to `Execute`, `ExecuteReader`, `Query`, and `QueryMultiple`.

The alternative, which I currently use, but do not like the additional layer is as follows:
- Create a new class which inherits from `DbConnection`
- Act as pass-thru on all of the overridable methods, except `CreateDbCommand` (which is the more involved one)
- In `CreateDbCommand`, hold the return value in a variable, modify the `CommandTimeout` property as desired, and return
- *See sample class here: [DB Connection Class with default CommandTimeout](https://gist.github.com/176f853a25e00e9abc37.git)*

I would initially take a stab at implementing it, and I am fairly confident I can. Any feedback would be greatly appreciated.

Thanks,
Jesse

mgravell:
I don't have any fundamental objections to that, as long as it defaults to
the existing "don't explicitly set the timeout" behavior - the main change
would be a trivial tweak to the `CommandDefinition.SetupCommand` method.
I'm sure a PR would be accepted with this.

On 12 May 2015 at 03:06, Jesse <notifications@github.com> wrote:

> Hello all,
>
> I've searched a bit, but haven't found anything on the subject.
>
> Is there any opposition to implementing a global (or default)
> CommandTimeout property within Dapper? I am opening this Issue in hopes
> someone can provide positive or negative feedback on the suggestion.
>
> The way this new property would work is by looking at the nullable
> commandTimeout parameter passed to the CommandDefinition class, and if
> null, use the new CommandTimeout property.
>
> I realize that Dapper works via Extension Methods, which are static, and
> therefore are not instantiated, which in turn helps in maintaining
> differing property values (if necessary). I find that this is OK, given
> that the option to specify the Command Timeout is provided on each call to
> Execute, ExecuteReader, Query, and QueryMultiple.
>
[...]

PR #284, opened 2015-05-12
SqlMapper global/default CommandTimeout

Irrational86:
The way this new property works is by looking at the nullable `commandTimeout` parameter passed to the `CommandDefinition` class, and if null, use the new `CommandTimeout` Setting.

Dapper works via Extension Methods, which are static, therefore not allowing instantiated versions to hold different values. I find that this is OK, given that the option to specify the Command Timeout is provided on each call to `Execute`, `ExecuteReader`, `Query`, and `QueryMultiple`.

An alternative, which I currently use, but do not like the additional layer is as follows:
- Create a new class which inherits from `DbConnection`
- Act as pass-thru on all of the overridable methods, except `CreateDbCommand` (which is more involved)
- In `CreateDbCommand`, hold the return value in a variable, modify the `CommandTimeout` property as desired, and return it
- *See sample class here: [DB Connection Class with default CommandTimeout](https://gist.github.com/176f853a25e00e9abc37.git)*

Additional discussion on Issue #282.

I hope this initial design helps pave the way for additional settings and features.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -253,6 +253,30 @@ namespace Dapper
     static partial class SqlMapper
     {
         /// <summary>
+        /// Permits specifying certain SqlMapper values globally.
+        /// </summary>
+        public static class Settings
+        {
+            static Settings()
+            {
+                SetDefaults();
+            }
+
+            /// <summary>
+            /// Resets all Settings to their default values
+            /// </summary>
+            public static void SetDefaults()
+            {
+                CommandTimeout = null;
+            }
+
+            /// <summary>
+            /// Specifies the default Command Timeout for all Queries
+            /// </summary>
+            public static int? CommandTimeout { get; set; }
+        }
+
+        /// <summary>
         /// Implement this interface to pass an arbitrary db specific set of parameters to Dapper
         /// </summary>
         public partial interface IDynamicParameters
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -183,7 +183,13 @@ namespace Dapper
                 cmd.Transaction = transaction;
             cmd.CommandText = commandText;
             if (commandTimeout.HasValue)
+            {
                 cmd.CommandTimeout = commandTimeout.Value;
+            }
+            else if (SqlMapper.Settings.CommandTimeout.HasValue)
+            {
+                cmd.CommandTimeout = SqlMapper.Settings.CommandTimeout.Value;
+            }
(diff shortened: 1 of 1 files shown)


Irrational86:
@mgravell  Thanks Marc!

perragradeen:
How is this used?

Link is broken
See sample class here: DB Connection Class with default CommandTimeout

perragradeen:
Found it ex.
     var items = connection.Query<MainItemForList>(query, param, commandTimeout: queryTimeoutInSeconds, buffered: false);

(from https://stackoverflow.com/questions/15683197/weird-timeout-issues-with-dapper-net/32524428#32524428)

Irrational86:
@perragradeen Link above has been updated. The sample class explains how to subclass/inherit the `DbConnection` class in order to set a global timeout.

This Pull Request has already been merged into Dapper, so if you're using Dapper, you can simply set `SqlMapper.Settings.CommandTimeout = 60`. Then, all queries and commands executed via Dapper will have a timeout of 60 seconds (or whatever value you specify).

The more granular option, as you demonstrated above, is to specify the timeout on each query or command independently.

perragradeen:
I cannot find "Settings" in the static Class Dapper.SqlMapper
Functions that exist are for example SetTypeName(...)

Irrational86:
@perragradeen This pull requests is targeted for the 1.5.0 release of Dapper. If you use NuGet, you should be able to use the beta package (though betas are usually not recommended for production use): https://www.nuget.org/packages/Dapper.

mgravell:
I should note that we're dogfooding the beta in production at
stackoverflow. Will review asap.

NickCraver:
I *really* want to rename these, with a unit. e.g. `CommandTimeoutSeconds` It'd be a breaking change on the global, but less impactful on the param names (still breaking if they're specified in code due to optional). Maybe in a 2.0 release...

Irrational86:
@NickCraver feel free to change it. I created the original PR, and since it's in beta, I am probably one of the very few using it.

mgravell:
IMO: if this is only in the beta - we can do this.

On 15 December 2015 at 01:38, Nick Craver <notifications@github.com> wrote:

> I *really* want to rename these, with a unit. e.g. CommandTimeoutSeconds
> It'd be a breaking change on the global, but less impactful on the param
> names (still breaking if they're specified in code due to optional). Maybe
> in a 2.0 release...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/StackExchange/dapper-dot-net/pull/284#issuecomment-164614131>
> .
>



-- 
Regards,

Marc

PR #285, opened 2015-05-15
Added support to automatically expand IEnumerables of >2100 parameters into a TVP when configured. 

mmurrell:
In our usage of Dapper, we occasionally run into lists with many parameters.  While it is rare that the count exceeds 2100, it's not unheard of.  

This patch allows me to register a UDT on a per-type basis, and will automatically switch to a TVP if the max is exceeded.

```C#
Dapper.SqlMapper.Expander.Add(typeof(int), new Dapper.SqlMapper.ExpanderRegistration() { UdtTypeName = "IntList", UdtFieldName = "Id", Type = typeof(int), Count = 2000 });
```

Then, code like this will use multiple parameters
```C#
var ids = Enumerable.Range(20, 200);
var sql = "SELECT COUNT(*) FROM dbo.AutoExpand WHERE Id IN @ids";
var results connection.Query(sql, new { ids })
```

e.g.
```SQL
SELECT COUNT(*) FROM dbo.AutoExpand WHERE Id IN (@ids1, @ids2, @ids3 ... )
```

And when the enumerable exceeds a configured count, Dapper automaticaly switches to a strongly typed TVP:
e.g. 
```SQL
DECLARE @ids IntList
INSERT INTO @ids (Id) VALUES (1)
... 
INSERT INTO @ids (Id) VALUES (2000)

SELECT COUNT(*) FROM dbo.AutoExpand WHERE Id IN (SELECT Id from @ids)
```

It's still a little rough around the edges, but if there's any interest in merging this in, I'll clean it up.  Let me know. 

Thanks

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -4758,7 +4758,7 @@ string name, object value = null, DbType? dbType = null, ParameterDirection? dir
             {
                 if (default(T) != null)
                 {
-                    throw new ApplicationException("Attempting to cast a DBNull to a non nullable type!");
+                    throw new ApplicationException("Attempting to cast a DBNull to a non nullable type! Note that out/return parameters will not have updated values until the data stream completes (after the 'foreach' for Query(..., buffered: false), or after the GridReader has been disposed for QueryMultiple)");
                 }
                 return default(T);
             }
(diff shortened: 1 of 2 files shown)


mgravell:
this is an interesting idea; will read today

PR #286, opened 2015-05-18
TypeHandler should be supported for enum

ShamsulAmry:
In the database, I'm storing single character status codes, and an `enum` to match the status codes. However, since the status codes are not integer and not direct string match against the enum I created, I implemented a `TypeHandler` to do custom deserialization from the status code to the enum. Then I found out Dapper will not call the `TypeHandler` for enum types, so here's the fix.

Basically I moved the `TypeHandler` cache check to the beginning of the block even before enum value processing because I believe user-defined TypeHandlers should come first before anything else, no?

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -3799,7 +3799,13 @@ Type type, IDataReader reader, int startBound = 0, int length = -1, bool returnN
                         var nullUnderlyingType = Nullable.GetUnderlyingType(memberType);
                         var unboxType = nullUnderlyingType != null && nullUnderlyingType.IsEnum() ? nullUnderlyingType : memberType;
 
-                        if (unboxType.IsEnum())
+                        if (typeHandlers.ContainsKey(unboxType)) 
+                        {
+#pragma warning disable 618
+                            il.EmitCall(OpCodes.Call, typeof(TypeHandlerCache<>).MakeGenericType(unboxType).GetMethod("Parse"), null); // stack is now [target][target][typed-value]
+#pragma warning restore 618
+                        }
+                        else if (unboxType.IsEnum())
                         {
                             Type numericType = Enum.GetUnderlyingType(unboxType);
                             if(colType == typeof(string))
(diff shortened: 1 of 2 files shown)


vdaron:
Related to #259

ShamsulAmry:
Related to #258 too

drusellers:
:+1: want

NickCraver:
I believe this is now properly handled, [see current code here](https://github.com/StackExchange/dapper-dot-net/blob/4326e0c4d165ae218cad4dae3c90e4fdbcb90c31/Dapper%20NET40/SqlMapper.cs#L3875-L3884):
```C#
if (hasTypeHandler)
{
#pragma warning disable 618
    il.EmitCall(OpCodes.Call, typeof(TypeHandlerCache<>).MakeGenericType(unboxType).GetMethod("Parse"), null); // stack is now [target][target][typed-value]
#pragma warning restore 618
}
else
{
    il.Emit(OpCodes.Unbox_Any, unboxType); // stack is now [target][target][typed-value]
}
```

If I'm missing something though, comment here and I'm happy to open the issue back up.

vdaron:
Hi, current master branch doesn't work for me. The LookupDbType will not return the ITypeHandler if one is defined for an Enum and deserializer IL generated code seems to ignore ITypeHandler for Enums.

This sample ITypeHandler is ignored for example :

```csharp

public enum TestEnum
{
   One,Two,Three
}

public class EnumAsIntTypeHandler<T> : ITypeHandler
{
   public void SetValue(IDbDataParameter parameter, object value)
   {
      parameter.DbType = DbType.Int32;
      parameter.Value = (int)value;
   }
   public object Parse(Type destinationType, object value)
   {
      return (T)value;
   }
}

...
SqlMapper.AddTypeHandler(typeof(TestEnum), new EnumAsIntTypeHandler<TestEnum>());
...
```

Thanks for feedback @NickCraver !

floyd-may:
+1, same behavior experienced as @vdaron. Need to be able to map letter codes to enums and back.

PR #287, opened 2015-05-20
Added support to automatically expand IEnumerables of >2100 parameters into a TVP when configured. 

mmurrell:
**Edit:** Not sure what I did to close #285.  This is the new commit to reopen.  I'll copy the comments from the original request.

**Original Request:** 
In our usage of Dapper, we occasionally run into lists with many parameters.  While it is rare that the count exceeds 2100, it's not unheard of.  

This patch allows me to register a UDT on a per-type basis, and will automatically switch to a TVP if the max is exceeded.

```C#
Dapper.SqlMapper.Expander.Add(typeof(int), new Dapper.SqlMapper.ExpanderRegistration() { UdtTypeName = "IntList", UdtFieldName = "Id", Type = typeof(int), Count = 2000 });
```

Then, code like this will use multiple parameters
```C#
var ids = Enumerable.Range(20, 200);
var sql = "SELECT COUNT(*) FROM dbo.AutoExpand WHERE Id IN @ids";
var results connection.Query(sql, new { ids })
```

e.g.
```SQL
SELECT COUNT(*) FROM dbo.AutoExpand WHERE Id IN (@ids1, @ids2, @ids3 ... )
```

And when the enumerable exceeds a configured count, Dapper automaticaly switches to a strongly typed TVP:
e.g. 
```SQL
DECLARE @ids IntList
INSERT INTO @ids (Id) VALUES (1)
... 
INSERT INTO @ids (Id) VALUES (2000)

SELECT COUNT(*) FROM dbo.AutoExpand WHERE Id IN (SELECT Id from @ids)
```

It's still a little rough around the edges, but if there's any interest in merging this in, I'll clean it up.  Let me know. 

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,11 @@ NuGet.exe
 *.nupkg
 .docstats
 *.ide/
+
+*.nupkg
+# The packages folder can be ignored because of Package Restore
+**/packages/*
+# except build/, which is used as an MSBuild target.
+!**/packages/build/
+# Uncomment if necessary however generally it will be regenerated when needed
+#!**/packages/repositories.config
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -724,7 +724,14 @@ namespace Dapper
             typeMap[typeof(object)] = DbType.Object;
 
             AddTypeHandlerImpl(typeof(DataTable), new DataTableHandler(), false);
+
+            Expander = new Dictionary<Type, ExpanderRegistration>();
+
         }
+        /// <summary>
+        /// A set of per-type configurations for automatically expanding IEnumerables into TVPs
+        /// </summary>
+        public static Dictionary<Type, ExpanderRegistration> Expander { get; set; }
 
         /// <summary>
         /// Clear the registered type handlers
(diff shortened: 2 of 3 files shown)


NickCraver:
I have a few concerns here, on a few fronts:

1) The "automatic" change - the penalty for TVPs is extremely severe. In our environment it'd be much better to loop over the action (pipelined and possibly async) than reverting to a TVP. I would much rather get an error and go fix it into a loop than have a TVP fallback (for *my* apps, just as an example).
2) I'm don't think a static is appropriate here - this changes behavior for commands where it may not be intended - most behavior of this type is per-command via `CommandFlags`.
3) It excludes the `packages` folder via the `.gitignore`

For number 2, you can for example be using Dapper to hit multiple data stores in an application. With a static, assumptions are made and breaks introduced when not-really-shared things there.

To expound on number 3, pull requests should haven't items like this - it's a rather major change that's not even mentioned here. While it may seem like an oversight, it's a very intentional exclusion. We consciously removed the NuGet restore dependency present to build the package without it by including the packages - the trade-off in repo size (and binary weight) were deemed worth it to greatly reduce the dependency chain.

I think if this behavior is desired, it should probably be activated by a new `CommandFlags` entry, allowing a great deal more granular control within the existing APIs. Admittedly, I'm biased against TVPs because they are prohibitively expensive on the performance side for Stack Overflow, but I definitely recognize use cases where they're valuable.

Issue #270, opened 2015-04-16
Query method with IDbCommand parameter


alexeyfadeev:
We have to support legacy project using DbCommand and manual mapping. We are thinking start using Dapper-dot-net for mapping instead. But how to pass IDbCommand parameter to Query method?

I have tried to add these methods to SqlMapper class, and it works! Is it possible to add such methods to project?

public static IEnumerable<T> Query<T>(this IDbConnection cnn, IDbCommand cmd)
{
    var data = QueryImpl<T>(cnn, cmd, typeof(T));
    return data.ToList();
}

private static IEnumerable<T> QueryImpl<T>(this IDbConnection cnn, IDbCommand cmd, Type effectiveType)
{
    var command = new CommandDefinition(cmd.CommandText);
    var identity = new Identity(command.CommandText, command.CommandType, cnn, effectiveType, null, null);
    var info = GetCacheInfo(identity, cmd.Parameters, command.AddToCache);

    IDataReader reader = null;

    bool wasClosed = cnn.State == ConnectionState.Closed;
    try
    {
        if (wasClosed) cnn.Open();
        reader = cmd.ExecuteReader(wasClosed ? CommandBehavior.CloseConnection | CommandBehavior.SequentialAccess : CommandBehavior.SequentialAccess);
        wasClosed = false; // *if* the connection was closed and we got this far, then we now have a reader
        // with the CloseConnection flag, so the reader will deal with the connection; we
        // still need something in the "finally" to ensure that broken SQL still results
        // in the connection closing itself
        var tuple = info.Deserializer;
        int hash = GetColumnHash(reader);
        if (tuple.Func == null || tuple.Hash != hash)
        {
            if (reader.FieldCount == 0) //https://code.google.com/p/dapper-dot-net/issues/detail?id=57
                yield break;
            tuple = info.Deserializer = new DeserializerState(hash, GetDeserializer(effectiveType, reader, 0, -1, false));
            if (command.AddToCache) SetQueryCache(identity, info);
        }

        var func = tuple.Func;
        var convertToType = Nullable.GetUnderlyingType(effectiveType) ?? effectiveType;
        while (reader.Read())
        {
            object val = func(reader);
            if (val == null || val is T)
            {
                yield return (T)val;
            }
            else
            {
                yield return (T)Convert.ChangeType(val, convertToType, CultureInfo.InvariantCulture);
            }
        }
        while (reader.NextResult()) { }
        // happy path; close the reader cleanly - no
[...]

PR #292, opened 2015-06-03
Query method with IDbCommand parameter added

alexeyfadeev:
Added Query<T> and QueryImpl<T> methods with IDbCommand argument. It can be usefull to upgrade legacy projects using DbCommand and manual mapping and start using Dapper instead.

Also with this method no need to rebind the DbParameter collection to DbCommand from DynamicParameter List every time.

In addition, it's possible to use Dapper and LINQ to SQL together. LINQ to SQL allows to convert LINQ query to DBCommand, then it should be passed to this new Dapper method.

Additional discussion on Issue #270.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -1511,6 +1511,18 @@ this IDbConnection cnn, Type type, string sql, object param = null, IDbTransacti
             return command.Buffered ? data.ToList() : data;
         }
 
+        /// <summary>
+        /// Executes a query, returning the data typed as per T
+        /// </summary>
+        /// <remarks>No need to rebind DbParameter collection to DbCommand</remarks>
+        /// <returns>A sequence of data of the supplied type; if a basic type (int, string, etc) is queried then the data from the first column in assumed, otherwise an instance is
+        /// created per row, and a direct column-name===member-name mapping is assumed (case insensitive).
+        /// </returns>
+        public static IEnumerable<T> Query<T>(this IDbConnection cnn, IDbCommand cmd)
+        {
+            var data = QueryImpl<T>(cnn, cmd, typeof(T));
+            return data.ToList();
+        }
 
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


mgravell:
Looks good; I will probably tweak it before deploy to add a  `bool buffered = true` optional parameter (the other optional parameters are redundant if the command is specified), and to see if we can do anything to merge the Impl method(s) rather than copy/paste, but: appreciated.

On "it's possible to use Dapper and LINQ to SQL together" - the caveat there is that there would be no binding to the DBContext; but: do you have an example of this usage?

mgravell:
I've been looking at this locally, and I wonder if this is actually the wrong way to approach this problem - I've come up with a number of thorny issues that make it easy to shoot yourself in the foot here.

I wonder - the key thing that we're trying to solve is the *object population* stage - i.e. consuming an `IDataReader` and populating objects. I wonder if the better thing to do here would be for me to add a method like

     pulblic static IEnumerable<T> Populate(this IDbDataReader reader) {...}

such that the caller can do (instead) something like:

    using(var reader = cmd.ExecuteReader()) {
        var myList = reader.Populate<Customer>().ToList();
    }

this seems to avoid a huge number of edge case problems that I'm thinking are likely to cause lots of issues. We can of course then provide a helper method that does the above (essentially the equivalent of your `Query<T>(cmd)`, but this is a different implementation to solve the same fundamental problem.

Also; we probably don't need to pass the connection in this case, since the connection is also an existing property of a command.

Thoughts?

alexeyfadeev:
Thank you for fast answer. I don't know Dapper source code well, and it would be better if this feature is performed by developer, more competent in Dapper structure. That's why I posted Issue #270 in April. A Desicion with Populate method is perfect.

Passing a connection certainly do not need, you are right.

About LINQ 2 SQL - Context is needed. Then why bother using it? Because LINQ queries are good and Dapper is fast.
Example, extension method:

public static List<T> DapperExecute<T>(this IQueryable<T> query, DataContext context)
{
    var command = context.GetCommand(query);
    return (List<T>)context.Connection.Query<T>(command);
}

alexeyfadeev:
Sorry, wrong markdown

    public static List<T> DapperExecute<T>(this IQueryable<T> query, DataContext context)
    {
        var command = context.GetCommand(query);
        return (List<T>)context.Connection.Query<T>(command);
    }

alexeyfadeev:
May be, better to add extension method to IDBCommand.

    var myList = cmd.Execute<Customer>().ToList();

alexeyfadeev:
I have removed Connection from arguments. It works.
But idea with Populate method for IDbReader and Query helper method above is better.

DamianReeves:
@mgravell I really feel having an extension method off of IDataReader is needed and would be helpful. I have current situations  where some other library I don't own gives me a IDataReader and it needs to be Deserialized/Materialized/Populated (whatever you want to call it) to an object. 
Right now Dapper has the code to do this, but to get at it you need to access the private `GetDeserializer` method.
The [Insight.Database](https://github.com/jonwagner/Insight.Database) project, which was originally based on Dapper, provides access through IDataReader and it is very helpful.

I end up taking in SqlMapper as a code file (which I wish we still directly supported through NuGet) and add the following:

    partial class SqlMapper
    {
        public static IEnumerable<T> ToSequence<T>(this IDataReader reader)
        {
            if (reader == null)
            {
                throw new ArgumentNullException("reader");
            }
            var deserializer = GetDeserializer(typeof(T), reader, 0, -1, false);
            while (reader.Read())
            {
                yield return (T)deserializer(reader);
            }
        }

        public static IList<T> ToList<T>(this IDataReader reader)
        {
            return reader.ToSequence<T>().ToList();
        }
    }

DamianReeves:
The code referenced above would be better if we could some how hook into the deserializer cache provided by Dapper, but that functionality was above my pay grade.

NickCraver:
@mgravell +1 on this approach:
```C#
public static IEnumerable<T> Populate<T>(this IDbDataReader reader) {...}
```

I have situations where I need to read things off a DataReader directly like schema and such which prevents using most of Dapper's utility. Having the inner-bits split out into a `Populate<T>` style call would be excellent for lots of low-level use cases like that.

I'll take a poke at splitting that out as a PR this week or next and see if I hit any issues.

mgravell:
I'm declining this as per the discussion above; an alternative approach will be provided to do the same thing after the FileSplitFromHell™.

Thanks for pointing out the need for this.

PR #294, opened 2015-06-05
Tests and Fixes #293

### LINKED ISSUE #293: Dapper.Contrib `Insert` throws when trying to select `@@identity` on SQL Server 2008R2
I've just realized my issue:

The Dapper.Contrib documentation says: "As with dapper, all extension methods assume the connection is already open, they will fail if the connection is closed. "

Since the first half of that statement was false, I assumed the full statement was false. (Dapper can and will open a closed connection on your behalf).  *However* they're independent statements.  I recommend the documentation be updated to "Unlike Dapper, these methods assume the connection is already open, and they wlil fail if the connection is closed." it will make people pay attention to the statement.

**-------- EDIT -------**

I just started playing with Dapper.Contrib for some of my simpler CRUD needs.  Everything seems to be working *except* inserts, which throw the following exception. (Stack Trace starts in Dapper.Contrib). The insert *does* successfully happen, but the returning of the identity value fails.

```
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: Cannot convert null to 'int' because it is a non-nullable value type
   at CallSite.Target(Closure , CallSite , Object )
   at SqlServerAdapter.Insert(IDbConnection connection, IDbTransaction transaction, Nullable`1 commandTimeout, String tableName, String columnList, String parameterList, IEnumerable`1 keyProperties, Object entityToInsert) in ~\DapperContrib.cs:line 549
   at Dapper.Contrib.Extensions.SqlMapperExtensions.Insert[T](IDbConnection connection, T entityToInsert, IDbTransaction transaction, Nullable`1 commandTimeout, ISqlAdapter sqlAdapter) in ~\DapperContrib.cs:line 252
```

My class looks like this:

```csharp
using LocalCopyOf.Dapper.Contrib.Extensions;
using Newtonsoft.Json;

namespace MySite.Models {
    [Table("section_header")]
    public class SectionHeader {
        [Key]
        public int? my_pk { get; set; }

        public string description { get; set; }

        public int? sort_key { get; set; }

        public string url { get; set; }

        public string role_needed { get; set; }

        [Computed]
        [JsonProperty("link_lines")]
        public Link[] Links { get; set; }

        public LinkHeader WithLinks(Link[] lines) {
            var header = (Link) this.MemberwiseClone();
            header.Links = lines;
            return header;
        }
    }
}
```

The table itself is defined as such:

```sql
CREATE TABLE [dbo].[section_header](
	[my_pk] [int] IDENTITY(1,1) NOT NULL,
	[description] [varchar](100) NULL,
	[sort_key] [int] NULL,
	[access_fk] [int] NULL,
	[url] [varchar](2000) NULL,
	[image_url] [varchar](2000) NULL,
	[internal] [bit] NULL,
	[role_needed] [varchar](50) NULL
) ON [PRIMARY]
```

There are various and sundry unneeded fields.

Crisfole:
I'm not 100% sure that this is the best way to fix this, but it does work (wrote a failing test, and this fixes that test).

### DIFF
--- a/Dapper.Contrib.Tests/Tests.cs
+++ b/Dapper.Contrib.Tests/Tests.cs
@@ -91,6 +91,17 @@ namespace Dapper.Contrib.Tests
             }
         }
 
+        public void InsertOnClosed()
+        {
+            using (var connection = GetConnection()) {
+                var id = connection.Insert(new User() {Name = "Adama", Age = 10});
+                var user = connection.Get<User>(id);
+                user.Id.IsEqualTo((int) id);
+                user.Name.IsEqualTo("Adama");
+                connection.Delete(user);
+            }
+        }
+
         public void InsertList()
         {
             const int numberOfEntities = 100;
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -268,18 +268,23 @@ namespace Dapper.Contrib.Extensions
                 if (i < allPropertiesExceptKeyAndComputed.Count() - 1)
                     sbParameterList.Append(", ");
             }
-
-            if (!isList)    //single entity
+            int returnVal;
+            bool wasClosed = connection.State == ConnectionState.Closed;
+            if (wasClosed) connection.Open();
+            if (!isList) //single entity
             {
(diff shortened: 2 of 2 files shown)


johandanforth:
Done

PR #298, opened 2015-06-09
Brackets around property names in Update & Delete

Tewr:
Modified Update & Delete method to be consistent with Insert by adding brackets around property names.I did this cause I thought it was annoying that I could do an insert without problems but had to resort to a type mapping when doing an update, all because of a property name which was a keyword in sql


johandanforth:
Thanks @Tewr It's in

PR #302, opened 2015-06-10
Update Contrib README with small nuance about Delete

shrayasr:
Deleting an entity will delete only by a specified key property and not by any specified parameter

Illustrated by the following case: 

```csharp
[Table ("employee")]
public class Employee
{
    public int Id { get; set; }
}

// Doesn't work since "Name" isn't a key
_conn.Delete(new Employee { Name = "foo" });

// Works since "Id" is a key
_conn.Delete(new Employee { Id = 14 });
```

### DIFF
--- a/Dapper.Contrib/Readme.md
+++ b/Dapper.Contrib/Readme.md
@@ -87,7 +87,7 @@ connection.Update(cars);
 
 `Delete` methods
 -------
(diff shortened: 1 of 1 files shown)


shrayasr:
:+1: Thaanks :)

PR #310, opened 2015-07-06
remove commented json so dnx tooling doesn't break

aggieben:
These comments break dnu out of the box, and there isn't any .net 3.5 support planned for dnx, so I recommend just removing these.


cbeall:
Along with the comments in the project.json file causing problems, I'm pretty sure the two project.lock.json (in Dapper and Dapper.DNX.Test)  files should be removed from source control and would need to be part of the gitignore.

PR #311, opened 2015-07-06
Runtime dependency

aggieben:
The `System.Runtime` package shouldn't be necessary for any of the full-framework TFMs, and the dependency on a now-outdated beta6 build causes breakage on latest beta6 nightlies.  Alternatively, if there is some obscure thing happening that made this necessary that I'm not aware of, I'd recommend a floating dependency (`"-*"`).

### DIFF
--- a/Dapper/project.json
+++ b/Dapper/project.json
@@ -13,7 +13,7 @@
         "net45": {
             "compilationOptions": { "define": [ "ASYNC" ], "warningsAsErrors": true },
             "dependencies": {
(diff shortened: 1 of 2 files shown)


aggieben:
So, it turns out I was wrong about where the breakage I was experiencing was coming from.  This change is still valid IMO, but turned out to not be the culprit (what _did_ turn out to be the culprit was a stale `project.json.lock` in another project).

mgravell:
Thanks Ben; sorry for delay

PR #317, opened 2015-07-14
DBMS selection for extra features overidable in one place

lukemcgregor:
Previously only overridable inside the Contrib extensions now overridable for
both core and extensions. Means that if you are using a wrapped
connection you can change the detection method to suit.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -1606,7 +1606,7 @@ this IDbConnection cnn, string sql, object param, IDbTransaction transaction, in
                 while (reader.Read())
                 {
                     object val = func(reader);
-					if (val == null || val is T) {
+                    if (val == null || val is T) {
                         yield return (T)val;
                     } else {
                         yield return (T)Convert.ChangeType(val, convertToType, CultureInfo.InvariantCulture);
@@ -5195,18 +5195,14 @@ string name, object value = null, DbType? dbType = null, ParameterDirection? dir
     /// </summary>
     partial class FeatureSupport
     {
-        private static readonly FeatureSupport
-            @default = new FeatureSupport(false),
-            postgres = new FeatureSupport(true);
-
         /// <summary>
-        /// Gets the feature set based on the passed connection
+        /// Gets the featureset based on the passed connection
         /// </summary>
         public static FeatureSupport Get(IDbConnection connection)
         {
             string name = connection == null ? null : connection.GetType().Name;
-            if (string.Equals(name, "npgsqlconnection", StringComparison.OrdinalIgnoreCase)) return postgres;
-            return @default;
+            if (Config.GetDbmsType(connection) == DbmsType.Postgres) return new FeatureSupport(true); //Postgresql supports arrays
+            return new FeatureSupport(false);
         }
         private FeatureSupport(bool arrays)
         {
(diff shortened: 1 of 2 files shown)


Dapper NET40/SqlMapper.cs
```diff
@@ -5195,18 +5195,14 @@ public void AddParameter(IDbCommand command, string name)
...
-        /// Gets the feature set based on the passed connection
+        /// Gets the featureset based on the passed connection
         /// </summary>
         public static FeatureSupport Get(IDbConnection connection)
         {
             string name = connection == null ? null : connection.GetType().Name;
-            if (string.Equals(name, "npgsqlconnection", StringComparison.OrdinalIgnoreCase)) return postgres;
-            return @default;
+            if (Config.GetDbmsType(connection) == DbmsType.Postgres) return new FeatureSupport(true); //Postgresql supports arrays
+            return new FeatureSupport(false);
```
NickCraver: This change causes an extra allocation for every call, the previous static cache was explicitly to avoid allocations.

Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -418,26 +418,35 @@ private static string GetTableName(Type type)
         /// Specifies a custom callback that detects the database type instead of relying on the default strategy (the name of the connection type object).
         /// Please note that this callback is global and will be used by all the calls that require a database specific adapter.
         /// </summary>
-        public static GetDatabaseTypeDelegate GetDatabaseType;
-
-
-        private static ISqlAdapter GetFormatter(IDbConnection connection)
+        [Obsolete("Set Dapper.Config.GetDbmsTypeFromConnection instead, this gets used for more than just extensions")]
+        public static GetDatabaseTypeDelegate GetDatabaseType
```
NickCraver: 2 items: Obsolete should wait for a future release (not suitable for a 1.43), and the getter was removed - breaking the interface here. A getter needs to be added back, as  that is used in some downstream code doing swaps.

NickCraver:
I added a few comments inline, can you take a look please?

Dapper NET40/SqlMapper.cs
lukemcgregor: Fair enough, ill put it back to the statics

Dapper.Contrib/SqlMapperExtensions.cs
lukemcgregor: No problems putting the getter back.

In terms of the obsolete flag, I can remove it until a later version if requited but I'm not too sure why this is an issue. The obsolete flag is only a compile time hint there is a newer way of doing this rather than a breaking change.

lukemcgregor:
@NickCraver, left comments inline and fixed noted issues. Just have a query around the obsolete flag outstanding.

NickCraver:
@lukemcgregor While there are a lot of bits here that make sense today, locking it down to an enum inside the lib eliminates adding adapters outside of the core lib later. After discussing with @mgravell I think we want to stick with strings here (though we could move them to constants) and change around how some of these things work. I think we'll move off the interface completely doe to versioning issues since it's not exposed in a usable way externally anyway.

Thank you for the request, it made us think quite a bit about how the best way to expose all of these bits moving forward and brought up a exposed issues with previous plans.

lukemcgregor:
This means you can't use a wrapped provider with dapper postgres.

PR #324, opened 2015-08-05
Added basic support for custom type maps in collection parameters

gjsduarte:
Added LookupDbType on the first iteration of the PackListParameters
method.

Possible fix for collection parameters ignoring custom type maps


NickCraver:
Thanks! Just now getting these pulled in, sorry for the delay :)

PR #338, opened 2015-09-10
fix Cast at QueryAsync<T>

phnx47:
I had a problem when converting long to int, when returning the auto-increment Id.


NickCraver:
Thanks for the fix! I'll add a unit test which addresses #346 shortly.

PR #339, opened 2015-09-11
Always use TypeHandler for parameters if available

kashifsoofi:
TypeHandlers can override the logic for built-in SQL data types (int,
byte, DateTime etc.) when these are return values via
ITypeHandler.Parse, but ITypeHandler.SetValue was never being used for
parameters. Moved the typeHandlers.TryGetValue() call up to make sure a
type handler is always used if available.

### DIFF
--- a/Dapper NET40/SqlMapper.cs
+++ b/Dapper NET40/SqlMapper.cs
@@ -925,7 +925,7 @@ namespace Dapper
         internal static DbType LookupDbType(Type type, string name, bool demand, out ITypeHandler handler)
         {
             DbType dbType;
-            handler = null;
+            typeHandlers.TryGetValue(type, out handler);
             var nullUnderlyingType = Nullable.GetUnderlyingType(type);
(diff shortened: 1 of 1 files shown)


kashifsoofi:
Credit goes to Matt Wilkinson for original change (#177)

drusellers:
:+1: want

NickCraver:
While I can see the use here, unfortunately this change can't really happen without a lot more love. There are no handlers for other types in the pipe such as most IEnumerable types, or any custom types where no such mapping actually exists. If you run the unit tests programs (e.g. build and `\Tests\bin\Debug\Smackdown.exe`), you can see a great many downstream breaks due to how parameters are handled in general.

NickCraver:
Since we can't merge either - I'll consolidate this into #177 so we can track the issue in one place.

kwaclaw:
It seems #177 has been closed as well, without any resolution to the issue.

Issue #336, opened 2015-09-08
Enclosing column names in square brackets breaks Dapper.Contrib in PostgresSQL


oliverw:
Enclosing column names in square brackets completely breaks Dapper.Contrib in PostgresSQL.

<code>insert into products([Title], [ManufacturerId],) values((('Vibe P1')::text), ((1)::int4)) RETURNING Id</code>

johandanforth:
@oliverw Could you please explain a bit more what you are trying to do. The dapper.contrib insert extension has been tested quite a lot I believe.

oliverw:
Well, the queries above, and the one below have been generated by Dapper.Contrib (latest version from NuGet). When running the query below through NpgSql 3.0 (though happens on 2.x as well)  it results in the following error (sorry for the German error message, I have yet to figure out how to force the PGSQL error log to be in English):

<code>
2015-09-16 12:24:17 CEST	FEHLER	Syntaxfehler bei '[' bei Zeichen 30	
2015-09-16 12:24:17 CEST	ANWEISUNG	update rss_product_feeds set [Url] = $1, [Region] = $2, [Marketplace] = $3, [Name] = $4, [Created] = $5, [LastProcessed] = $6, [TimesProcessed] = $7 where [Id] = $8	
</code>

If I patch SqlMapperExtensions.cs to omit generating the square brackets, everything works as expected.

NickCraver:
@oliverw sanity check here: I see an extra comma in your column list - you sure that's not a contributing factor here...or is that a mis-paste?

AndrewSmart:
See [mrjoltcola's](http://stackoverflow.com/questions/19657101/what-is-the-difference-between-square-brackets-and-single-quotes-for-aliasing-in) explanation for a detailed understanding of this issue.

This square bracket issue causes Dapper.Contrib to not work with MySQL, and as oliverw noted, PostgresSQL.

As a solution, I suggest augmenting the [ISqlAdapter interface](https://github.com/StackExchange/dapper-dot-net/blob/bbf93cb260b10ca88ae036d74b2789a6fd72fb27/Dapper.Contrib/SqlMapperExtensions.cs#L628) with columnList and parameterList creation methods, such that the columns & values are quoted with appropriate delimiters for each SQL provider where appropriate.

tms:
[...]

PR #345, opened 2015-09-29
Dapper.Contrib GetWhere implementation

majorsilence:
Retrieve data with custom where clause without writing sql.

### DIFF
--- a/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
+++ b/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
@@ -122,6 +122,46 @@ namespace Dapper.Contrib.Extensions
             return list;
         }
 
+        /// <summary>
+        /// Returns a list of entites from table
+        /// </summary>
+        /// <typeparam name="T">Interface or type to create and populate</typeparam>
+        /// <param name="connection">Open SqlConnection</param>
+        /// <returns>Entity of T</returns>
+        public static async Task<IEnumerable<T>> GetWhereAsync<T>(this IDbConnection connection, object param, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
+        {
+            var type = typeof(T);
+
+            var props = param.GetType().GetProperties();
+            if (!props.Any())
+            {
+                throw new DataException("GetWhere<T> must have param");
+            }
+
+            var sql = new StringBuilder();
+            var name = GetTableName(type);
+
+            sql.Append("select * from ");
+            sql.Append(name);
+            sql.Append(" where ");
+
+            bool setAnd = false;
+            foreach (var prop in props)
+            {
+                if (setAnd)
+                {
+                    sql.Append(" AND ");
+                }
+
+                sql.Append(prop.Name);
+                sql.Append("=@");
+                sql.Append(prop.Name);
+                setAnd = true;
+            }
+
+            return await connection.QueryAsync<T>(sql.ToString(), param, transaction, commandTimeout: commandTimeout);
+
+        }
 
         /// <summary>
         /// Inserts an entity into table "Ts" asynchronously using .NET 4.5 Task and returns identity id.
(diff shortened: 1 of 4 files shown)


johandanforth:
Looks interesting. I would like to hear from a few others what they think before I pull this one in :)
So what do you guys think?

mgravell:
Looks useful; code observations:

1. the `throw new DataException("GetWhere<T> must have param");` exception is vague and unclear - the problem is actually the lack of properties
2. it will be specific to particular SQL variants thanks to the `@` - that's a really awkward problem to work around, I realize
3. in some ways it would be nice if the `GetWhereAsync` could use an `await GetOpenConnectionAsync()`

majorsilence:
For point 2 I have added a new Symbol property to ISqlAdapter.  The other option I can think of is adding a new GetWhere method to ISqlAdapter.

mgravell:
Is it expected that external developers will implement ISqlAdapter? Or is
it intended to be internally provided only? Because if the former: this is
a compilation breaking change...

On 5 October 2015 at 19:46, Peter Gill <notifications@github.com> wrote:

> For point 2 I have added a new Symbol property to ISqlAdapter. The other
> option I can think of is adding a new GetWhere method to ISqlAdapter.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/StackExchange/dapper-dot-net/pull/345#issuecomment-145628785>
> .
>



-- 
Regards,

Marc

NickCraver:
I'm also curious about `ISqlAdapter`, but more concerned with what it will and won't support. For example, can I pass a list that translates to an `IN` clause? What about dates between? Greater than, less than? Like clauses? How about `IS NULL` and `IS NOT NULL`?

None of the above are supported, and we're edging into equals-only ORM generation land which IMO is better to leave out than to bring in extremely limited (and therefore confusing). I think to do it *well* we're really leaving Dapper's core goals and getting into pretty severe complication amongst each provider due to differences in many of these syntaxes across at least *one* of the platforms. From the practical usability standpoint, what would you pass in? Usually you'd use the same type as you'd deserialize from, but that's not really practical as they'd all be parameters in the WHERE clause - meaning you're making duplicated types (possibly anonymous) to do the query. I'm not sure that's significantly worse than SQL text alternative (what we're good at), but worth pointing out for real-world scenarios.

Thoughts from others?

majorsilence:
I should probably rethink this.  @NickCraver has some good points that is affecting my team.  We use the current implementation of GetWhere but are left wanting many of the features listed but missing.

johandanforth:
I'm not sure but I never thought ISqlAdapter was to be implemented by external developers. The interface can be useful for a few other issues though, like #336, where it can be used to build different select column clauses depending on sql-version. @NickCraver Could you please explain more what you mean with "we're edging into equals-only ORM generation land which IMO is better to leave out than to bring in extremely limited". I don't really fully understand (english is not my first language...), and I want to :)

NickCraver:
@johandanforth by that I mean we don't support a ton of the *very common* query cases, much less anything really flexible here. It's really getting into full ORM territory. You simple can't pass an object and form a query - it comes down to that really. Unfortunately for any `WHERE`/`AND` statement needs the operator and 1 or more operands (for example a date range would be 2). In general that's getting into *much* heavier ORM territory and not really outside the goals of Dapper.

I'm speaking only for me here - but my thoughts are that *if* we go down this route of offering an interface like this at all (I don't think we should), then we need to at the very minimum handle all of the common query cases, `<`, `>`, `=`, `LIKE`, `IS NULL`, `IN`, `BETWEEN`, etc. - and simply can't be done with the pass-in-an-object model that's so lightweight. and pervasive in Dapper today. 

It's a sliding spectrum of "what should be automatic?" to "what can we generate?". At the end of the day anything is building a string for a query and throwing it at SQL. If it requires so much work to do operators, etc. then at what point is it better to just write the SQL query (how Dapper works today)? I'd argue that happens well before it's worth having an interface like this. I think that if a user's goals are much more towards the generation end of things, then Dapper simply isn't the best tool for the job and another ORM is likely a much better fit.

tms:
I agree with @NickCraver, it seems like a dark road to go down. If this is something that a lot of people might find useful, maybe another option would be to extend on `SqlBuilder` with some helpers to generate common templates? Then you'd have more access to the flexibility of specifying your own SQL, but save some typing.

It's still likely to be provider-specific in some cases though, so it might not be appropriate for Dapper proper.

PR #347, opened 2015-10-01
Only set UdtTypeName when value isn't DBNull

perliedman:
Without this fix, trying to set a column of type `geometry` (and presumably other UDT types) causes an `ArgumentException` with the message: "UdtTypeName property must be set only for UDT parameters."

Only setting the `UdtTypeName` when the value is not `DBNull` resolves the problem.


mgravell:
Thanks - much appreciated

Issue #351, opened 2015-10-09
Dapper.Contrib not passing key column with value


erick2red:
Let me explain, because I could not find a proper title for this

I have this schema: `ObjectX(ObjectXId(Guid), Name(Varchar))`, and this POCO

```csharp
public class ObjectX
{
  [Key]
  public Guid ObjectXId { get; set; }

  public string Name { get; set; }
}
```

Now, using Dapper.Contrib I write

```csharp
var o = new ObjectX();
o.ObjectXId = mycustomGuid; // and this is unique
o.Name = "Foo";
sqlConnection.Insert(o);
```

And this code is failing with an exception inside Dapper, saying `ObjectXId` cannot be null. For some reason Dapper.Contrib is throwing away the value I passed as key.

Is it assuming the `[Key]` column will be autogenerated, or autoincremented?

AndrewSmart:
I just ran into the same issue.

> Is it assuming the [Key] column will be autogenerated, or autoincremented?

Yes:
https://github.com/StackExchange/dapper-dot-net/blob/bbf93cb260b10ca88ae036d74b2789a6fd72fb27/Dapper.Contrib/SqlMapperExtensions.cs#L271-L282

Duplicate of #11. #11 was closed on June 7th with a claim that it works, but it would not then either:
https://github.com/StackExchange/dapper-dot-net/blob/e408faba82d1d48a19f82b472d82e7d3f0aedba4/Dapper.Contrib/SqlMapperExtensions.cs#L270-L281

I would say Dapper.Contrib does not currently support inserts for application supplied keys.

This conflicts with the [readme.md](https://github.com/StackExchange/dapper-dot-net/blame/master/Dapper.Contrib/Readme.md#L23-L24):
> For these extensions to work, the entity in question MUST have a key-property, a property named "id" or decorated with a [Key] attribute.

My question is would the Dapper project welcome a patch to allow an application supplied key for inserts? I presume a new data annotation such as [NonAutoIncrementKey] or [ExplicitKey] would be an appropriate solution?

johandanforth:
[...]

PR #352, opened 2015-10-13
Fixes for issue #351 and #336

johandanforth:
Affects Dapper.Contrib only. Added support for application supplied keys, introduced a new `[ExplicitKey]` attribute for this. Added new tests for this as well as a couple of test projects for MsSql and SQLite.


Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -657,19 +702,58 @@ public int Insert(IDbConnection connection, IDbTransaction transaction, int? com
     {
         var cmd = String.Format("insert into {0} ({1}) values ({2})", tableName, columnList, parameterList);
         connection.Execute(cmd, entityToInsert, transaction, commandTimeout);
-        var r = connection.Query("select @@IDENTITY id", transaction: transaction, commandTimeout: commandTimeout);
+        var r = connection.Query("select @@IDENTITY id", transaction: transaction, commandTimeout: commandTimeout).ToList();
+
+        if (r.First().id == null) return 0;
+        var id = (int) r.First().id;
```
NickCraver: I'm not following how this would work, e.g. if the ID column is a GUID a lot of these chains still assume integer behavior. We'd only really be supporting `int` and "less than `int`"...or am I missing something?

Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -628,27 +659,41 @@ public class ComputedAttribute : Attribute
...
+        var multi = connection.QueryMultiple(cmd, entityToInsert, transaction, commandTimeout);
 
-        var id = (int)multi.Read().First().id;
-        var propertyInfos = keyProperties as PropertyInfo[] ?? keyProperties.ToArray();
-        if (!propertyInfos.Any()) return id;
+        var first = multi.Read().FirstOrDefault();
+        if (first == null || first.id == null) return 0;
+
+        var id = (int)first.id;
+        var pi = keyProperties as PropertyInfo[] ?? keyProperties.ToArray();
```
NickCraver: IMO this code was much more readable before - is there a legitimate reason for stripping readability from many (most repeated) variable names in this?

Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -628,27 +659,41 @@ public class ComputedAttribute : Attribute
 public partial interface ISqlAdapter
 {
     int Insert(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, String tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert);
+    
+    //new methods for issue #336
+    void AppendColumnName(StringBuilder sb, string columnName);
+    void AppendColumnNameEqualsValue(StringBuilder sb, string columnName);
```
NickCraver: Note: **breaking interface change**

Dapper.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.22728.1
+# Visual Studio 2013
+VisualStudioVersion = 12.0.31101.0
```
NickCraver: This will cause problems - we have a VS loading regression here.

Dapper.Contrib.MsSqlTests NET45/Program.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private static void DropTables()
+        {
+
+            using (var connection = new SqlConnection("Data Source = .\\SQLEXPRESS;Initial Catalog=DapperContribMsSqlTests;Integrated Security=SSPI"))
+            {
+                connection.Open();
+                connection.Execute("alter database DapperContribMsSqlTests set single_user with rollback immediate");
+                connection.Execute(@" drop table Stuff");
```
NickCraver: Do these not error on the first run? The pattern is to try/catch here, or the T-SQL could contain the IF check.

NickCraver:
I think a lot of this has good changes, but I don't agree with the direction of `Insert`, because it doesn't account for non-interger keys and IMO introduces more problems inadvertently. While an object's property would be set right, for non-interger or integer convertible types we're returning 0. IMO that's very misleading and can easily lead to breaks.

I'd propose we have `T Insert<T>` and `Task<T> InsertAsync<T>` for other key types where the return value is actually useful. This can be a non-breaking change with existing methods calling into their `int` versions (which would be JIT inlined). This would offer a wider range of support and cover issues such as #351 more properly.

NickCraver:
Side note: I'm not sure about the ISqlAdapter changes - I kind of agree on having all of this on the base, but I think it needs to be an abstract class at this point (rather than an unversioned interface) given the number of things we're seeing diverge on platforms down-the-road style. It's a wider issue than this PR because there are more things elsewhere like array handling, TVPs, etc. Thoughts? /cc @mgravell 

What are your thoughts on this? /cc @mgravell as well

Dapper.Contrib/SqlMapperExtensions.cs
johandanforth: Yes you're right, it's really an attempt to prevent breaking the interface.

Dapper.Contrib/SqlMapperExtensions.cs
johandanforth: True, I was not sure about changing the variable names actually and I'm thinking of going back to the old, more readable names!

Dapper.sln
johandanforth: Aow... sorry about that one.

Dapper.Contrib.MsSqlTests NET45/Program.cs
johandanforth: Actually not sure what you mean Nick. SetupMsSqlDatabase(), SetupTables and DropTables() are part of the setup, not the tests.

johandanforth:
Thanks for the review @NickCraver ! Good thoughts on the Insert() parts, and I agree with you on going with  `T Insert<T>`  and . `Task<T> InsertAsync<T>`. As for ISqlAdapter - it's clear I need to read up on effects of changing interfaces or having abstract classes instead and what is best for projects like this. Looking forward to what you and @mgravell says. I'll be happy to do the necessary work on the code :)

NickCraver:
@johandanforth We met up and concur - can you implement the Insert<T> underneath as-is today? The `int Insert()` can just call into `Insert<int>` underneath without breaking the API. There are much better ways to handle the outgoing bits for GUID, etc. via `OUTPUT` and `RETURNING` than are done today - I can take care of that.

If you can fix the `Insert` methods up (or I can) and rename the parameters, we're ready to do a lot of PR merging and code splitting to make life much easier for anyone working on Dapper. Also, the plan is to make ISqlAdapter into an abstract class (which isn't practically exposed today since the dictionary is private), but I'd much rather do that post pull request since PR merging is the bottleneck.

Also, can you make sure you've enabled 2-factor auth? Removing from the org was due to not detecting it - if you enable 2-factor auth on GitHub I can re-add you to the project :)

Thanks!

NickCraver:
I merged this in manually with the fixes so we can keep rolling /cc @mgravell - I figure we can do the split today then I can work the `Insert`. As-is @johandanforth's work fixes longs and such so it's net-win already. Making it work for all key types next can just be another commit. 

Thanks @johandanforth! and please ping me when the 2-factor is ready to go - if I add you before that's in place, our security scripts will just keep reverting it.

johandanforth:
Been reallyt busy, but I'll take some time to both read through all comments from your massive work on the repo and fix the 2-factor thingy soon. You guys have sure been busy lately ;)

johandanforth:
Ping @NickCraver I've got 2FA setup now!

johandanforth:
I've now given a lot of love to the Insert methods, added a T Insert<T> as discussed above. Will create a PR when I've tested it a bit more, but it looks great and I'm pretty pleased with what I'm looking at :blush: /cc @NickCraver @mgravell

Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -657,19 +702,58 @@ public int Insert(IDbConnection connection, IDbTransaction transaction, int? com
...
+    }
+}
+
+public partial class MySqlAdapter : ISqlAdapter
+{
+    public int Insert(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, String tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert)
+    {
+        var cmd = String.Format("insert into {0} ({1}) values ({2})", tableName, columnList, parameterList);
+        connection.Execute(cmd, entityToInsert, transaction, commandTimeout);
+        var r = connection.Query("Select LAST_INSERT_ID()", transaction: transaction, commandTimeout: commandTimeout);
```
AndrewSmart: I got the time to try this out, thank you again for your work! Though looks like I'm late it was already merged into master.

You got the last_insert_id() right for MySQL, instead of last_insert_rowid()
I ran Dapper.Contrib.Tests, and the ShortIdentity() test failed here:
```
public void ShortIdentity()
        {
            using (var connection = GetOpenConnection())
            {
                var id = connection.Insert(new Stuff() { Name = "First item" });
                id.IsEqualTo(1);
```
id was 0, so it was not equal to 1 as expected.

I just followed the lead of the SQLiteAdapter.Insert method in changing the above to:
```
var cmd = String.Format("insert into {0} ({1}) values ({2}); select last_insert_id() id", tableName, columnList, parameterList);
var r = connection.QueryMultiple(cmd, entityToInsert, transaction, commandTimeout);
var id = (int)r.Read().First().id; //(int) cast necessary as MySQL returns last_insert_id() as ulong.
//if (id == null) return 0; //this line no longer necessary.
```
With that change all tests passed when using MySQL.

johandanforth: nice @AndrewSmart I'll change the code and make sure it comes along in the end (If I ever get through all the "under the cover" changes I'm deep in... :confused: I'm *so* thinking of starting on a fresh Dapper.Contrib.2 :D

neeasade:
Hi, I'm looking to insert with Guid generated ID and I'm having a bit of trouble flowing through related issues -- the conclusion here was a delay/Dapper.Contrib.2 without support for non-database-generated GUID as PK currently, right?

Issue #355, opened 2015-10-15
Dapper.Contrib Should use OUTPUT instead of SCOPE_IDENTITY


johandanforth:
Currently Dapper.Contrib uses `SCOPE_IDENTITY` for returning id of MsSql inserts, but should use `OUTPUT INSERTED` instead, according to this support statement from Microsoft https://support.microsoft.com/en-us/kb/2019779

I'm going to look at that.

oreleraki:
Hi, I'm stumbling the same issue, after research i arrived to here. It appears neither v1.* nor v2.* versions has this fix, which is very important when using non Integer Keys. Are there any plans to put this fix into v1.* and v2.* ?

PR #356, opened 2015-10-15
fix for issue #355

johandanforth:
Use OUTPUT INSERTED instead of SCOPE_IDENTITY for MsSql inserts in Dapper.Contrib

### DIFF
--- a/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
+++ b/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
@@ -298,20 +298,21 @@ public partial class SqlServerAdapter
 {
     public async Task<int> InsertAsync(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, String tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert)
     {
-        var cmd = String.Format("insert into {0} ({1}) values ({2});select SCOPE_IDENTITY() id", tableName, columnList, parameterList);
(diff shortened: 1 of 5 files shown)


NickCraver:
Hey @johandanforth, thanks for the push - most of it looks good but we need to not break people. For example, removing `public async Task<int> InsertAsync` isn't a good option - we need that exist and just call `InsertAsync<T>` underneath - just a simple wrapper. The method signatures that are there shouldn't change since it breaks anyone downstream that references the lib and hasn't recompiled and re-released. In the case of `Task<int> InsertAsync<T>` I would expect a `Task<TKey> InsertAsync<T, TKey>`, the current implementation has changed from returning *they key* to *the type itself* which is a very different method and not what I think you intended.

Can you please add back the original methods and fix types? I'd use `TKey` when it's a key and `T` when it's the type, this makes it very clear in intellisense. Preferably add the current (now wrapper) methods above the new ones so they aren't even a diff which makes things way cleaner. I'm traveling this weekend but will take a second review look as soon as I'm able. Thanks!

johandanforth:
Hmm, I'm not sure how I managed to mess up the Task<int> InsertAsync<T> and turn it into Task<long>... probably beacuse the "sync" version returned a long and the async returned an int and I just moved the code over the the async version of Insert...  I'll fix that and make sure the old interface stays the same (which was my intent all along) . As you say, underneath that I just call the new TKey Insert<TKey> method. I'll use TKey for keys and T for types. I'll also move the methods - good tip :)

johandanforth:
I'm closing this one and will send another PR when I've had another look at the code again. No breaking interface, no break...

PR #358, opened 2015-10-20
Fix broken link in README

### LINKED ISSUE #357: Can be faster link broken
The section in the readme about select mapping with dapper against various SQL frameworks has a link explaining that all can approaches can be faster. This link is broken.

Atvaark:
The website just changed their URI a little bit.

Fixes #357


NickCraver:
Thanks for the fix!

PR #363, opened 2015-10-24
Contrib stuff

johandanforth:
I'm down with a terrible cold, but I'd like you to have a look at this code now. I've tried to take into consideration all the feedback, especially about not breaking the interface, and hopefully this is close to it. I've ran the code with the old tests from "way back", and they all pass (tried on MsSql, MsSqlCe and SQLite) with the new <TKey> Insert<Tkey>() and the old long Insert<T>().

Also found a bug and added tests for the new [ExplicitKey] attribute.

Anyway, I'm back to bed. Looking forward to read feedback on this. /cc @mgravell @NickCraver

### DIFF
--- a/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
+++ b/Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
@@ -298,20 +298,21 @@ public partial class SqlServerAdapter
 {
     public async Task<int> InsertAsync(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, String tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert)
     {
-        var cmd = String.Format("insert into {0} ({1}) values ({2});select SCOPE_IDENTITY() id", tableName, columnList, parameterList);
-        var multi = await connection.QueryMultipleAsync(cmd, entityToInsert, transaction, commandTimeout);
+        var propertyInfos = keyProperties as PropertyInfo[] ?? keyProperties.ToArray();
+        if (propertyInfos.Any())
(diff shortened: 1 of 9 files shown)


NickCraver:
Rather than one huge PR/commit set, can you please use individual pull requests per feature/bug? With this massive approach it breaks many *other* pull requests at once and also makes the decision a whole accept/reject of many things at once (and if any are rejects, it's a full reject or wait on otherwise useful changes we could pull in immediately). It also makes any review much more grueling and greatly delays it happening.

I see many issues in the initial pass such as `InsertAsync<TKey>` vs. `InsertAsync<T, TKey>` (as we discussed previously), but I think it's a much better idea to discuss all of those in individual PRs. Also, @mgravell is splitting all of these files out into something far more manageable which will (to be blunt) steam roll many PRs anyway...which is why we made a push to merge all pending PRs in before doing so.

johandanforth:
Sure I could do individual request, but how? I can't see any option in the website or the tools I have to do a PR when looking at a specific commit. I've googled the topic, but they all suggest differnt ways that involves so many steps it looks ridicolous! Is there a "simple" way of creating a PR from a commit in the github webpage? Like from this one? https://github.com/johandanforth/dapper-dot-net/commit/4752a21208be2cd165239c456258e81abf820718

johandanforth:
Right, I'll see if I can get all the git-commands right... BBL

johandanforth:
Sorry @NickCraver I give up trying to create separate PR's because I don't speak git well enough to prevent things from getting worse. I usually work with many small commits and all but one commit above is around the Insert method. I'll see if I can reset my fork, do the changes again to the code and create just one or two pull requests for you.

PR #369, opened 2015-10-27
new TKey Insert<TKey> method and updated readme

johandanforth:


NickCraver:
These methods aren't quote right - they'll behave *very* differently from current. If I did a poor job of explaining, I apologize. For example: `Task<int> InsertAsync<T>()` loses all generic type safety going down by throwing away the per-type compilation (which is part of the optimization). What *should* happen is a call to `Task<TKey> InsertAsync<T, TKey>` underneath, so that `InsertAsync<T>` (remaining for convenience and non-breaking changes) calls `InsertAsync<T, int>`.

Honest assessment is we're so far off here, it's better to start fresh - I intended to do this change but after Marc completed the split...which is now done. I'll try and get to it this weekend.

PR #371, opened 2015-10-28
fixed GetAll when used with ExplicitKey

janzakrzewskipl:


NickCraver:
Thanks! Bringing this in to fix the bug. Long term though, I think we likely want to change how `[ExplicitKey]` works. I think with attribute inheritance we can eliminate a lot of duplicate code and checks here. /cc @johandanforth

PR #377, opened 2015-11-03
Removing project.lock.json files.

sebastienros:
These files are already ignored but must have been added before the rule was added.


sebastienros:
Will recreate after a fresh pull

mgravell:
K, I'll stop pushing for a bit (sorry if I broke your PR)! Will be happy to merge.

sebastienros:
Or just delete the two `project.lock.json` files, which is the purpose of this PR.

PR #378, opened 2015-11-03
Removing project.lock.json files.

sebastienros:
These files are already ignored but must have been added before the rule was added.


sebastienros:
This branch has conflicts for the exact reason this branch exists ;)

NickCraver:
What the hell, git? I `git rm`ed these bastards before the rebase - how do they keep coming back? Apologies for the incompetence, I'll use more napalm this time.

mgravell:
Napalm applied; thanks

PR #382, opened 2015-11-04
Fixed missing check for explicit key in Contrib GetaAync()

johandanforth:


Dapper.Contrib NET45/SqlMapperExtensionsAsync.cs
```diff
@@ -32,17 +32,18 @@ public static partial class SqlMapperExtensions
             if (!GetQueries.TryGetValue(type.TypeHandle, out sql))
             {
                 var keys = KeyPropertiesCache(type);
-                if (keys.Count() > 1)
-                    throw new DataException("Get<T> only supports an entity with a single [Key] property");
-                if (!keys.Any())
-                    throw new DataException("Get<T> only supports en entity with a [Key] property");
+                var explicitKeys = ExplicitKeyPropertiesCache(type);
+                if (keys.Count() > 1 || explicitKeys.Count() > 1)
```
NickCraver: Shouldn't this be `if (keys.Count() + explicitKeys.Count() > 1)`?

NickCraver:
Pulling this in as it fixes the one breaking test - I wan to revisit how explicit keys are implemented in general though - we can be much more DRY here with attribute inheritance I think.

johandanforth:
👍

PR #384, opened 2015-11-06
DNX improvements

henkmollema:
Several improvements regarding DNX:

- Correctly use floating versions in `project.json`
- Move `global.json` to root folder
- Add `NuGet.config` with stable NuGet v3 feed
- Add `global.json` and `NuGet.config` to solution items
- Use version `1.50-*` when developing

I did not commit the `project.lock.json`'s as they should be removed by #378


NickCraver:
Most of this looks good, +1 - but why change the literal version in Dapper's project.json? That's the version we actually build and should be only changed on release really.

henkmollema:
I tend to develop against `-*` and pin it to a version when I release. Just a personal preference though, doesn't really matter. I'll revert it.

NickCraver:
@henkmollema This one *is* the release - it totally makes sense for the test project to do that (it's already that way), however, this one is the actual version of the NuGet package we're building. I just wanted to be sure we're talking about the same file here: `Dapper/project.json`

henkmollema:
Yeah, we're talking about the same file. :smile: 

What I mean is: when developing, I don't know which version it's gonna be yet when releasing. `beta1` is the version which is _already_ released (as it's on NuGet).

NickCraver:
I'm curious - how are you choosing what the NuGet version is? This seems to be a matter of taste since there are a few ways. Do you go with specifying it on the package command line when generating?

henkmollema:
In a similar way as MS does with it's ASP.NET 5 packages: when releasing merge `dev` into `master`, pin the versions in the project.json's files on master branch and then use `dnu pack` to create nupkg's.

NickCraver:
+1 on this, eyeball @mgravell since you're last in here?

mgravell:
Will look once I've had coffee. I'm still very much in the "learning phase"
on the DNX stuff, so very much appreciate all input. Coffee first, though!

mgravell:
Looks good; agree with Nick on the version, but you already reverted that. I think the floating versions were *working*, but agree it is more explicit with the obvious wildcard. Didn't know you could move the global.json up a level - that file was generated by the IDE. I wonder if the IDE will screw it all up if I change version, though. Much appreciated.

henkmollema:
@mgravell the global.json is supposed to be at 'repository' level. When you create new project with Visual Studio the tooling will also add it to the root folder and to the solution items, just like I did.

Regarding the wildcards: without it the error list gets full of errors such as:
> Dependency specified was `System.Reflection.Emit.Lightweight` >= `4.0.1-beta-` but ended up with `System.Reflection.Emit.Lightweight` `4.0.1-beta-23409`.


/cc @NickCraver

PR #386, opened 2015-11-06
Code improvements to utilize C# 6 features

henkmollema:
- Overall formatting (except the big `SqlMapper(Async)`)
- Use `var` where possible
- Use string interpolation
- Use expression bodies
- Use auto, read-only properties
- Use null progagation
- Use `nameof` operator
- Arrange `this.` and namespace qualifiers
- Use dictionary initializers

### DIFF
--- a/Dapper/WrappedReader.cs
+++ b/Dapper/WrappedReader.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Data;
 using System.Collections;
+
 #if DNXCORE50
 using IDbDataParameter = global::System.Data.Common.DbParameter;
 using IDataParameter = global::System.Data.Common.DbParameter;
@@ -50,22 +51,12 @@ namespace Dapper
             }
         }
 
-        public override int Depth
-        {
-            get { return Reader.Depth; }
-        }
+        public override int Depth => Reader.Depth;
 
-        public override bool IsClosed
-        {
-            get { return reader == null ? true : reader.IsClosed; }
-        }
-        public override bool HasRows
-        {
-            get
-            {
-                return Reader.HasRows;
-            }
-        }
+        public override bool IsClosed => reader?.IsClosed ?? true;
+
+        public override bool HasRows => Reader.HasRows;
+
         public override bool NextResult()
         {
             return Reader.NextResult();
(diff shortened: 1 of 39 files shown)


Dapper/WrappedReader.cs
```diff
@@ -209,19 +194,12 @@ public override bool IsDBNull(int i)
...
-        public override object this[int i]
-        {
-            get { return Reader[i]; }
-        }
+        public override object this[int i] => Reader[i];
     }
 #else
-
-    internal class WrappedReader : IDataReader, IWrappedDataReader
+    internal class WrappedReader : IWrappedDataReader
```
NickCraver: Better to leave IDataReader on here to be explicit, clearer to read

NickCraver:
Most of this is great - and something I was going to get to this weekend, so +1. Can you revert the `var` changes though? That's a matter of style preference, and being explicit makes the code more readable and approachable. In some cases it actually changes the behavior and what can/won't break due to the interface assumptions by the compiler. I think we'd much rather leave all of those explicit.

Heads up: you'll likely have to rebase after we pull in existing PRs since this creates more conflict than any of the others. That sucks, but it's the only way to go with "changes everywhere" type PRs - I hope that makes sense!

henkmollema:
Do you want to revert _every_ `var` or just the ones I changed in this PR?

If the mess is too big, I might create a new branch and do the changes again. Most of them were done by Resharper anyway. 

By the way: what do you think about the formatting changes? I.e.: brackets at one-line `if` statements and correct tabs etc? I could create a PR without that and just the C# 6 features, might be more clear what's changed.

NickCraver:
I'd only revert what was done here - to minimize the change surface. If creating a new PR: I'd simply do 2 commits: one C#6 and one formatting - they can be looked at separately that way but still accomplish the one big merge. By tabs I'll assume you meant spaces ;)

henkmollema:
Will do, could you ping me when some big PR's are merged?

And yes, spaces of course :smile:

mgravell:
Lots of niceness here. I'm not too concerned about the `var` personally - the only one that reduces readability is the `IEnumerable multiExec = GetMultiExec` ones - IMO those should remain, as it isn't necessarily obvious what is coming back without it. But: I definitely love C# 6.

henkmollema:
Created a new PR: #391

PR #387, opened 2015-11-10
Increase extensibility of SqlBuilder.

MrJerB:
By setting the access modifier to protected for the AddClause() method, users of the class may be able to inherit and extend the functionality.

### DIFF
--- a/Dapper.SqlBuilder/SqlBuilder.cs
+++ b/Dapper.SqlBuilder/SqlBuilder.cs
@@ -119,9 +119,9 @@ namespace Dapper
         }
 
 #if CSHARP30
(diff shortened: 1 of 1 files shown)


NickCraver:
@MrJerB given that this will make a mess of the git history - would you rather I make the change here, or do you want to try submitting through another means? I'm perfectly happy with the change, just not the collateral damage here :)

MrJerB:
@NickCraver It's your repository mate heh. I will close this PR and submit one later today once I'm done from work. :+1:

MrJerB:
@NickCraver Finally got around to sorting this out.

NickCraver:
Merged in, thanks for fixing it up :)

PR #392, opened 2015-11-11
Add `System.Xml` framework assembly

henkmollema:
Without it, Resharper complains about needing the `System.Xml` module when using the `DataTable` class. Project compiles fine without it though. The .NET 4.0 .csproj file included it as well.

### DIFF
--- a/Dapper/project.json
+++ b/Dapper/project.json
@@ -17,13 +17,15 @@
         "warningsAsErrors": true
       },
       "frameworkAssemblies": {
-        "System.Data": "4.0.0.0"
+        "System.Data": "4.0.0.0",
+        "System.Xml": "4.0.0.0"
       }
     },
     "net40": {
       "compilationOptions": { "warningsAsErrors": true },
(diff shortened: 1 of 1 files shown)


henkmollema:
Not sure if we need this. Resharper errors are pretty annoying though. Any opinion on it?

NickCraver:
Brutal honesty: I can't get excited care about this one *at all*. Everything changes *again* next week on .Net Standard/Core anyway with the 18th announcement...so this (and much of the project) gets yet another overhaul anyway. We love it every time this happens, it's so much fun! For some definition of "fun".

I'm going to merge this simply because clicking merge here is the least effort for all parties involved :)

henkmollema:
Yeah, it's confusing. But actually this is about the framework references of the full .NET framework. `Dapper NET40.csproj` includes it as well:
```xml
<Reference Include="System.XML" />
```

Not sure why Resharper complains though. Thanks for merging anyways!

PR #402, opened 2015-11-19
Upgrading platform monikers

sebastienros:
Haven't run the unit tests myself, hope the CI is doing it, but builds and create the nuget packages correctly using `dnu pack --configuration release`

### DIFF
--- a/Dapper.DNX.Tests/Assert.cs
+++ b/Dapper.DNX.Tests/Assert.cs
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
-#if DNXCORE50
+#if DOTNET5_2
 using ApplicationException = System.InvalidOperationException;
 #endif
 
--- a/Dapper.DNX.Tests/Program.cs
+++ b/Dapper.DNX.Tests/Program.cs
@@ -71,7 +71,7 @@ namespace SqlMapper
 
         static void Main()
         {
-#if DNXCORE50
+#if DOTNET5_2
                 Console.WriteLine("CoreCLR");
 #else
                 Console.WriteLine(".NET: " + Environment.Version);
@@ -105,7 +105,7 @@ namespace SqlMapper
             RunPerformanceTests();
 #endif
 
-#if DNXCORE50
+#if DOTNET5_2
             Console.WriteLine("(end of tests; press return)");
             Console.ReadLine();
 #else
@@ -164,7 +164,7 @@ end
         }
         private static bool HasAttribute<T>(MemberInfo member) where T : Attribute
         {
-#if DNXCORE50
+#if DOTNET5_2
             return member.CustomAttributes.Any(x => x.AttributeType == typeof(T));
 #else
             return Attribute.IsDefined(member, typeof(T), true);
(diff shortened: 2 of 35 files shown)


sebastienros:
The CI needs to use rc1 instead of beta-8 I am not sure if you set it in the AppVeyor configuration on in a local build script. You'll know better.

NickCraver:
Checking the build - un momento

NickCraver:
Info found (though even the ASP.Net team doesn't use the new CLI tools...so we'll see how they use Appveyor as well) - poking another build to run here.

sebastienros:
Yeah !

NickCraver:
Appveyor updated - and builds passed. Yanking this in - I'll work on porting the unit tests to xUnit next so we can get them setup much easier in the build and with local runners.

Thanks for the work!

Disclaimer: I'm not sure this moniker structure is "correct" either - but I agree it's more correct than the current. I want to see if we can grab any of @davidfowl's time on this sometime this week or next to actually make this a concrete example of the proper setup.

sebastienros:
Well, @davidfowl was by my side when I did the PR, just pure chance, and I showed it to him for some feedback, he approved ;)

NickCraver:
Ah excellent - nevermind then! Much appreciated all around. We'll get a NuGet build up when @mgravell wakes across the pond - I'll crack at the unit tests and see if I can get them working proper with the new setup.

sebastienros:
I don't know how you intend to run the unit test on AppVeyor, but just in case you want to go with Sqlite, you need to know it's broken with Dapper right now on ~~dnx~~ dotnet until RC2, because Sqlite didn't like how you remove the `@` from the parameters. At least you can run these tests on deskptop clr.

NickCraver:
I have to admit - I'm still confused here. The TFMs in the annoucement here: https://github.com/aspnet/Announcements/issues/98 mention the style used here: `dotnet5.x

sebastienros:
I am confused, what is confusing you?

NickCraver:
Oh I see now, the work to make the tooling match the latest announcement (the `netstandard` monikers) isn't done yet:
> NOTE: The netstandard TFM does not work in packages as yet. It's still implemented as dotnet. Please follow the issues for more details and progress updates.

I (wrongly) assumed that'd be in place before the announcement and everyone jumping in :-/ Is there any ETA on that, since it'll be another round of changes? It's the way unit tests and such are picked up, so that'd be handy info.

davidfowl:
Working on it. 😀

sebastienros:
I used 1.2 because you were targeting `dnx451` previously and I didn't want to change that if that was on purpose. Also it compiled fine for me, and on the CI. I have core rc1-final locally. If you want to bump to 5.4 then it's up to you.

mgravell:
I'm very confused why it didn't compile locally - either in the IDE or via `dnu pack`. I'll try pushing a rc1 build that targets 5.2 to appveyor

mgravell:
The CI server seems to agree with me: https://ci.appveyor.com/project/StackExchange/dapper-dot-net-27k3h/build/1.0.103

NickCraver:
I also can't build anymore on 5.2...though I could when this was pushed. I think some packages have changed their builds underneath and this now breaks. I think SQL Client's port on CoreCLR needing new items that aren't currently in managed code will ultimately dictate that we stay on 5.4 anyway, so I'd say let's stick with this setup and see if we have any need for a build between...which I don't think we do.

PR #403, opened 2015-11-19
Port tests to xUnit

NickCraver:
This allows us to run the tests much faster and all from Visual Studio, command line, and Appveyor.

Possible TODO after this merge: Flag missing test methods. And sanity check the crazy that is project.json.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ NuGet.exe
 .docstats
 *.ide/
 *.lock.json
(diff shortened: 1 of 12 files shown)


NickCraver:
Side note: I started splitting up the massive unit test file, but there's a lot more to do here. Also note: the async tests have not been running due to defines unless I'm missing something - do we want to enable these now?

If so I'll fix those up.

PR #407, opened 2015-11-22
A full CoreCLR/DNX friendly conversion

NickCraver:
This eliminates the dual solutions and brings Contrib into the project.json fold.
- Tests are ported to xUnit though SQLite is disabled on DNX due to Issue #375 (Interop loads fail in the VS runner and  params are busted regardless).
- Projects are consolidated down to what's currently possible with project.json.
- Contrib moved to: `Dapper.Contrib` and `Dapper.Tests.Contrib` (so tests appear at the end, no other reason vs. `Dapper.Contrib.Tests`)
- EF moved to: `Dapper.EntityFramework(.StrongName)`
- Rainbow moved to: `Dapper.Rainbow`
- SqlBuilder moved to: `Dapper.SqlBuilder`
- Packages are removed since it's not current-tooling friendly to keep them (nor was it using them).
- xUnit implemented in Contrib via abstracts - this allow us to very quickly add any database provider to the rig in a very concise way.
- Contrib unit tests fixed in various places - they were not parallel-runner friendly
- Nuspecs removed, since they aren't used anymore (to eliminate confusion)
- Removed speed tests for libraries that aren't on current versions of .Net (what would we be comparing?)
- Perf tests moved to `Dapper.Tests` (consolidation from `Dapper.DNX.Tests`), but aren't activated yet (no define)

PR #410, opened 2015-11-24
Provided a name for MySql LAST_INSERT_ID call to allow the result to …

### LINKED ISSUE #404: Insert<T>() in MySQL always return 0
I have a problem about Insert(). When I use MSSQL for my DB. I inserted a record, this method return a long value, it is ID of new record. But in MySQL, I alway receive 0 from Insert method.

joncloud:
…map correclty.

Fixes #404


NickCraver:
This will blow up against #407, but I'll manual merge once @mgravell has a chance to review that huge changeset.

NickCraver:
Hey, all the checks pass! (but this wasn't tested on MySQL) - this is proper but I'll get MySQL testing hooked up to also confirm in an automated way as well. Thanks for the commit!

NickCraver:
For others finding this: yes I'm aware the entire setup here will fail horribly in any high-concurrency environment. I plan on changing these to use the proper returns from a single query once testing is in place and I can test locally as well.

PR #411, opened 2015-11-25
Make DefaultTypeMap public again

henkmollema:
It used to be public before. I use it to provide a fallback typemap when there is no mapping configured using [Dapper.FluentMap](https://github.com/henkmollema/Dapper-FluentMap).

Is there a particular reason it's not public anymore?

### DIFF
--- a/Dapper/DefaultTypeMap.cs
+++ b/Dapper/DefaultTypeMap.cs
@@ -8,7 +8,7 @@ namespace Dapper
     /// <summary>
     /// Represents default type mapping strategy used by Dapper
     /// </summary>
-    sealed class DefaultTypeMap : SqlMapper.ITypeMap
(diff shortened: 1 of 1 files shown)


mgravell:
It *probably* got made non-public during the file-split; let me check...

henkmollema:
It used to be partial and public:
```csharp
public partial class DefaultTypeMap
    {

    }
```
https://github.com/StackExchange/dapper-dot-net/blob/1.42/Dapper%20NET40/SqlMapper.cs

mgravell:
Confirmed: it was `public`, by this bit:

    public partial class DefaultTypeMap
    {

    }

Will merge

henkmollema:
Thanks! :+1:

PR #412, opened 2015-11-26
Add {Query|Read}FirstOrDefault methods with optimized paths

mgravell:
also removes redundant `return await` constructs (prefer to return the existing Task)

### DIFF
--- a/Dapper/SqlMapper.GridReader.Async.cs
+++ b/Dapper/SqlMapper.GridReader.Async.cs
@@ -33,6 +33,15 @@ namespace Dapper
             }
 
             /// <summary>
+            /// Read the first row of the next grid of results, returned as a dynamic object
+            /// </summary>
+            /// <remarks>Note: the row can be accessed via "dynamic", or by casting to an IDictionary&lt;string,object&gt;</remarks>
+            public Task<dynamic> ReadFirstOrDefaultAsync()
+            {
+                return ReadFirstOrDefaultAsyncImpl<dynamic>(typeof(DapperRow));
+            }
+
+            /// <summary>
             /// Read the next grid of results
             /// </summary>
             public Task<IEnumerable<object>> ReadAsync(Type type, bool buffered = true)
@@ -40,12 +49,30 @@ namespace Dapper
                 if (type == null) throw new ArgumentNullException(nameof(type));
                 return ReadAsyncImpl<object>(type, buffered);
             }
+
+            /// <summary>
+            /// Read the first row of the next grid of results
+            /// </summary>
+            public Task<object> ReadFirstOrDefaultAsync(Type type)
+            {
+                if (type == null) throw new ArgumentNullException(nameof(type));
+                return ReadFirstOrDefaultAsyncImpl<object>(type);
+            }
+
             /// <summary>
             /// Read the next grid of results
             /// </summary>
             public Task<IEnumerable<T>> ReadAsync<T>(bool buffered = true)
             {
                 return ReadAsyncImpl<T>(typeof(T), buffered);
+            }
+
+            /// <summary>
+            /// Read the first row of the next grid of results
+            /// </summary>
+            public Task<T> ReadFirstOrDefaultAsync<T>()
+            {
+                return ReadFirstOrDefaultAsyncImpl<T>(typeof(T));
             }
 
             private async Task NextResultAsync()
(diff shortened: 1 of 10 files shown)


Dapper/SqlMapper.cs
```diff
@@ -604,7 +612,21 @@ public static IEnumerable<dynamic> Query(this IDbConnection cnn, string sql, obj
...
+        /// Executes a single-row query, returning the data typed as per T
+        /// </summary>
+        /// <returns>A sequence of data of the supplied type; if a basic type (int, string, etc) is queried then the data from the first column in assumed, otherwise an instance is
+        /// created per row, and a direct column-name===member-name mapping is assumed (case insensitive).
+        /// </returns>
+        public static T QueryFirstOrDefault<T>(
+            this IDbConnection cnn, string sql, object param = null, IDbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null
+        )
+        {
+            var command = new CommandDefinition(sql, (object)param, transaction, commandTimeout, commandType, CommandFlags.None);
```
NickCraver: Extra (object) cast

Dapper/SqlMapper.GridReader.cs
```diff
@@ -75,6 +101,32 @@ private IEnumerable<T> ReadImpl<T>(Type type, bool buffered)
...
             }
 
+            private T ReadFirstOrDefaultImpl<T>(Type type)
+            {
+                if (reader == null) throw new ObjectDisposedException(GetType().FullName, "The reader has been disposed; this can happen after all data has been consumed");
+                if (IsConsumed) throw new InvalidOperationException("Query results must be consumed in the correct order, and each result can only be consumed once");
+                IsConsumed = true;
+
+                T result = default(T);
+                if(reader.Read())
```
NickCraver: nit: space

Dapper/SqlMapper.cs
```diff
@@ -619,6 +641,20 @@ public static IEnumerable<dynamic> Query(this IDbConnection cnn, string sql, obj
...
+        /// </summary>
+        /// <returns>A sequence of data of the supplied type; if a basic type (int, string, etc) is queried then the data from the first column in assumed, otherwise an instance is
+        /// created per row, and a direct column-name===member-name mapping is assumed (case insensitive).
+        /// </returns>
+        public static object QueryFirstOrDefault(
+            this IDbConnection cnn, Type type, string sql, object param = null, IDbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null
+        )
+        {
+            if (type == null) throw new ArgumentNullException(nameof(type));
+            var command = new CommandDefinition(sql, (object)param, transaction, commandTimeout, commandType,  CommandFlags.None);
```
NickCraver: extra cast

mgravell: yup; that dates back to when `param` was `dynamic`; can fix throughout

Dapper/SqlMapper.cs
```diff
@@ -588,9 +588,17 @@ public static IEnumerable<dynamic> Query(this IDbConnection cnn, string sql, obj
         }
 
         /// <summary>
+        /// Return a dynamic object with properties matching the columns
+        /// </summary>
+        /// <remarks>Note: the row can be accessed via "dynamic", or by casting to an IDictionary&lt;string,object&gt;</remarks>
+        public static dynamic QueryFirstOrDefault(this IDbConnection cnn, string sql, object param = null, IDbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
+        {
+            return QueryFirstOrDefault<DapperRow>(cnn, sql, param as object, transaction, commandTimeout, commandType);
```
NickCraver: same as others - is the `as` doing anything here?

Dapper/SqlMapper.GridReader.Async.cs
```diff
@@ -94,6 +121,42 @@ private Task<IEnumerable<T>> ReadAsyncImpl<T>(Type type, bool buffered)
                 }
             }
 
+            private Task<T> ReadFirstOrDefaultAsyncImpl<T>(Type type)
+            {
+                var dbReader = reader as DbDataReader;
+                if (dbReader != null) return ReadFirstOrDefaultAsyncImplImpl<T>(dbReader, type);
+
+                // no async API available; use non-async and fake it
+                return Task.FromResult<T>(ReadFirstOrDefaultImpl<T>(type));
```
NickCraver: just confirming intent: the entire goal is to centralize the throws? If so, +1

NickCraver:
Looks good, +1 - noted the only small (and can be fixed later) items I saw.

It's a shame we have to read through the entire result set if someone does `Select Top 500000 *`, but I also agree we shouldn't waste CPU (or a forced close) for every other case to handle whoever does that.

Dapper/SqlMapper.cs
mgravell: @NickCraver bleugh; you opened a can of worms there; fixed in additional commit

Dapper/SqlMapper.GridReader.cs
mgravell: cannot parse

NickCraver: if (reader.Read())
=P

Dapper/SqlMapper.GridReader.Async.cs
mgravell: "centralize the throws" - cannot parse; what throws?

mgravell: The reason this *specific* method exists is so that we don't have a redundant `return await SomethingThatReturnsATask();`, because that unnecessary wrapping annoys me

NickCraver: Since the reader is null going into `FromResult`, that code path is strictly for throwing the error I assumed - that's why I wanted to confirm your intent here vs. throwing. I'm cool with it - it should be the exceptional/rare case anyway and makes sense, but worth a sanity check is all.

PR #415, opened 2015-11-26
Use single-row/single-result hint when possible

mgravell:


NickCraver:
Looks good, merging.

mgravell:
Depending on the provider, this should solve the issue you commented on
earlier about having to pay bandwidth if the user returns 1000 rows when
asking for 1 :)

Issue #281, opened 2015-05-07
Invalid attempt to read from column ordinal n.  With CommandBehavior.SequentialAccess, you may only read from column ordinal x or greater


pranavmehta:
It seems like this issue was fixed in r1.31 [Issue #148] (https://github.com/StackExchange/dapper-dot-net/issues/148) but I'm still seeing this error in Dapper 1.4, though in a slightly different context. I'm trying to read off a datareader obtained via ExecuteReaderAsync.

Reading the column values in correct order, works fine, and does not generate this exception.

Apologies if this is a dupe or I misunderstood something. 
Thanks! 
PS: Big fan of Dapper

dotnetchris:
I just encountered this with ExecuteReaderAsync. The code in question worked fine with ExecuteReader but switching to async/await directly results in this exception. 

The sole changes were changing the original:

```c#
   using (var reader = connection.ExecuteReader(
```
to 

```c#
   using (var reader = await connection.ExecuteReaderAsync(
```

NickCraver:
PR #416 opened to resolve this, want @mgravell eyes on it in case there's some ancient reason I'm unaware of for the current behavior.

NickCraver:
This one's fixed, it'll be in the next release.

dotnetchris:
thanks @NickCraver !

PR #416, opened 2015-11-28
Fix for #281 - Change CommandBehavior on ExecuteReaderAsync to match sync version

NickCraver:
Issue #281 stems from differences between `ExecuteReader` and `ExecuteReaderAsync`. Though the overloads differ the basics are:

`ExecuteReaderAsync` -> `ExecuteReaderImplAsync` uses `CommandBehavior.SequentialAccess`
`ExecuteReader` -> `ExecuteReaderImpl` uses `CommandBehavior.Default`

There's also an overload in the middle where ExecuteReader allows specifying command behavior which doesn't exist on async...but I wouldn't add that until someone asks for it.

Doing this as a PR because I'm unaware of any reason this would be `SequentialAccess` and differ - I *assume* it's a historical artifact from a copy of `Query` code forever ago...but there may be a reason, so this gets some extra eyes.

### DIFF
--- a/Dapper.Tests/Tests.Async.cs
+++ b/Dapper.Tests/Tests.Async.cs
@@ -734,6 +734,19 @@ SET @AddressPersonId = @PersonId", p))
                 ", new Product { Id = 1 })).Single();
             id.IsEqualTo(2);
         }
+
+        [Fact]
+        public async Task Issue1281_DataReaderOutOfOrderAsync()
+        {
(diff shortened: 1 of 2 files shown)


mgravell:
Agree; if we're passing the reader out to an external consumer, we shouldn't be specifying sequential - the library guarantees that internally, but that is not a good assumption externally.

Issue #418, opened 2015-11-30
explicitKey bug


itlizhi:
table using [explicitKey],first [update] then [insert],  the [ExplicitKey] not insert.
guess： [update<T>]  keyProperties.AddRange(explicitKeyProperties);

johandanforth:
Can you plese provide some more info @itlizhi ? What database platform? Can you give some sample code to reproduce the error?

itlizhi:
using sqlserver 2008 r2
![20151201095158](https://cloud.githubusercontent.com/assets/10509513/11490193/43b7449e-9811-11e5-955a-10507656cd07.png)

johandanforth:
Which version of Dapper and Dapper.Contrib? I actually got the same error as you with the current Nuget version 1.44. I'll see if I can find the same bug in the current source. Might be a bug actually.

itlizhi:
<package id="Dapper" version="1.42" targetFramework="net45" />
 <package id="Dapper.Contrib" version="1.44" targetFramework="net45" />

johandanforth:
Thanks @itlizhi I think I found the problem yesterday and I will look at a fix this weekend. For some reason the `ExplicitKey` property is cached as an ordinary key in the `Update` extension method, which is wrong. That is why it works if `Insert` is called first. But I just had a glance at it yesterday, I'll look at it on saturday or sunday, unless someone beats me to it. It's easy to test and reproduce.

PR #420, opened 2015-12-06
Fix and test for issue #418

johandanforth:


NickCraver:
Fixed unit tests locally, but need to see if this is the proper fix - lemme wake up and take another look before merging in :)

NickCraver:
Merged this in, but going to re-work it as well. We can do a combined properties cache *once* rather than on each access, but I honestly question whether explicit keys and regular keys can be combined in another way further up - will take a look when time allows.

johandanforth:
I agree with you, this was kind of a quick-fix. I'm very busy atm with customer work and wish I could spend some more timer on dapper, but you know how it is...

PR #422, opened 2015-12-16
Snapshotter now supports private properties.

BlackjacketMack:
Private properties can serve as helpful passthroughs to more complex
public properties.  SqlMapper honors private properties when mapping
properties, so it seems natural for Snapshotter to follow suit on the
'get' side of things within the Dapper ecosystem.


mgravell:
Makes sense to me; thanks

PR #431, opened 2015-12-30
Added Dapper.snk signing to Dapper.Contrib

alexazarh:
I added signing to Dapper.Contrib


mgravell:
This is a breaking change. This type of thing should *only* be done with a major rev bump.

alexazarh:
To which version should I bump it?
Btw,
Latest stable version of Dapper.Contrib is 1.43 in nuget but in github I see only 1.42.

NickCraver:
It shouldn't be bumped. *If* this is desired, it'd coincide with a 2.0 release of Dapper and associated libs.

mgravell:
Additional: we didn't even bump core for this: we added a second package.
It may be worth thinking what milestones would constitute a major release,
and drop everything to one (strong named) package at the same time.

alexazarh:
We want to use the latest stable version and we need it to be strongly named. We care less about whether the major, minor, build or revision is bumped. 

So its up to you guys to decide whats the best and fastest way to release a strong named version for the last stable version :smiley: 

Addition: I think its a good idea to combine everything into one strong named package.

mgravell:
Can I ask: what is the motivation for needing it strongly named? And by
that I don't just mean "because our app is strongly named" - I mean further
up. I'm always intrigued what the active reasons for this are. I have a
mental list of such scenarios, but: it is a very short list, and I always
like to gather more inputs...

alexazarh:
Our application is an enterprise one and is installed on premise. So it must be strongly named.

devlead:
@alexazarh meanwhile perhaps [StrongNameSigner](https://github.com/brutaldev/StrongNameSigner/) is an option.

mgravell:
I fail to see how that helps. If someone has enough access to the machine
to swap files, they can completely disable strong name checking *anyway*.
Strong naming should never be considered a security feature.

Ah, good ol' strong names: the gift that gives on keeping :)

mgravell:
Gives on keeping? Strewth. I meant: keeps on giving.

Note; I have had the snk conversation enough times that I don't disagree
there should be a snk version, *just to stop the noise*. I'm not suggesting
we shouldn't do it (one way or another).

NickCraver:
@alexazarh Are you strong naming just to do so, or because someone who sets policy thinks it adds security in some way? (it does not). If you're just trying to satisfy a requirement, have you looked at StrongNamer? https://github.com/dsplaisted/strongnamer

We'll revisit signing with 2.0, though.

alexazarh:
@NickCraver I kind of satisfying a requirement. In any case, I ended up implementing my solution with a different ORM.

PR #444, opened 2016-01-27
Created additional Query method overloads for more types mapping.

malek8:
I needed to map more than 7 types using Query method, so I created additional overloads from 8 to 12 more types.


NickCraver:
Due to the maintenance cost this brings and the low percentage use case it supports - we are not really for putting these overloads in the core library. You can read the comments on a previous very-similar PR here: https://github.com/StackExchange/dapper-dot-net/pull/245

PR #454, opened 2016-02-09
Target net451 rather than dnx451

henkmollema:
Class libraries should target `net451` rather than `dnx451`, according to this: https://github.com/aspnet/Announcements/issues/98

Also updated global.json to use DNX v1.0.0-rc1-update1

### DIFF
--- a/Dapper.StrongName/project.json
+++ b/Dapper.StrongName/project.json
@@ -33,7 +33,7 @@
         "System.Data": "4.0.0.0"
       }
     },
-    "dnx451": {
+    "net451": {
       "frameworkAssemblies": {
         "System.Data": "4.0.0.0",
         "System.Xml": "4.0.0.0"
--- a/Dapper/project.json
+++ b/Dapper/project.json
@@ -31,7 +31,7 @@
         "System.Xml": "4.0.0.0"
       }
     },
-    "dnx451": {
+    "net451": {
       "frameworkAssemblies": {
         "System.Data": "4.0.0.0",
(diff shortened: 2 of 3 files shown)


mgravell:
I love the irony of "upgrading" *from* "final" to "update1" ;p But that's nothing to do with this! Sounds good; will check it in the morning.

NickCraver:
I would have accepted this earlier but need to make sure xunit still runs tests correctly in VS, that's the only landmine I'm aware of here.

mgravell:
Yeah, that's what I was going to test for in particular; last time I checked: libs can use monikers, but test rigs need to cite explicit platforms.

mgravell:
Merged, thanks. I have it working with "dnx test", but "dotnet run" is still very unhappy - not sure that is related. I added a few other fixes related to tooling (not a criticism of this pull request), most notably this one (@NickCraver you'll like this): https://github.com/dotnet/cli/issues/1342

NickCraver:
Hah - yeah that relative pathing is going to cause issues for many other projects too :-/

henkmollema:
Yeah, `dnx451` and `dnxcore50` are for apps (console, web, test) and `net451` and `dotnet5.4` for class libs respectively. They are indeed compatible.

mgravell:
>  Perhaps will send a new PR by then.

When RC2 drops, I'll be visiting just about every package I have touched, so... we'll probably handle that one by default. If we somehow forget or screw it up, though, then sure.

Issue #433, opened 2016-01-06
Use TypeHandler for all types


chilversc:
Currently there is a separation between built-in types and custom types that use `ITypeHandler`.
There seems to be a common theme of bugs/enhancements that all revolve around this separation.

* #206 TypeHandler<> can't override default typeMap for base types
* #275 TypeHandler not working correctly with DynamicParameters
* #259 Custom type handlers for enums are ignored
* #224 Non-nullable TypeHandler passed DBNull when selecting scalar results

I was thinking its worth investigating if the type handling could be unified, so that both the in-built types and custom type handlers use the same mechanism so that the code handling types can be simplified in to, look up type handler, check for an optimizer, otherwise just use the type handler.

This would need to introduce a new enhanced type handler that can optionally generate IL to convert the to/from the database value. This would then allow custom type handlers to also generate optimized IL to convert the value.

This shouldn't have any significant impact on query times as type handlers are only looked up once when generating the IL for the query.

vdaron:
+1

NickCraver:
I honestly don't know if I'm for or against this, I'll just note why we haven't dedicated much time to it: it's a potentially breaking change, so it'll have to wait for a 2.0 release.

That being said, we're all for someone doing the digging on options as hours are the most limited resource around here - it'd be much appreciated.

chilversc:
Another case this could help with is #213 as it would allow the user to replace the in-built handling for strings.

kwaclaw:
Would also help with #303, #471, #607.

NickCraver:
Tagging for the v2 release.

vitidev:
And that's bad. Already 2019, but the problem is not solved. 
Does anyone understand why it is impossible to solve the problem with this [pullrequest ](https://github.com/StackExchange/Dapper/pull/471) and by adding a flag like

`SqlMapper.OverrideHandlersForStandardTypes = true;`

so that the change is no longer breaking
 and not wait Dapper 2.0 for many years

PR #458, opened 2016-02-11
Custom type handlers for enums are ignored

### LINKED ISSUE #259: Custom type handlers for enums are ignored
I have an enum that is serialized to the DB using a single letter code.
I tried adding a custom type handler for the enum to handle the conversion but dapper seems to ignore the custom type handler.

Given the example code I expected to see the values;
ID = 10, Type = B
ID = 1, Type = Foo

What I actually got was:
ID = 1, Type = 2
DataException, Error parsing column 1 (Type=F - String)

---

```csharp
void Main()
{
    SqlMapper.AddTypeHandler(new ItemTypeHandler());
    using (var connection = new SqlConnection("Server=(local); Database=tempdb; Integrated Security=SSPI")) {
        connection.Open();
        connection.Query ("SELECT @ID AS ID, @Type AS Type", new Item {ID = 10, Type = ItemType.Bar}).Dump();
        connection.Query<Item> ("SELECT 1 AS ID, 'F' AS Type").Dump();
    }
}

class ItemTypeHandler : SqlMapper.TypeHandler<ItemType>
{
    public override ItemType Parse(object value)
    {
        var c = ((string) value) [0];
        switch (c) {
            case 'F': return ItemType.Foo;
            case 'B': return ItemType.Bar;
            default: throw new ArgumentOutOfRangeException();
        }
    }
    
    public override void SetValue(IDbDataParameter p, ItemType value)
    {
        p.DbType = DbType.AnsiStringFixedLength;
        p.Size = 1;
        
        if (value == ItemType.Foo)
            p.Value = "F";
        else if (value == ItemType.Bar)
            p.Value = "B";
        else
            throw new ArgumentOutOfRangeException();
    }
}

class Item
{
    public long ID { get; set; }
    public ItemType Type { get; set; }
}

enum ItemType
{
    Foo = 1,
    Bar = 2,
}
```

chilversc:
I gave highest precedence to custom type handlers. This aims to fix #259 by ensuring if a custom type handler is registered the custom type handler is always used instead of any built-in handling.

### DIFF
--- a/Dapper.Tests/Tests.Enums.cs
+++ b/Dapper.Tests/Tests.Enums.cs
@@ -56,6 +56,23 @@ namespace Dapper.Tests
             obj.B.IsEqualTo(EnumParam.B);
             obj.C.IsEqualTo((EnumParam)0);
         }
+
+        [Fact]
+        public void TestEnumParameterUsesTypeMapper()
+        {
+            SqlMapper.AddTypeHandler(ItemTypeHandler.Default);
+            var result = connection.QuerySingle<string>("SELECT @Foo", new {Item.Foo});
+            "F".IsEqualTo(result);
+        }
+
+        [Fact]
+        public void TestEnumResultUsesTypeMapper()
+        {
+            SqlMapper.AddTypeHandler(ItemTypeHandler.Default);
+            var result = connection.QuerySingle<Item>("SELECT 'F'");
+            Item.Foo.IsEqualTo(result);
+        }
+
         enum EnumParam : short
         {
             None, A, B
@@ -87,6 +104,46 @@ namespace Dapper.Tests
         class TestEnumClassNoNull
         {
             public TestEnum EnumEnum { get; set; }
+        }
+
+        enum Item
+        {
+            None,
+            Foo,
+            Bar
+        }
+
+        class ItemTypeHandler : SqlMapper.TypeHandler<Item>
+        {
+            public static readonly ItemTypeHandler Default = new ItemTypeHandler();
+
+            public override Item Parse(object value)
+            {
+                var c = ((string) value)[0];
+                switch (c)
+                {
+                    case 'F': return Item.Foo;
+                    case 'B': return Item.Bar;
+                    default: throw new ArgumentOutOfRangeException();
+                }
+            }
+
+            public override void SetValue(IDbDataParameter parameter, Item value)
+            {
+                parameter.DbType = DbType.AnsiStringFixedLength;
+                parameter.Size = 1;
+                parameter.Value = Format(value);
+            }
+
+            private string Format(Item value)
+            {
+                switch (value)
+                {
+                    case Item.Foo: return "F";
+                    case Item.Bar: return "B";
+                    default: throw new ArgumentOutOfRangeException();
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


NickCraver:
While this may fix #259, it can very easily be a very quiet breaking change for many other use cases. I'd recommend we not merge anything like this until a major release.

chilversc:
Yeah, I couldn't see any way to avoid that. Could add a flag to enable the new behaviour but not sure how you'd feel about that.

NickCraver:
Will need to do lots of scenario testing here, but I don't disagree with the change itself - I think consistent behavior is probably the best option given the confusion of the current one. It just has to wait for a major release - I don't think a switch is really worth the complication it brings here. Will wait to see what @mgravell thinks though.

peterthomastlc:
Wouldn't it be safer and clearer (in LookupDbType) just to change the line
`handler = null;`
to
`typeHandlers.TryGetValue(type, out handler);`

This would ensure custom TypeHandlers are called for built-in types and enums, whilst minimising side effects.

chilversc:
@peterthomastlc the problem is it changes behaviour for existing apps. If an existing app had a custom type handler for and enum before this change the handler wouldn't have been taking effect. If they rely on that behaviour after the change they would start getting an incorrect result. At best it would crash, worse you'd get data corruption.

This would affect more than just enums as well since it moves custom type handlers to a higher priority than other in-built checks such as type maps and IEnumerable.

As such the only viable option I can see is a flag to set the custom type handler's priority if this change were to be introduced. Though that has the downside of increasing the complexity of the API for users and the complexity when it comes to maintenance and testing (as now you have 2 paths to test depending on this flag).

This is only one issue surrounding type handling that makes me think a better model could be added instead of hacks to support this one specific issue. Then custom and in-built could all use the same pattern, and so that custom could optionally support their own IL generation to optimise get/set as per #433. Though such a change wouldn't be trivial (especially when handling backward compatibility with the old model).

bladeoflight16:
@chilversc So wait, you're suggesting that someone might have:

1. Had fully working mapping code out of the box.
1. Decided to write a custom type mapper, even though the out of the box functionality was working.
1. Added code to register the type mapper.
1. Observed that the type mapper does *nothing*.
1. Relied on their mapper code being ineffective and dead, so that the built in mapping functionality would be invoked, for the the rest of the app to continue working.

Now, I'm not saying it's impossible. It could happen. But don't you think that this would be both exceedingly rare and extremely silly? I cannot fathom a case in which someone would even create a type mapper in the first place if they rely on the current behavior because relying on the current behavior necessarily implies that Dapper's mapping code is just working for them out of the box, doesn't it?

rhubley:
@bladeoflight16 The more likely scenario would be 

1. writing a custom type mapper
2. custom type mapper does not work
3. change code to work with Dapper out of the box
4. left Custom type mapper include, because it wasn't breaking tests

I mean who hasn't tried fixing something 9 different ways and left tries 1-8 in the code because maybe you needed 1 or 2 of those changes as well as the final fix.

bladeoflight16:
If you don't know what changes you made fixed the bug, you don't know whether it's fixed or not.

phillip-haydon:
This is where OSS falls down. Major bug in software, 6 years and no fix merged.

djimmo:
Maybe this helps. I combined some of the ideas I found online into something that currently works

https://gist.github.com/djimmo/b54a1643f15bd4e54303992ddce4a968

PR #459, opened 2016-02-13
actual/expected in asserts are the wrong way round

chilversc:
All the tests are wrote in the form `actual.IsEqualTo(expected)`. This
made for some odd error messages when a test failed.


mgravell:
Thanks

PR #487, opened 2016-03-26
Fix postgres column name quoting when RETURNING is specified

DeathCradle:
It appears as if the PostgresAdapter encapsulates all column names in quotes, just not the ones that are specified with the RETURNING clause.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -820,7 +820,7 @@ public partial class PostgresAdapter : ISqlAdapter
         var id = 0;
         foreach (var p in propertyInfos)
         {
-            var value = ((IDictionary<string, object>)results.First())[p.Name.ToLower()];
+            var value = ((IDictionary<string, object>)results.First()).Single(x => x.Key.ToLower() == p.Name.ToLower()).Value;
             p.SetValue(entityToInsert, value, null);
             if (id == 0)
                 id = Convert.ToInt32(value);
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -810,7 +810,7 @@ public partial class PostgresAdapter : ISqlAdapter
                 if (!first)
                     sb.Append(", ");
                 first = false;
-                sb.Append(property.Name);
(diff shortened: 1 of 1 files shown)


mgravell:
Makes sense, but: can I check the intent of the manual dictionary crawl
there? Is this targeting a specific scenario? (I think I can see the
edge-case, but it would be good to be explicit)

DeathCradle:
My particular issue was due to having table columns pascal case (e.g. `Id`), rather than in postgres convention of lowercase_with_underscores (from what I read - i am actually new to postgres and dapper).

Due to the pascal case the `Id` column was not being found in the result dictionary due to `p.Name.ToLower()`. I wasn't confident that removing `.ToLower()` wouldn't cause issues, so I compare both as lower case in order to try to cover all scenarios.
If it is in fact as simple as removing the `.ToLower()`, then great!

mgravell:
I need to look at it on a computer (not a 'phone) to see the full context,
but foreach over a dictionary is usually a bad sign. I'll have a look,
though.

DeathCradle:
Yeah that's understandable.
In order to try to assist more, it seems in a postgres query an unquoted column such as `Id` will be automatically lowered to `id` (which is a reason why I wasn't confident with the `.ToLower()` removal). 
However, the change to use `AppendColumnName` will ensure the column is in the same case as the property, (note, since my columns were pascal cased I was receiving: `{"42703: column \"id\" does not exist"}`).

Now since the column name will be the same as the property name; the removal of `.ToLower()` should be fine. Perhaps the `.ToLower()` was there as the `RETURNING` clause was not quoted and postgres lowered column names (which doesn't work with pascal case table columns).
`var value = ((IDictionary<string, object>)results.First())[p.Name];`.

I'm sure you'll look at it further, but I hope that helps/makes sense.

hey-red:
Merge this pls!

GrayDelacluyse:
Bump, this fixed my issue as well.

NickCraver:
Can you please take a look at #722 for V2? It's meant to address all of these cases in a global way, I hope.

NickCraver:
Closing out to merge into #722, since we need a global solution here - currently either direction hoses some users. This case-insensitive match (effectively) can break users today who have 2 columns with different casing...we've got to fix it with something that works everywhere. Please look at #722 for that plan.

PR #495, opened 2016-04-04
Bumping dependencies version numbers

sebastienros:
This is necessary to build using the latest dotnet CLI.


mgravell:
Looks good, thanks. I will need to update build tools etc, and get a new build out.

PR #510, opened 2016-04-24
Tweaked the performance tests to return more than one record and display the output in a better manner

ewassef:
The tests needed to be tweaked to better test the product and compare the results.

1- Modified EF to conform to best practices for performance
2- Returned more than one result set to map into
3- Changed the output to a timespan instead of elapsed milliseconds in the display


ewassef:
@mgravell What do you think?

mgravell:
Sorry, I've had an insanely hectic couple of weeks. Will get to it.

ewassef:
Conflicts resolved in https://github.com/StackExchange/dapper-dot-net/pull/521

PR #512, opened 2016-04-28
Fixed #497 by exposing SqlMapperExtensions.AdapterDictionary as public.

### LINKED ISSUE #497: Allow custom ISqlAdapter
This is related to #336 - it would be useful if we could just plug in our own SqlAdapters. 

Use case - now that all the column names are quoted, column name case-insensitivity disappears in PostgreSQL (I think MySQL too). For some that may be good, others bad; rather than messing around decorating POCOs, just allow the ability to inject our own SqlAdapter. 

For the fix for the above, it's just a matter of changing the `AppendColumnName `and `AppendColumnNameEqualsValue` methods. But I have to fork the whole thing to do that. Others might want to plug in a more complex mapper.

It is possible to inject a `GetDatabaseType` custom callback, but it would be really useful too to have a `GetDatabaseFormatter` too.

Thank you.

joncloud:
Updated AdapterDictionary initialization to use 6.0 syntax.
Updated all implementations of ISqlAdapter to expose virtualized members.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -317,7 +317,7 @@ public partial interface ISqlAdapter
 
 public partial class SqlServerAdapter
 {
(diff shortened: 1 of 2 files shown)


marthai:
STREQ : is equal to the string

NickCraver:

PR #532, opened 2016-06-11
Trim package dependencies

henkmollema:
Trimmed the package dependencies by not referencing `NETStandard.Libary` and the kitchen sink but in stead reference each required package individually. This reduces the package footprint of Dapper.

Reference: https://dotnet.github.io/docs/scenarios/library-authoring/trimming.html

--

I'm not sure if you are interested in this. But since Dapper is all about performance, I figured I give it a shot.

### DIFF
--- a/Dapper/project.json
+++ b/Dapper/project.json
@@ -55,14 +55,24 @@
         "define": [ "ASYNC", "COREFX" ]
       },
       "dependencies": {
-        "NETStandard.Library": "1.5.0-rc2-24027",
+        "Microsoft.NETCore.Platforms": "1.0.1-rc2-24027",
+        "Microsoft.NETCore.Runtime": "1.0.2-rc2-24027",
+        "System.Collections": "4.0.11-rc2-24027",
+        "System.Collections.Concurrent": "4.0.12-rc2-24027",
+        "System.Collections.NonGeneric": "4.0.1-rc2-24027",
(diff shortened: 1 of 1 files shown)


NickCraver:
I hesitate to change this because the standard library is changing to where *the tooling* prunes instead. The goal there is a very simple configuration and the most efficient footprint as well. Think about how the Mono linker works and it likely becomes a whole lot clearer where things are headed.

Given all the packages are setup and working correctly at the moment, I'm inclined to wait and see what Microsoft does here. I'm having a chat with the .NET guys on Tuesday about upcoming stuff, I'll bring this up specifically then update here soon as I can.

henkmollema:
>  the standard library is changing to where the tooling prunes instead.

I'm not sure what you mean. Could you elaborate on that?

In general I understand your point. I was reading about trimming dependencies and thought about doing that for my Dapper extensions. However, if Dapper doesn't trim them, trimming my extensions doesn't really makes sense anyway.

I'm looking forward to the chat about these things. Would like it if you'd share the stuff. 😄

henkmollema:
Any update on this @NickCraver?

NickCraver:
@henkmollema Sorry! I should have sucked this in before RTM but I forgot it existed (it would have saved me a half hour...doh). The package dependencies were trimmed as part of the 1.50-rc3 release though, for the RTM references.

henkmollema:
Allright! :+1:

PR #533, opened 2016-06-11
Fixed #530 using technique that fixed #418 for async code.

### LINKED ISSUE #418: explicitKey bug

### LINKED ISSUE #530: Fix #418 for async code too
It seems like that when #418 was fixed, it was missed that it needs to be fixed in the file SqlMapperExtensions.Async.cs too.

@johandanforth commited the fix for the non async file.

joncloud:


cremor:
While that solution works, wouldn't it be cleaner to have the ToList call already in the KeyPropertiesCache method? That way it wouldn't be possible to forget it in callers.

Of course that solution would lead to more - maybe unnecessary - copies. But it seems like the various xyzPropertiesCache methods already return list copies when the requested type is already in the cache. Only if the cache is newly build the original list is returned (except for TypePropertiesCache that already returns a copy in both cases).

joncloud:
It looks like this was solved and merged with #1048.

PR #538, opened 2016-06-16
Fixed some syntax errors in project.json files

cremor:
Similar to 89908e0f028dcbc7fcf255f3716bb9bb695f438a


NickCraver:
Thanks! I'm not actually sure how this is working at all...

PR #539, opened 2016-06-16
Created the Dapper.Contrib.StrongName project

### LINKED ISSUE #536: Create a Dapper.Contrib.StrongName NuGet package
There are already strong name packages for the base Dapper package and the Entity Framework extensions. Could we please also get a strong name package for the Contrib extensions?

cremor:
This fixes #536.


NickCraver:
We want to combine all libraries into a single strongly named default in 2.0 (which we want to do sooner than later) rather than making 2 packages. For example, Dapper's main https://www.nuget.org/packages/Dapper.StrongName/ will disappear and https://www.nuget.org/packages/Dapper/ will be strongly named itself in v2, since this is a breaking change.

NickCraver:
Going to close this out, since this will break with the `.csproj` move - we'll make sure Contrib has a strong name in the v2 release, rather than confusing people with another package now.

PR #554, opened 2016-07-11
fix invalid condition used in QueryFirst and QuerySingle

neuecc:
I think `(row & Row.Single) != 0` is row is not Row.First so should swap condition.

By the way, in MySQL if use CommandBehavior.SingleRow, MySQL Connector execute `SET SQL_SELECT_LIMIT=1` on each query. (at command.cs - HandleCommandBeehaviors)
It causes performance issue so I need disable command behavior option...

### DIFF
--- a/Dapper/SqlMapper.Async.cs
+++ b/Dapper/SqlMapper.Async.cs
@@ -272,7 +272,7 @@ namespace Dapper
                 try
                 {
                     if (wasClosed) await ((DbConnection)cnn).OpenAsync(cancel).ConfigureAwait(false);
-                    reader = await ExecuteReaderWithFlagsFallbackAsync(cmd, wasClosed, (row & Row.Single) != 0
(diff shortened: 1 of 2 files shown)


mgravell:
I think the code is correct as written.

`(row & Row.Single) != 0` is `true` for `Single` and `SingleOrDefault`. In those cases we have to **read multiple rows** in order to validate that we have actually only got one, hence **not** passing `SingleRow`.

Conversely, `(row & Row.Single) != 0` is `false`  for `First`  and `FirstOrDefault`. In that scenario we **can** use the "single row" optimization, precisely because  we don't care whether multiple rows would  have otherwise been returned.

mgravell:
Re the performance thing; we're discussing options as to whether we should disable those flags.

mgravell:
Closing; I think the code change is invalid (please let me know if I've misunderstood - it happens); the performance thing should be addressed by #602

Issue #555, opened 2016-07-11
Dapper cancelling only ExecuteReaderAsync not Reader.ReadAsync with MultiMap


SelvinPL:
Dapper will fail with such Test

```
        async Task<IEnumerable<Product>> Test()
        {
            CancellationTokenSource cancel = new CancellationTokenSource();
                try
                {
                    const string sql = @"select 1 as id, 'abc' as name, 2 as id, 'def' as name
union all select 1 as id, 'abc' as name, 2 as id, 'def' as name";
                    var productQuery = await connection.QueryAsync<Product, Category, Product>(new CommandDefinition(sql, cancellationToken: cancel.Token), (p, c) => {
                        p.Category = c;
                        //we cancel it in the middle of reading
                        cancel.Cancel();
                        return p;
                    }).ConfigureAwait(false);
                    return productQuery;
                }
                catch (Exception agg)
                {
                    //should be SqlException with message canceled
                }
            return null;
        }

        public async Task TestCancelInReaderReadAsync()
        {
            var res = await Test();
            res.IsEqualTo(null);
        }
```

solution:
replace all code like(in SQlMapper.Async.cs):

```
using (var cmd = (DbCommand)command.SetupCommand(cnn, info.ParamReader))
{
/*...*/
}
```

with

```
using (var cmd = (DbCommand)command.SetupCommand(cnn, info.ParamReader))
{
     using (var reg = command.CancellationToken.Register(() => cmd.Cancel()))
     {
/*...*/
     }
}
```

solution is simillar to [this answer on SO](http://stackoverflow.com/questions/25540793/cancellation-token-on-dapper/34981640#34981640) but he is doing it at wrong place (and not dissposing the registration) ...

SelvinPL:
I do some more investigation ... In fact it is because QueryAsync with multimap is using MultiMapImpl which doesn not use reader.ReadAsync() but reader.Read()

SelvinPL:
Any chances for fixing this? 
I'm using Dapper in destop UI app... There is a filter where user can type something then I execute the query with given filter... 
If user type there something more I would like cancel query and do new query with new filter - I don't wana "download" unnecesry data from the server
I've stuck with 1.50.x with my modification ...

PR #558, opened 2016-07-12
Temporary fix for #555

SelvinPL:
Registering CancellationToken with DbCommand.Cancel will allow
cancellation even with synchrouns DbReader.Read


NickCraver:
First, thank you for raising this up and identifying the areas. Unfortunately, this one is a bit more nuanced, and required breaking API changes to handle before. But _some_ of that is being addressed in #1590 - the approach there is to opt-in via `CommandBehavior`, but in vNext we want to globally address this across all APIs and _really_ solve it for all cases. Marc's doing some work there now, but it's all generators so PRs against current code won't translate well before vNext.

PR #575, opened 2016-07-27
Small changes

MartLegion:


NickCraver:
I don't like to be blunt here, but I do see these changes all the time. Let me provide some context: while these appear as "small changes" they actually actively harm other pull requests in queue by (IMO, needlessly) creating a *lot* of merge conflicts (note the 42 files involved), creating more work for everyone. I would vote to close this based just on that reason (the same reason we haven't done such cleanup with pending pull requests either).

As an aside, note changes like this also serve only to create extra git history:

```diff
-                {"mysqlconnection", new MySqlAdapter()},
+                {"mysqlconnection", new MySqlAdapter()}
```

C# allows a comma at the end for code generation, but as a practice it also cleans up source control history by not having 2 line changes every time you add 1 item. For example, when you `git blame` the line, you get only the 1 relevant commit.

MartLegion:
Closed as "request"

PR #577, opened 2016-07-30
Fix #573 - OrWhere yielding AndWhere statement

### LINKED ISSUE #573: SqlBuilder OR clause
The SqlBuilder seems to be adding "OR" clause the wrong way...

are adding "AND" instead of "OR".

```
builder.Where("MyField = @Param",paramObj);
builder.OrWhere("MyField2 = @Param2",paramObj2); 
```

results this statement:
`WHERE MyField = @Nome AND MyField2 = @Param2`

Thanks

liamsymonds:
Fixes #573 where calling the SqlBuilder.OrWhere statement was actually yielding an 'AND WHERE' string.


NickCraver:
Thanks!

PR #593, opened 2016-08-29
missing xunit dependencies in project.json

mjashanks:
when running tests from Coreclr CLI ...i.e. `dotnet test` ... tests would not run due to error _No executable found matching command "dotnet-test-xunit"_


NickCraver:
Isn't the `"xunit": "2.1.0",` extra? It's already included for all frameworks up top.

mjashanks:
Yes, tested it - you're right - thanks!

 updated on that branch.

mjashanks:
implemented in other commits

Issue #596, opened 2016-09-01
Unbuffered QueryAsnyc multi-mapping throws an InvalidOperationException 


mburbea:
``` csharp
using (var conn = new SqlConnection(appToken.ConnectionString))
{
    conn.Open();
    var ret = conn.Query("select a='dog',b=3", (string a, int b) => a + b, splitOn: "b", buffered: false);
    Console.WriteLine(string.Join("", ret));
    ret = conn.QueryAsync("select a='dog',b=3", (string a, int b) => a + b, splitOn: "b", buffered: false).Result;
    Console.WriteLine(string.Join("", ret));
}
```

The call to Query returns `dog3`, the QueryAsync throws the following exception:

```
System.InvalidOperationException was unhandled
  HResult=-2146233079
  Message=Invalid attempt to call FieldCount when reader is closed.
  Source=System.Data
  StackTrace:
       at System.Data.SqlClient.SqlDataReader.get_FieldCount()
       at Dapper.SqlMapper.GetColumnHash(IDataReader reader, Int32 startBound, Int32 length)
       at Dapper.SqlMapper.<MultiMapImpl>d__140`8.MoveNext()
       at System.String.Join(String separator, IEnumerable`1 values)
       at Program.Main(String[] args) in 
```

The store procedure I am calling takes a while to execute and I'd rather let it be called async before I start streaming data out of the query.

mburbea:
Figured out the cause. 
https://github.com/StackExchange/dapper-dot-net/blob/master/Dapper/SqlMapper.Async.cs#L650
The error is that it returns the enumerable, but the using block closes the connection before enumeration begins.
```csharp
                using (var reader = await ExecuteReaderWithFlagsFallbackAsync(cmd, wasClosed, CommandBehavior.SequentialAccess | CommandBehavior.SingleResult, command.CancellationToken).ConfigureAwait(false))
                {
                    if (!command.Buffered) wasClosed = false; // handing back open reader; rely on command-behavior
                    var results = MultiMapImpl<TFirst, TSecond, TThird, TFourth, TFifth, TSixth, TSeventh, TReturn>(null, CommandDefinition.ForCallback(command.Parameters), map, splitOn, reader, identity, true);
                    return command.Buffered ? results.ToList() : results;
                }
```
I'll spin up a PR to fix.

RodrigoMaf:
Read about and use **async/await**

Ex.: 
using (var conn = new SqlConnection(appToken.ConnectionString))
{
[...]

PR #597, opened 2016-09-01
Allow multi-map async queries to be unbuffered

mburbea:
See #596 for more detail. Originally, the code would dispose the
dataReader that it opened before enumeration would begin. Now when you
dispose the enumerator, it will also close the command, which is line
with other behavior.
Buffered behavior remains consistent. Has no impact on sync version.


mburbea:
@mgravell , I'll fix the merge conflict but is there any object to merging this?

mgravell:
Needs a test/usage-illustration, but  other than that: looks good

NickCraver:
Any update on this?

NickCraver:
@mburbea Any update on example usage/test for this?

mburbea:
@NickCraver, hey sorry it's been a while since I looked at this issue  I'm not currently working on a project that uses the dapper library (that was back in september) so I have not had time to look further into this. If you rather I close the issue. I can do that.

Karillion:
Would be great to see this fixed in the current version of Dapper rather than 2,0. That seems to be a way off still.

bungeemonkee:
Is there a reason this can't get merged (or otherwise fixed) until V2? I'm working on something at the moment that uses only async calls and could really benefit from disabling buffering in several places.

Issue #484, opened 2016-03-22
Async versions of QueryFirst/Single/FirstorDefault/SingleOrDefault with CommandDefinition are missing


Tornhoof:
Hi,
currently there is:

``` C#
static T QueryFirstOrDefault<T>(this IDbConnection cnn, CommandDefinition command)
```

but there is no 

``` C#
static Task<T> QueryFirstOrDefaultAsync<T>(this IDbConnection cnn, CommandDefinition command)
```

only the full version exists:

``` C#
static Task<T> QueryFirstOrDefaultAsync<T>(this IDbConnection cnn, string sql, object param = null, IDbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
```

Same applies to First/Single/SingleOrDefault

Please add the CommandDefinition versions of async First etc. methods too.

mgravell:
Checked and agreed, yeah that seems to be missing

NickCraver:
Adding to the v2 list.

tskimmett:
I don't see how this is a breaking change.

mgravell:
@tskimmett I don't think anyone suggested that it was (although we haven't done full depth check) - merely setting priorities and expectations

diegofrata:
@mgravell @tskimmett It is a breaking change as it adds a parameter to an existing method. Any binaries compiled against a previous version of Dapper will be unable to find the method and it will break in runtime.

tskimmett:
It doesn’t require adding a parameter. It is simply a method overload (new signature). No recompilation required unless you want to use the new signature.

Tornhoof:
Those methods are extra methods, not extra optional parameters, so it's not a breaking change.
Anyway the related PR was merged a year ago.

diegofrata:
Ha! I actually mistook the **entire** discussion for the addition of CancellationToken parameter to the Async methods as I navigated from the v2.0 planning issue into this one. My bad, should have read it more carefully.

Perhaps someone wants to close this?

diegofrata:
That's how I got here: https://github.com/StackExchange/Dapper/issues/688

say25:
This issue seems to be closed by #600

NickCraver:
Yep, closing out!

PR #600, opened 2016-09-07
Add CommandDefinitions versions for async Query API

Tornhoof:
Trivial ports of the non-typesafe versions to the typesafe versions of the async query api.
Solves issue #484


mgravell:
Looks good to me; cheers

Issue #563, opened 2016-07-15
QueryAsync no longer throws SQL Exception from 1.50.*


OliverRC:
We've run into an issue with the latest 1.50.\* release when raising errors in SQL.
When calling `QueryAsync` on a stored procedure that selects a result set and then does a `RaiseError` no `SQLException` is thrown anymore. This used to be the behavior in 1.42.0.

It's almost as if the error is being "swallowed" somewhere.

Example code:

```
private static async Task<IList<Example>> TestAsync()
{
  const string connectionString = "User ID=sa;Password=password;Data Source=TestServer;Initial Catalog=DapperTest;Persist Security Info=False";

  using (var connection = new SqlConnection(connectionString))
  {
    connection.Open();
    // In 1.42.0 this throws an error. From 1.50.* it doesn't.
    return (await connection.QueryAsync<Row>("dbo.TestProc", commandType: CommandType.StoredProcedure)).ToList();
  }
}
```

Object

```
public class Example
{
  public override string ToString()
  {
    return SomeValue.ToString();
  }

  private int SomeValue { get; set; }
}
```

Stored Procedure

```
CREATE PROCEDURE dbo.TestProc
AS BEGIN

  DECLARE @SomeValue1 INT = 123;
  DECLARE @SomeValue2 INT = 456;
  DECLARE @SomeValue3 INT = 789;

  -- If RAISEERROR is done before the SELECT, an exception will be received in C#, using Dapper 1.50.1.
  --RAISERROR('before select', 16, 1);

  SELECT @SomeValue1 SomeValue
    UNION ALL
    SELECT @SomeValue2 SomeValue
    UNION ALL
    SELECT @SomeValue3 SomeValue;

  -- If RAISEERROR is done after the SELECT, no exception will be received in C#, using Dapper 1.50.1
  RAISERROR('after select', 16, 1);

END;
```

OliverRC:
Is there any update on this? It is currently an issue in our production codebase which we'd like to get resolved.

JeremyTCD:
This issue is related to #525 . It's a pretty big issue for me too, using a dirty fix in some `SqlMapper` functions for now:
                   
    else if ((row & Row.FirstOrDefault) == 0) 
    {  
        // Extra read retrieves SqlException
        await reader.ReadAsync(cancel);
        ThrowZeroRows(row);
    }

OliverRC:
Any update on this?

OliverRC:
[...]

PR #602, opened 2016-09-08
Make CommandBehavior optimizations more configurable

mgravell:
Default to SingleRow=allowed, SingleResult=disallowed - but expose options; this fixes issues like #563 automatically, and allows the performance aspect  of #554 to be fixed via a global setting

### DIFF
--- a/Dapper/SqlMapper.Settings.cs
+++ b/Dapper/SqlMapper.Settings.cs
@@ -21,6 +21,7 @@ namespace Dapper
             /// <summary>
             /// Gets or sets whether dapper should use the CommandBehavior.SingleResult optimization
             /// </summary>
+            /// <remarks>Note that a consequence of enabling this option is that errors that happen <b>after</b> the first select may not be reported</remarks>
             public static bool UseSingleResultOptimization
             {
                 get { return (AllowedCommandBehaviors & CommandBehavior.SingleResult) != 0; }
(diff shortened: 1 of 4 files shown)


Dapper/SqlMapper.Settings.cs
```diff
@@ -7,6 +10,47 @@ partial class SqlMapper
...
+            {
+                if (AllowedCommandBehaviors == DefaultAllowedCommandBehaviors
+                    && (behavior & (CommandBehavior.SingleResult | CommandBehavior.SingleRow)) != 0)
+                {
+                    if (ex.Message.Contains(nameof(CommandBehavior.SingleResult))
+                        || ex.Message.Contains(nameof(CommandBehavior.SingleRow)))
+                    { // some providers just just allow these, so: try again without them and stop issuing them
+                        SetAllowedCommandBehaviors(CommandBehavior.SingleResult | CommandBehavior.SingleRow, false);
+                        return true;
+                    }
```
NickCraver: IMO, these should be independent disables - would split this if/disable into one per `CommandBehavior`. For example if we added a third `CommandBehavior` that auto-disables this would get hairy...baby/bathwater situation possibly. I'm trying to think of a better detection method than this but I don't have better suggestions, +1 on going with the `.Contains()` you have.

Dapper/SqlMapper.Settings.cs
```diff
@@ -7,6 +10,47 @@ partial class SqlMapper
...
+            // disable single result by default; prevents errors AFTER the select being detected properly
+            const CommandBehavior DefaultAllowedCommandBehaviors = ~CommandBehavior.SingleResult;
+            internal static CommandBehavior AllowedCommandBehaviors { get; private set; } = DefaultAllowedCommandBehaviors;
+            private static void SetAllowedCommandBehaviors(CommandBehavior behavior, bool enabled)
+            {
+                if (enabled) AllowedCommandBehaviors |= behavior;
+                else AllowedCommandBehaviors &= ~behavior;
+            }
+            /// <summary>
+            /// Gets or sets whether dapper should use the CommandBehavior.SingleResult optimization
```
NickCraver: total nit: capitalize Dapper (same repeated below)

Issue #515, opened 2016-05-09
Dapper.Contrib Update ignores mapping


xalikoutis:
Hi,

some fields in my table have underscores so i did
`DefaultTypeMap.MatchNamesWithUnderscores = true;`
which works just fine when i retrieve entity with Get<> but throws "Invalid column name 'FullName'." 
when i update

my model 
`public class SmUser
    {
        public int Id { get; set; }
        
        public string Name { get; set; }
        
        public string FullName { get; set; }
       
        public string Email { get; set; }
        
    }`

Whet i run Get it works just fine
`var user = cn.Get<SmUser>(id);`

but when i update the entity
`
user.Email = user.Email + "z";
cn.Update(user);
`
i get 
`An exception of type 'System.Data.SqlClient.SqlException' occurred in System.Data.dll but was not handled in user code`
`"Invalid column name 'FullName'."`

and the sql query
`update smuser set [Name] = @Name, [FullName] = @FullName, [Email] = @Email where [Id] = @Id`

FullName is the entity field full_name is the db field

xalikoutis:
This feature is not supported . In Dapper.Contrib.SqlMapperExtensions.cs you fetch only the properties from the entity type and construct directly the update statement.

DefaultTypeMap.MatchNamesWithUnderscores or any implementation of SqlMapper.ITypeMap is not taken in account

stevo-knievo:
Same for `Insert<>` 

Dapper: 1.50.0-rc2 and Dapper.Contrib: 1.50.0-beta8

pdcmoreira:
I've created a pull request that implements the mapping for Insert and Update the last week, but I don't know if the maintainer have seen it yet.

stevo-knievo:
Nice work! I'm happy to test it as soon as gets integrated.

secretwep:
Wow, I was really counting on this working.  The db schema I'm using still has some old tables in it with underscored columns and the classes all utilize CamelCase property names.  I went to test some updates and that didn't go so well.

tariq-n-ahmad:
Is this *really* not addressed 4 years later?

koshovyi:
`MatchNamesWithUnderscores = true` still now working in insert case :(

agarzas:
+1

blai30:
+1

BunnyMan1:
My take on this to update a Class with lots of properties:
```
var query = $@"UPDATE {tableName} SET ";
[...]

PR #605, opened 2016-09-09
MatchNamesWithUnderscores for Insert and Update

pdcmoreira:
Insert and Update now take in consideration the DefaultTypeMap.MatchNamesWithUnderscores option and follow the model properties' PascalCase to table columns' snake_case convention. #515

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -17,6 +17,15 @@ using System.Threading;
 
 namespace Dapper.Contrib.Extensions
 {
+    public static class Utility
+    {
+        public static string PascalCaseToSnakeCase(string pascalCaseString)
+        {
+            return string.Concat(pascalCaseString.Select((character, index)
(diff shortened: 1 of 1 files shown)


rhubley:
looks like there are several AppendColumnNameEqualsValue methods, you only fixed one to be columnName = @propertyName the rest are still columnName = @columnName

pdcmoreira:
@rhubley I've changed several occurrences:
https://github.com/StackExchange/dapper-dot-net/pull/605/files

Can you please check? I might have missed something, but it's currently working for Insert and Update.

NickCraver:
The general problem with supporting this is that functionality was *one way* (reading). To respect it writing it a breaking change and I'm not sure it's a good one. I'm going to close this out in favor of something else: I think we need a `[Column]` attribute to solve this in a very general way for everyone. There are quite a few issues in it and it makes a more direct and explicit solution available for all.

A link to the v2 planning should appear below this shortly for discussions which are very welcome. We want your input. If you don't agree with this general solution, ping me and I'll reopen this...just trying to consolidate a lot of activity for v2.

PR #615, opened 2016-09-16
Update readme.md

hdwatts:
Missed second bracket in ID example


mgravell:
Fair observation. Thanks for the fix.

PR #618, opened 2016-09-21
GridReader.IsConsumed wouldn't be true

azyobuzin:
I found that GridReader.IsConsumed wouldn't be true when I use Read method with splitOn parameter. Here is a screenshot of this issue.

So I fixed this. Here is the expected behavior.


mgravell:
Looks great, thanks

Issue #385, opened 2015-11-06
Let's make the type mapper more pluggable.


NickCraver:
I'm seeing a big picture of a lot of issues solved in a single way here. People want to *slightly* adjust how the type mapping works in various ways - and I think we can do a better job of allowing this. 

Here are some examples:
 - #376 Changing the mapping using a `[Column]` attribute (I'm thinking `[DataMember]`)
 - #360 Customizing the `Query<T>` mapper
 - #350 More automated multi-mapping
 - #326 An Azure-side issue with `Id` columns that could possibly be worked around
 - #319 Changing how `TimeSpan` maps
 - #254 Option to throw with unmapped columns (use cases for both directions)
 - #251 Colmaps for Oracle
 - #198 Supporting Nodatime

This one is a more questionable example (as in questionable if this would help):
 - #262 Query abstract types

There may be a few I missed, but you get the idea. You can replace the `DefaultTypeMap` since #110 was merged a few weeks ago. However, we can do better here by allowing DefaultTypeMap to be plugged a bit rather than clipboard inherited all over. We'd also need to look at what functionality could move down into it to be pluggable (such as column names).

I'm not arguing for this to all be in core - I honestly think it'd be a bad idea if all of it was. I think making it pluggable makes some of the odder cases, if common, appear as mappers you could easily plug in from contrib. For example, what if #376 could be fixed by:
```C#
SqlMapper.TypeMapProvider = t => Dapper.Contrib.DataMemberMapper;
```
and #350 by:
```C#
SqlMapper.TypeMapProvider = t => MyCustomMultiMapper;
```

Posting this here for thoughts.

henkmollema:
I created a utility to map POCO properties to database column names using a fluent API. It uses the `CustomPropertyTypeMap`: https://github.com/henkmollema/Dapper-FluentMap (and specifically [`FluentTypeMap`](https://github.com/henkmollema/Dapper-FluentMap/blob/master/src/Dapper.FluentMap/TypeMaps/FluentTypeMap.cs))

To accomplish this I have to add each type map using `SqlMapper.SetTypeMap()`, which is a bit clumsy.

megakid:
Agree with this.  I wanted to mark all `DateTime` coming via Dapper as `DateTimeKind.Utc` as the MySql.Data marks them as `DateTimeKind.Unspecified`.  This proved virtually impossible and had to do it as a post step by reflecting over all properties and marking `DateTime` properties as UTC.

[...]

PR #623, opened 2016-09-28
Dapper Contrib - Column mapping

mjashanks:
This allows us to apply Property -> Column mapping in a pluggable way.

I originally made this pull request, which was too opinionated: https://github.com/StackExchange/dapper-dot-net/pull/376

That previous PR was closed, and batched into: https://github.com/StackExchange/dapper-dot-net/issues/385

I believe that issue 385 is not the correct place for this, as it is talking specifically about making the mapper more pluggable (i.e. Mapping the data from SqlDataReader to a POCO) - which would only cover SELECTs.

 My PR, is more about the generated SQL that is executed, thus covers INSERT, UPDATE, DELETE & SELECT - i.e. the responsibility of Dapper.Contrib

I have implemented in a similar way to `TableNameMapper` .. where the user must override the following delegates:

`GetAllColumns` - returns all mappings for a type
`GetPersistentColumns` - returns writeable (i.e. Insert & update) mappings for a type
`GetKeyColumns` - returns key mappings for a type

This would be a breaking change for anybody who has implemented their own ISqlAdapter, but that is all

I have only tested this for SQL Server

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ Test.DB.*
 TestResults/
 Dapper.Tests/*.sdf
 .dotnet/*
+.idea.*/
+Dapper.userprefs
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -33,7 +33,7 @@ namespace Dapper.Contrib.Extensions
                 var key = GetSingleKey<T>(nameof(GetAsync));
                 var name = GetTableName(type);
 
-                sql = $"SELECT * FROM {name} WHERE {key.Name} = @id";
+                sql = $"SELECT * FROM {name} WHERE {key.ColumnName} = @id";
                 GetQueries[type.TypeHandle] = sql;
             }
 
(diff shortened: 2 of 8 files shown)


NickCraver:
At a cursory glance, this also appears to be a breaking change for anyone with *non*-aliased columns, inherited classes, etc. The `Select *` => column list change is quite a huge breaking change. What's the goal here? I don't want to dismiss the ideas off-hand, but this implementation (as-is) would break far too many use cases.

mjashanks:
The goal overall is to be able to override the default Column Name To Property Name mapping. So, I believe it would be necessary to generate a column list for `Get`,  in order to do this.

However, I don't understand how this breaks anything. The default behaviour here is to generate a column names, equal to the POCO's property names. Is this not already the assumption with `select *` ? I.e.  `select *` will produce a datareader that has items named by column names. These items then get mapped directly to Property names. Am I misunderstanding anything ?

NickCraver:
@mjashanks There's a matching problem in either direction here. Let me see if I can clear it up a bit:

We have a choice at the first level: do we enumerate *all* of the columns for the `SELECT` list, or just the mapped ones with properties? If only the mapped, that's a breaking change. If we don't do only mapped and do all properties, then we hit another issues: properties that aren't in the database at all, generating an invalid `SELECT` fo columns that don't exist.

In general, Dapper is a "get everything, map what's there", because that's the only safe scenario to do such. Contrib *could* behave differently in generation forcing mapping of all columns with attributes that that's hard for 2 reasons: a) it's a breaking change, and b) you (currently) have no way to control classes that aren't yours, e.g. from a shared library or something else.

mjashanks:
@NickCraver - appreciate you taking the time to explain. 

I have modified this so that it uses ``select *``- unless the behaviour is overridden. Thus this aspect should no longer be a breaking change.

frankhommers:
I have made a PR with the same functionality, only a different implementation. I believe it's a bit more compatible. https://github.com/StackExchange/dapper-dot-net/pull/653

NickCraver:
@mjashanks I'm consolidating all of the discussion here into #722, would you please weigh in and help us design this feature to work for as many cases as we reasonably can?

Issue #280, opened 2015-04-30
Sqlbuilder template regex


madagaga:
The current regular expression in Sqlbuilder template causes issue when clause is at the begining.
For example `"SELECT /** whatever **/ F1, F2 .. From T1  /** where **/"`
the result is `"SELECT "` 

Replacing with this regex will fix the issue: `\/ \*\*[^\*]*\*\*\/`

jods4:
After replacing placeholders, Dapper removes unused templates with the regex on L67, which is `/\*\*.+\*\*\/`.
The issue here is that Regex is greedy by default, so `.+` will match as much as possible (i.e. the whole string) and backtrack from there until it matches the end of the pattern (`**/`). This is inefficient (backtracking a long string) and matches far too much if your template looks like `/** 1 **/ keep this /** 2 **/`.

The proposed solution is ok-ish in this case but not perfect. By not matching `*`, the following block will not match: `/** x*y **/`.

I suggest using a non-greedy modifier `.+?`, the full regex becomes `/\*\*.+?\*\*\/`.
A non-greedy regex will match as little as possible, which is exactly what we want. 
It will perform better and match the comments correctly, even when there's a star inside.

Another option comes to mind: obviously the syntax `/** xyz **/` was chosen because it is a SQL comment, hence valid SQL. So why remove the unused templates _at all_? I guess it would be even more efficient to do nothing, no?

jods4:
PR #625 fixes this bug, assuming you want to keep removing the unused comments/templates.
@NickCraver would be nice if you could merge and publish an updated package. This has been open for 1.5 years and can generate at best incorrect SQL, at worse security issues.

mgravell:
Merged PR

PR #625, opened 2016-10-05
Non-greedy match for comments regex

### LINKED ISSUE #436: SqlBuilder regex expression removes chars outside /** **/ clauses
The issue is on SqlBuilder line 67:

    private static readonly Regex _regex = new Regex(@"\/\*\*.+\*\*\/", RegexOptions.Compiled | RegexOptions.Multiline);

Given the following sql input:

     select table.* FROM SomeTable as table /**innerjoin**/  /**fulljoin**/  /**leftjoin**/  /**rightjoin**/ INNER          JOIN some join data /**where**/  /**orderby**/ OFFSET @skip ROWS FETCH NEXT @take ROWS    ONLY

it deletes "INNER JOIN some join data" if where/order by clause are not present.
It can be tested online on:
http://regexstorm.net/tester
with above sqlinput and `\/\*\*.+\*\*\/` expression.

If I change the expression to `/\*\*(.+?)\*\*/` seems to work fine.

jods4:
Fixes #436, #280
My editor has also normalized a few line endings to LF like 99% of the
file was.

### DIFF
--- a/Dapper.SqlBuilder/SqlBuilder.cs
+++ b/Dapper.SqlBuilder/SqlBuilder.cs
@@ -64,7 +64,7 @@ namespace Dapper
                 _builder = builder;
             }
 
-            private static readonly Regex _regex = new Regex(@"\/\*\*.+\*\*\/", RegexOptions.Compiled | RegexOptions.Multiline);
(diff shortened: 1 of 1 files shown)


mgravell:
Thanks; looks good

PR #638, opened 2016-11-05
Add SqlMapper.RemoveTypeMap() method

hakon1912:
This is intended to solve the same issue as #471 and #458, without introducing a breaking change.


NickCraver:
Thanks for the contrib!

PR #640, opened 2016-11-08
Changed Insert methods and interface to return id as long.

pjsvis:
The contrib extension insert methods were throwing when ids of greater
than 2147483647 were returned. This occurred because they id was being
cast to an int. Changing to long fixes the problem.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -330,7 +330,7 @@ namespace Dapper.Contrib.Extensions
                     sbParameterList.Append(", ");
             }
 
-            int returnVal;
+            long returnVal;
             var wasClosed = connection.State == ConnectionState.Closed;
             if (wasClosed) connection.Open();
 
@@ -688,7 +688,7 @@ namespace Dapper.Contrib.Extensions
 
 public partial interface ISqlAdapter
 {
-    int Insert(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, string tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert);
+    long Insert(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, string tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert);
     
     //new methods for issue #336
     void AppendColumnName(StringBuilder sb, string columnName);
@@ -697,7 +697,7 @@ public partial interface ISqlAdapter
 
 public partial class SqlServerAdapter : ISqlAdapter
 {
-    public int Insert(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, string tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert)
+    public long Insert(IDbConnection connection, IDbTransaction transaction, int? commandTimeout, string tableName, string columnList, string parameterList, IEnumerable<PropertyInfo> keyProperties, object entityToInsert)
     {
         var cmd = $"insert into {tableName} ({columnList}) values ({parameterList});select SCOPE_IDENTITY() id";
         var multi = connection.QueryMultiple(cmd, entityToInsert, transaction, commandTimeout);
@@ -705,7 +705,7 @@ public partial class SqlServerAdapter : ISqlAdapter
         var first = multi.Read().FirstOrDefault();
         if (first == null || first.id == null) return 0;
 
-        var id = (int)first.id;
+        var id = (long)first.id;
         var propertyInfos = keyProperties as PropertyInfo[] ?? keyProperties.ToArray();
         if (!propertyInfos.Any()) return id;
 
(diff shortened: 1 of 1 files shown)


NickCraver:
The problem is this is a huge breaking change. We can't just change it. Instead, we need to create generic overloads which has just kept getting punted down the priority list :-/

pjsvis:
No worries Nick. The fix works for our codebase so we will incorporate the source and retire the contrib dependency

philn5d:
Would the generic solution involve insert<T> plus the int overload calling insert<int> or would it be better to pass in a conversion func? If the former, then how to ensure proper casting for every case? Plenty of open issues for similar casting during mapping on queries. Perhaps the same typemapping strategy should be used to map return values based on T.

philn5d:
Oops, forgot to escape insert&lt;T&gt;

NickCraver:
@phillip-haydon The version I'm picturing isn't very complicated, it just need doing. And honestly Contrib just needs a lot of love in general that we haven't had time for.

Here's what exists today:

```c#
public static long Insert<T>(this IDbConnection connection, T entityToInsert, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
```

What I'm picturing (roughly, the point is generics) is:
```c#
public static TKey Insert<T, TKey>(this IDbConnection connection, T entityToInsert, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
```
And the existing `int` implementation remains (no breaking changes), but it's really just calling this with `int` as the `TKey` underneath (or `long` which *should* be safe, I think), e.g.:
```C#
public static long Insert<T>(this IDbConnection connection, T entityToInsert, IDbTransaction transaction = null, int? commandTimeout = null) where T : class =>
    Insert<T, long>(connection, entityToInsert, transaction, commandTimeout)
```
What this doesn't solve is composite keys which are across multiple fields. But at a larger scope, that's a bigger problem with Dapper.Contrib in general. Any thoughts on such an approach?

Honestly, we should make a proper API discussion here as an issue with proposed things on all providers. We want `Async` methods, cancellation tokens, etc. here and it's better to look at it holistically. Realistically I can't give that a lot of love in the near future but would *really* welcome some help.

philn5d:
I'm a bit of a noob as far as contributing, but I'd like to help with this.  For sure about the composite keys, was thinking about that too. Also, not sure if
```
return val as TKey;
```
Is always going to work. As I'm gathering that some types need special mapping rules, especially for supporting composite keys.

I will take a stab at a first version though, without the composite case, as the 80% case is int, long or Guid.

I cloned the repo recently, but I believe I haven't got it running local yet...perhaps some additional steps to take first.

phillip-haydon:
Haha @NickCraver, you pulled in the wrong Phillip :)

philn5d:
Ok, got a bit of a start here. One issue is really with the SQL specific adapters and that interface. Those each return int as well.

philn5d:
This is a bit of a trip down the rabbit hole. In order to test this, I would need to install ALL supported databases OR rewrite most of Contrib and core to make it testable. If I could fake IDbConnection it would be doable, but QueryMultiple and other extension methods are used. I've never faked an extension method before - going to see if its possible. 

In the long view, I may be taking things in a direction that aren't suitable for the goals of Dapper. Dapper seems to use a lot of static and there may be a specific reason for that. They are hard to double in unit test contexts, which is presenting obstacles to ramping up for making changes to Contrib. I would like to contribute, but I don't want to spin gears too much in the wrong direction.

Some ideas I've had so far along the way - there are a few main components in contrib: SqlGeneration, mapping (properties to SQL, results to properties), caching, and glue to Dapper core. At the moment, these are sort of mashed together. So far, most of the issues with Contrib are related to results mapping. Currently, only the int id PK is supported. It should support different types and combo-keys. Maybe give the option to return the whole row to an object as well, else how to handle multiple values for the multi-part key? Possibly separate type just for that? Dynamic? Think about if the return type is TKey, that's essentially saying TKey could be Person object unless its limited to decimal, long, Guid, etc.

philn5d:
Well, did some experimentation. Returning dynamic from the Insert would work IFF the SQL itself actually returned the Guid, but it doesn't in MySql anyways. Needs more work. One drawback to returning dynamic is that it has to be cast by the consumer, 
    ```var id = conn.Insert(...);```
has issues when you then do a
    ```var entity = conn.Get<User>(id);```
Have to specify type instead of var for the id or cast in some other way. Maybe (T) where T : struct or make an overload for each supported type. I also tried the approach of a return type mapping hash table, that works and would be more flexible if the consumer could supply alternate type mapping (would be able to support combo-keys. Other issue with combo key is that the last_inserted_id might not return the whole key More on this in progress.

rajivverma:
Util we have a generic solution, can we have a official build with long return type?
I am sure many of us would be facing this issue.

NickCraver:
Such a build would be a breaking change. We'll try and get this into the next major version. I'm working on MiniProfiler and Marc's working on StackExchange.Redis now, then we'll circle back to Dapper for a v2 pass. It's just too much to coordinate the 3 at once.

NickCraver:
I'm closing this out as we won't be going this way, but have referenced the discussion for fixing this in the v2 releases, link will appear below this in a few minutes.

PR #641, opened 2016-11-14
make CreateParamInfoGenerator and GetTypeDeserializerImpl can be cust…

fredliex:
hi, I need to custom emit for special rule,
so modify some code to allow SqlMapper's CreateParamInfoGenerator and GetTypeDeserializerImpl can be customized.

hope can commit this request, thanks a lot,

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -180,7 +180,7 @@ namespace Dapper
                     {
                         if (!paramReaderCache.TryGetValue(newIdent, out appender))
                         {
-                            appender = SqlMapper.CreateParamInfoGenerator(newIdent, true, RemoveUnused, literals);
+                            appender = SqlMapper.CreateParamInfoGeneratorImp(newIdent, true, RemoveUnused, literals);
                             paramReaderCache[newIdent] = appender;
                         }
                     }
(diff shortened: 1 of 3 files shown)


NickCraver:
@mgravell Thoughts here? I'm *very* hesitant on making any API like this public. I think the very narrow use case vs. the brittleness it introduces likely isn't worth the tradeoff.

NickCraver:
@fredliex Can you please explain your use case here?

fredliex:
@NickCraver 
in my case, need more control for object to param and DataReader to typed model.
I use attribute to define behavior, ex:
1. auto TrimEnd char to string property.
2. null member property <--> not null columen
3. Enum <--> special string
4. property name a  <---> column name b
5. mapping nonpublic property/field
6. etc...

example in 
https://github.com/fredliex/Framework.Dapper/blob/master/Framwork.Dapper/Framework.Dapper.Test/DataTest.cs

sorry for the comment is chinese language.

NickCraver:
@fredliex I think the blunt answer here is: I don't think this belongs in the core lib. It'd be very brittle and greatly restrict us in refactoring the inner bits. I think if you're going *that* far custom, a fork is in order.

PR #643, opened 2016-11-18
Update Readme.md

AArayeshi:
Hi, I have developed a separate ORM performance Benchmark project that now exist in the GitHub. So in this change I have just added my project link to your Readme.md file as a performance testing tools. Thanks.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -229,7 +229,7 @@ Performance benchmarks are available [here](https://github.com/StackExchange/dap
 
 Feel free to submit patches that include other ORMs - when running benchmarks, be sure to compile in Release and not attach a debugger (ctrl F5).
 
-Alternatively, you might prefer Frans Bouma's [RawDataAccessBencher](https://github.com/FransBouma/RawDataAccessBencher) test suite.
+Alternatively, you might prefer Frans Bouma's [RawDataAccessBencher](https://github.com/FransBouma/RawDataAccessBencher) test suite or [OrmBenchmark](https://github.com/InfoTechBridge/OrmBenchmark).
 
 Parameterized queries
 ---------------------


mgravell:
Sure, I don't  have a  problem with that. Can I ask what the current metrics show?

philn5d:
Interesting Amir, what is the impact of different query text between providers? "SELECT col1, col2, ... ,coln FROM Posts ..." (ADO lib) is a bit different from "SELECT * FROM Posts" (Dapper lib). Is there an impact on the underlying provider performance which may skew results?

Would be good to see "cold" v "warm" results. (i.e., no query plan cache v query plan cache). "DBCC FREESESSIONCACHE
DBCC FREEPROCCACHE
DBCC DROPCLEANBUFFERS" to do a "cold" run v warm run comparison OR warm up the caches with an initial run before benchmarking.

Ever consider making a fake provider to minimize the impact of the underlying provider? Maybe return DataSet which is what the providers are returning anyways. At that point you'd really only be benchmarking the mapping and taking the provider out of the picture.

jamesholwell:
+1 what @philn5d said

mgravell:
(those  are great points, but should probably be raised as issues on https://github.com/InfoTechBridge/OrmBenchmark

AArayeshi:
Thanks Marc, I have created my project based on your performance test project in dapper. I have tried to categorize results and make easy plug-able project. The result of test is a little different based running environment but there is good competition between Dapper and PetaPoco.
The performance tests are broken in the following lists:
A-Performance test of executing one SQL command frequently (500 times):
A-1-Performance of execute one select statement and map a returned row to a POCO object over 500 iterations
A-2-Performance of execute one select statement and map a returned row to a dynamic object over 500 iterations
B-Performance test of mapping database records to POCO/Dynamic objects (5000 row):
B-1-Performance of mapping 5000 rows returned by one select to POCO objects in one iteration
B-2-Performance of mapping 5000 rows returned by one select to Dynamic objects in one iteration

AArayeshi:
Thanks @philn5d for your good points. 
About ADO select, your rights, I will fix it. 
About cold run or warm run, I am not sure but I think if there was some configuration for ORMs to manage caching or turn on/off the cache, we can have this type of bench marking. Also some provider same as Dapper has multiple test such as (Buffered/UnBuffered) that they will boost each other based on witch one run firstly (specially for caching), so second test will have a better result. By the way this is initial project and there is much to do.
Thanks.

PR #651, opened 2016-12-01
added firebird support

pdesterlich:
added firebird sql adapter to allow record insert and key retrieving

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -49,6 +49,7 @@ namespace Dapper.Contrib.Extensions
                 {"npgsqlconnection", new PostgresAdapter()},
                 {"sqliteconnection", new SQLiteAdapter()},
                 {"mysqlconnection", new MySqlAdapter()},
+                {"fbconnection", new FbAdapter() }
             };
 
         private static List<PropertyInfo> ComputedPropertiesCache(Type type)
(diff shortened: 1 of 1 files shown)


masterbb:
Works fine, tested it a bit. You just seem to have forgotten the Async class.
I do somewhat require this, but I don't know how to push something to a pull request. I can create my own pull request if you don't mind.

pdesterlich:
You're more than welcome to create your pull request providing async support too.
I'm not using async (yet), so i only changed the part i use and i created a pull request just in case someone had the same problem.

mgravell:
that looks pretty awesome; one bit of feedback: when I wrote some very
similar code in the core of dapper, some folks (correctly) noted that the
use of interpolated strings *could* hypothetically be abused on a
compromised server to do very very bad things. Of course  by that note: if
you're already compromised you're kinda screwed anyways! I need to do a
review to see how much we are/aren't using interpolated strings generally
to assess whether this is meaningful feedback (it would be silly to
criticize this if we used it extensively every where else in the code)

NickCraver:
@mgravell It's used in the other adapters, so I'm for taking this as-is. I agree we should clean them all up though. I'll make a note for the v2 push to do so.

say25:
It seems that FB Async was added in https://github.com/StackExchange/Dapper/commit/a7ff0b3d9f33bbb19e628258b4ccb820f75f3fb9

PR #652, opened 2016-12-01
Implemented ColumnNameMapping

frankhommers:
I would love to see this pulled!

```cs
            SqlMapperExtensions.ColumnNameMapper = (propertyInfo) =>
            {
                Regex idAtEndRegex = new Regex("[a-z]Id$", RegexOptions.CultureInvariant);
                string name = idAtEndRegex.Replace(propertyInfo.Name, x => { return x.Value[0] + "_" + x.Value.Substring(1); }).ToLower();
                return name;
            };
```

This will allow you to write a custom Property.Name to column mapping.

The default behaviour does not change.


frankhommers:
Sorry tabs were wrong.

PR #679, opened 2017-01-19
Update README with documentation for [ExplicitKey] attribute

markashleybell:
Added a bit of documentation relating to the changes made in #352, as I was initially confused about the difference between the `[Key]` and `[ExplicitKey]` attributes.


NickCraver:
Thanks for the updates!

PR #681, opened 2017-01-23
Added: Atrribute Column

scampos-s3d:
Added the attribute column. It allows to define the column name.

### DIFF
--- a/Dapper.Contrib/Readme.md
+++ b/Dapper.Contrib/Readme.md
@@ -119,6 +119,16 @@ Dapper.Contrib makes use of some optional attributes:
         public string Name { get; set; }
     }
     ```
+* `[Column("column_name")]` - use another column name instead of the name of the property
+
+	```csharp
+    public class Employee
+    {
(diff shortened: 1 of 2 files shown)


NickCraver:
The reasons we've never included such a thing is: it doesn't cover most cases. For example if you called any get methods, a column wouldn't map *back* properly. Including on `Select *` style approaches.

There's also a question of "does such a feature even need a new attribute?" or would `[DataMemeber]` suffice? But for anyone with currently decorated classes (e.g. for JSON serialization), this would be a breaking change and quite unexpected behavior.

In short: this would have to manifest in far more ways, and likely be in Dapper itself, not contrib. I'm not sure it meets the bar on demand, since it's nothing like the built-in behavior. It's worth taking a look at holistically in Dapper v2 though, from a view point of which bits of Contrib should migrate to Dapper core.

NickCraver:
Closing this out to consolidate - please check out and add your thoughts to the feature design in #722!

PR #686, opened 2017-01-26
Add github issue and pul request templates

ktaranov:
More details [here](https://github.com/blog/2111-issue-and-pull-request-templates)


NickCraver:
While I do appreciate the pull request, I'm just not a fan of templates like this. I don't think they're necessary and with respect to the Dapper issues, they don't even cover a lot of the issues we hit. 

I just don't believe this is a net-win to include in the repo, because it has the inevitable downside of making issues that don't "fit" feel unwelcomed and creates a higher barrier for all.

ktaranov:
Thanks for explanation Nick.

PR #689, opened 2017-02-01
Fix for InsertAsync in PostgresAdapter

GermanTikhonov23:
Dapper: 1.50.2
Dapper.Contrib: 1.50.0
Npgsql: 3.2.0

When I'm trying to InsertAsync entity I get exception **Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: Cannot apply indexing with [] to an expression of type 'object'**. Sync implementation works fine. Comparing Sync and Async versions of PostgresAdapter I found one difference.There is a cast to **IDictionary<string, object>** of **results.First()** before [] usage. Is seems that someone fixed issue in Sync but forgot to fix the same in Async implementation. I copied that cast from Sync to Async.


NickCraver:
All of these need the overhaul for v2, but the fix is good to get in ASAP regardless - could you please rebase so I can merge it in?

NickCraver:
Merging in - I'll clean all these up for functionality and performance in v2. Consistency with sync version is best for now, +1 on this change as-is for that reason.

PR #693, opened 2017-02-06
Add GetRowParser documentation/example

bretcope:
There are obviously several different ways you could use `GetRowParser`, but I just wanted to give a gist of how to use it so people, 1. know that the feature exists, and 2. don't have to search through code and tests for an example.


NickCraver:
Awesome - thanks!

PR #696, opened 2017-02-10
Remove DynamicParamaters.Output() dependency on templates

doc-def:
The current implementation of DynamicParameters.Output(), only updates the output target if the DynamicParameters instance is constructed with a template: i.e. public DynamicParameters(object template);

See: http://stackoverflow.com/questions/39904049/dynamicparameters-outputt-doesnt-map-the-output-to-the-model

As the target is included in the 'Output' definition, I see no reason for this limitation, and propose a simple fix to the DynamicParameters class that removes this dependency.

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -209,15 +209,15 @@ namespace Dapper
                         });
                     }
                 }
+            }
 
-                // Now that the parameters are added to the command, let's place our output callbacks
-                var tmp = outputCallbacks;
-                if (tmp != null)
+            // Now that any template parameters have been added to the command, let's place our output callbacks
+            var tmp = outputCallbacks;
+            if (tmp != null)
+            {
+                foreach (var generator in tmp)
                 {
-                    foreach (var generator in tmp)
-                    {
-                        generator();
-                    }
+                    generator();
                 }
             }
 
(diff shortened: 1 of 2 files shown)


Dapper/DynamicParameters.cs
```diff
@@ -463,11 +463,13 @@ public DynamicParameters Output<T>(T target, Expression<Func<T, object>> express
...
-                    parameter.ParameterDirection = parameter.AttachedParam.Direction = ParameterDirection.InputOutput;
+                    parameter.ParameterDirection = ParameterDirection.InputOutput;
+                    parameter.Size = sizeToSet;
 
-                    if (parameter.AttachedParam.Size == 0)
+                    if (parameter.AttachedParam != null)
                     {
-                        parameter.Size = parameter.AttachedParam.Size = sizeToSet;
+                        parameter.AttachedParam.Direction = ParameterDirection.InputOutput;
+                        parameter.AttachedParam.Size = parameter.AttachedParam.Size == 0 ? sizeToSet : parameter.AttachedParam.Size;
```
NickCraver: This can simply be: 
```c#
if (parameter.AttachedParam.Size == 0) parameter.AttachedParam.Size = sizeToSet;
```

Dapper/DynamicParameters.cs
```diff
@@ -463,11 +463,13 @@ public DynamicParameters Output<T>(T target, Expression<Func<T, object>> express
 
                 if (parameters.TryGetValue(dynamicParamName, out parameter))
                 {
-                    parameter.ParameterDirection = parameter.AttachedParam.Direction = ParameterDirection.InputOutput;
+                    parameter.ParameterDirection = ParameterDirection.InputOutput;
+                    parameter.Size = sizeToSet;
```
NickCraver: This is a behavior change, which may affect `max` column sizes. It's no longer conditional.

doc-def: I had to split both the direction and size assignments into two due to `parameter.AttachedParam` being null if a template object is not passed to the `DynamicParameters` constructor.

Happy to not change the `Size` property at all here, and only do it if `AttachedParam` is not null.

Obviously `Direction` will still need to change.

Dapper/DynamicParameters.cs
doc-def: Agree, only left it the way it was to keep coding style / minimise changes as not sure on the policy for these things.

I'll update the code when I'm more awake tomorrow.

doc-def:
I think it might be better to move the hooking up of the output callbacks, i.e.
```
                var tmp = outputCallbacks;
                if (tmp != null)
                {
                    foreach (var generator in tmp)
                    {
                        generator();
                    }
                }
```
Until after the non-template parameters have been created / processed, in other words after the `foreach (var param in parameters.Values)` loop in the same `AddParameters` method.

That way, all parameters should have an `AttachedParam` and I should be able to revert the code causing the concern.

doc-def:
(Sorry, new to github, did not mean to close this)

Issue #698, opened 2017-02-11
SqlMapperExtensions PostgresAdapter does not escape column names for primary key


colinbreame:
A table is created in Postgres like this:

```
CREATE TABLE Metric (
	"MetricId"		SERIAL,
	"Name"			VARCHAR(255),
	"CategoryId"	INT,
	"Type"			VARCHAR(255),
	"Text"			TEXT,

	PRIMARY KEY (MetricId)
)
```
When inserting the following SQL is created:

```
insert into Metric ("Name", "CategoryId", "Type", "Text") values ($1, $2, $3, $4) RETURNING MetricId
```

However in Postgres the columns names that are not escaped are lower cased.  So the column name MetricId does not equal "MetricId".  As my property name is "MetricId" this is causing an issue, as it is referred to inconsistently.

The suggestion solutions (as applied to PostgresAdapter) would be to escape the primary key column name.

NickCraver:
This really depends on the user, but we want to make it much easier to configure and work in many scenarios in v2. Collapsing this into #722 fo discussion.

PR #700, opened 2017-02-12
Dapper.Contrib unit tests for Postgres

colinbreame:
Contains changes to add unit tests for Postgres and fixes to make them pass.

This relates to the issue I created #698

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -33,8 +33,10 @@ namespace Dapper.Contrib.Extensions
                 var key = GetSingleKey<T>(nameof(GetAsync));
                 var name = GetTableName(type);
 
-                sql = $"SELECT * FROM {name} WHERE {key.Name} = @id";
-                GetQueries[type.TypeHandle] = sql;
+                var adapter = GetFormatter(connection);
+                var queryBuilder = new StringBuilder($"select * from {name} where ");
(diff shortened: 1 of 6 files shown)


NickCraver:
First, I appreciate the PR. Looking through the code, it's quite a change in current behavior and very breaking. Any case sensitivity changes cannot be in a 1.x release, they need to be a v2 item (see V2 planning here: https://github.com/StackExchange/Dapper/issues/688). The code isn't "fixed" for the unit tests, it's *changed* for the unit tests. In other cases, these cases/methods were already working *for some* (which makes this a breaking change).

We totally want unit tests here, and in general - but we need to handle column escaping in a much better way *globally*. The current situation is a mess across providers. I'll think through an initial spec and post (I hope) this week for discussion. We need to design an API for this and define behavior that's consistent across providers. For example, some ideas not well handled today:
- A `bool` to control whether we default to case sensitive or not
- A `[Column]` attribute to map individual column names specifically
- A general function (maybe `Func<Type, MethodInfo, string>`) to define behavior *unless* there's an explicit attribute
- Array behavior control
- An `Update` method?

There's a lot to think through here and it needs a coherent design. Barring that, we shouldn't bring in any more *breaking* changes, as they don't improve the situation. They help some people and unexpectedly (and in violation of SemVer) break others.

NickCraver:
Consolidating all of these PRs into #722, mostly for the reasons above - I'd really appreciate it if you could chime in and help design this feature set!

PR #701, opened 2017-02-12
Update readme to remove notice about requiring open connections

vondruska:
From https://twitter.com/Nick_Craver/status/815928282310971393, remove the note from the readme that mentions open connections are required for Dapper to function. Maybe this could be edited into something saying that if the connection is closed, Dapper will attempt to reopen it?


NickCraver:
Thanks!

Issue #703, opened 2017-02-13
Dapper.Contrib Key caching issue


rasmus-s:
There is a bug in the caching logic of the KeyProperties. Method UpdateAsync() and DeleteAsync() alter the original cached list of properties key [code ref](https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.Async.cs#L217).
So, if the first operation is a Update or Delete the "KeyPropertiesCache" return a pointer to the original list and the Update/Delete method add the explicitKeyProperties to the list. When the Insert() get called the "allPropertiesExceptKeyAndComputed" is wrong.

Test to reproduce it:

        public async Task TestMethod()
        {
            var myClass = new MyClass()
            {
                AccountId = Guid.NewGuid(),
                Name = "Some name"
            };

            using (IDbConnection db = new SqlConnection(connectionString))
            {
                //var x = await db.InsertAsync(myClass); // Add and test pass
                var t = await db.UpdateAsync(myClass);
                var r = await db.InsertAsync(myClass);                
            }
        }

Suggested fix that all returns from cache-layer (KeyPropertiesCache, ExplicitKeyPropertiesCache, ComputedPropertiesCache etc.) should call .ToList() on the item before return it.

cremor:
Related: #530

Rmkrs:
We also run into this issue. Since I don't want to branch off, and there is no merged fix, I ended up having to use reflection to call the SqlMapperExtensions.KeyPropertiesCache(type) to "warmup" the cache and work around the issue.

Ugh :(

abadyl:
Any idea when this fix will be merged?

popaandrei:
I am waiting also for this fix

paulisfluent:
Another vote for this fix from me.  (Is there a better way of indicating support for a fix than commenting here?)

jsign:
We're waiting too. 

The nature of the problem makes tons of existing code buggy. 
Every place where there's an UpdateAsync is waiting for adding a InsertAsync afterwards to raise this issue.

decipherer:
We're also running into this problem. Luckily we caught this in unit tests and we were able to work around it before running the code in production. But I'm amazed that this issue has been around for over a year without a fix. The issue seems clear and there even is a proposed solution. I've tested it on my local machine and it seems to fix the issue. Please fix this :)

mrtristan:
i just spent hours on this. I was really loving dapper to this point.

[...]

PR #704, opened 2017-02-13
Fix Dapper.Contrib alter caching keys issue #703

rasmus-s:
This PR contains fix for issue #703 - where cache keys get alter if you call Update/Delete and then Insert. Contains test to validate it. Due to type should not be in cache from other tests, I had created a new Class called ObjectW.


MarcWils:
Any idea when this PR will be approved & merged?

NickCraver:

PR #707, opened 2017-02-22
Added Belgrade SqlClient library in the performance benchmark

jovanpop-msft:
In the performance benchmark is added [Belgrade Sql Client data access library](https://github.com/JocaPC/CLR-Belgrade-SqlClient). This is a wrapper around ADO.NET classes that executes queries in async manner using callbacks. 
The following changes are added:
 - PerformanceTests.cs - Added code that gets one post by id using Belgrade SQL Client SqlMapper class.
 - project.json - Added reference to "Belgrade.Sql.Client": "0.6.2" NuGet package
 - project.json - Added new flag "BELGRADE" that controls should the Begrade test be executed.

### DIFF
--- a/Dapper.Tests/PerformanceTests.cs
+++ b/Dapper.Tests/PerformanceTests.cs
@@ -259,6 +259,34 @@ namespace Dapper.Tests
                 }, "Simple.Data");
 #endif
 
+#if BELGRADE
+
+                Try(() =>
+                {
+                    var query = new Belgrade.SqlClient.SqlDb.QueryMapper(TestSuite.GetOpenConnection());
+                    tests.Add(id => query.ExecuteReader("SELECT TOP 1 * FROM Posts WHERE Id = " + id,
+                        reader =>
+                        {
+                            var post = new Post();
+                            post.Id = reader.GetInt32(0);
+                            post.Text = reader.GetString(1);
+                            post.CreationDate = reader.GetDateTime(2);
+                            post.LastChangeDate = reader.GetDateTime(3);
+
+                            post.Counter1 = reader.IsDBNull(4) ? null : (int?)reader.GetInt32(4);
+                            post.Counter2 = reader.IsDBNull(5) ? null : (int?)reader.GetInt32(5);
+                            post.Counter3 = reader.IsDBNull(6) ? null : (int?)reader.GetInt32(6);
+                            post.Counter4 = reader.IsDBNull(7) ? null : (int?)reader.GetInt32(7);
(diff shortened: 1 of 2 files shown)


NickCraver:
Finally trying to get through some of these as time allows, thanks!

NickCraver:
Heads up: I noticed when porting Performance Tests to the new .csproj world that the results are not correct here. This is an async handler not being awaited, so we're not *actually* getting results like the other providers. I'll add proper async handling in the new setup and we can continue to tweak from there.

Ultimately, I think we want to go with BenchmarkDotNet here, but even then there's a fundamental issue: it's an apples to oranges situation where we're looking at wall clock time where as Belgrade focuses on concurrency. Both are awesome goals, but it's hard to directly compare the 2 approaches since they have tradeoffs for their goals in both directions.

Here's current output, note that the `async` overhead is *severely* penalizing Belgrade in an iteration test...so we should see how to best approach it, maybe async is another category/setup?

```
λ  .\Dapper.Tests.Performance.exe
Using ConnectionString: Data Source=.;Initial Catalog=tempdb;Integrated Security=True
Running 500 iterations that load up a post entity

Running...
Hand Coded took 104ms
DataTable via IDataReader.GetValues took 111ms
Mapper QueryFirstOrDefault took 112ms
Mapper Query (buffered) took 114ms
Dynamic Mapper QueryQueryFirstOrDefault took 114ms
Dynamic Mapper Query (buffered) took 115ms
PetaPoco (Fast) took 120ms
Susanoo Dynamic Mapping Static took 120ms
Dynamic Massive ORM Query took 123ms
Susanoo Mapping Static took 125ms
Dapper.Contrib took 130ms
Susanoo Dynamic Mapping Cache Retrieval took 143ms
PetaPoco (Normal) took 146ms
Linq 2 SQL Compiled took 149ms
Susanoo Mapping Cache Retrieval took 149ms
Mapper Query (non-buffered) took 163ms
Dynamic Mapper Query (non-buffered) took 167ms
Entity Framework SqlQuery took 192ms
NHibernate Session.Get took 206ms
NHibernate SQL took 217ms
NHibernate HQL took 223ms
NHibernate Criteria took 272ms
Linq 2 SQL ExecuteQuery took 307ms
Entity Framework No Tracking took 316ms
Entity Framework took 318ms
Simple.Data took 433ms
Belgrade Sql Client took 766ms
Linq 2 SQL took 825ms
NHibernate LINQ took 1289ms
```

As you can see, I'm working on making *all* providers we reasonably can work again. I'll get a branch up to GitHub soon to collaborate on.

PR #716, opened 2017-03-02
Ability to override default field comparer used in DapperTable (EX: e…

mkonstan:
Ability to override default field comparer used in DapperTable (EX: enable case insensitive comparison)


NickCraver:
Rather than adding a lot of one-offs here, I'm aiming to make this more pluggable in a general way - can you please review/comment on #722?

NickCraver:
Merging into #722 plan (and prepping for project format changes here)

PR #726, opened 2017-03-15
Removed dependency on dotnet sdk version 1.0.0-preview2-003121

jovanpop-msft:
This is proposed fix for https://github.com/StackExchange/Dapper/issues/725

On the machine with dotnet version 1.0.0-preview2-003156, I'm getting he following error when I open the solution file:

> The project is configured to use .NET Core SDK version 1.0.0-preview2-003121 which is not installed under the path C:\Program Files\dotnet. These components are required to build and run this project.

I have removed sdk version from global.js file and everything works fine. Is there some reason why is the solution bound to this version? I would propose to remove this.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,4 @@
 {
-  "sdk": {
-    "version": "1.0.0-preview2-003121"
-  },
   "projects": [
     "Dapper",
     "Dapper.EntityFramework",


NickCraver:
I think this depends on your local setup and at which point you downloaded on of the several iterations near each other here. Let me pull and test this on the laptop tonight...if it builds, yeah sure. I really want to get the entire thing on 2017 `.csproj`, but need to get through the PRs first.

NickCraver:
I tested this locally: 
screen shot 2017-03-22 at 20 48 03

...so yeah, it depends on what you have installed, unfortuantely. I'm going to close this out because it'll break others - but I'll work on switching over to the VS 2017 project format to eliminate the root problem: the tooling's all over the map.

Issue #649, opened 2016-11-29
Unable to pass an empty table-valued parameter through Dapper


burns47:
The following code will throw an exception on 1.50.2.0, with the message "to send a table-valued parameter with no rows, use a null reference for the value instead."

`Connection.Execute("TestTVP", new { Table = Enumerable.Empty<SqlDataRecord>() }, 
		transaction: Transaction, commandType: CommandType.StoredProcedure);
`

Following that advice, I used this code and got this message: "DBNull value for parameter 'Table' is not supported.  Table valued parameters cannot be DBNull"
`Connection.Execute("TestTVP", new { Table = null as IEnumerable<SqlDataRecord> }, 
		transaction: Transaction, commandType: CommandType.StoredProcedure);
`

I know this is probably SqlClient's fault, but shouldn't dapper handle this?

NickCraver:
We don't use TVPs much so I'm not sure exactly where this is thrown - we'll have to dig. Can you get some more info for when we're able to do this? Are you using the full SQL client, on full framework, against SQL Server? That'll save some time.

burns47:
No problem -- I'm targeting .NET 4.5.2 in a class library (so I guess full SqlClient and full framework), and I'm running against SQL Server 2016 LocalDB (maybe that's it?  I could've sworn this has worked in the past).  For reference, here's the stack trace when you run with an empty IEnumerable<SqlRecord>:

System.ArgumentException was unhandled by user code
  HResult=-2147024809
  Message=There are no records in the SqlDataRecord enumeration. To send a table-valued parameter with no rows, use a null reference for the value instead.
  Source=System.Data
  StackTrace:
       at System.Data.SqlClient.TdsParser.TdsExecuteRPC(SqlCommand cmd, _SqlRPC[] rpcArray, Int32 timeout, Boolean inSchema, SqlNotificationRequest notificationRequest, TdsParserStateObject stateObj, Boolean isCommandProc, Boolean sync, TaskCompletionSource`1 completion, Int32 startRpc, Int32 startParam)
       at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, Boolean inRetry, SqlDataReader ds, Boolean describeParameterEncryptionRequest)
[...]

PR #730, opened 2017-03-23
Fix handling of empty table-valued parameters for SqlDataRecord.

GoldenCrystal:
Hello,

When using Dapper's support for enumerables of SqlDataRecord, we encountered a bug where DBNulls were incorrectly passed for SQL Server Table Valued Parameters.
This generates an error, as SQL "NULL" is not an allowed value for table types in SQL Server. Since enumerabels of zero elements are forbidden by the SqlClient driver, the only valid value to use for expressing an empty table is the C# null value, which was previously converted to DBNull.Value by Dapper.

This PR at least partially fixes issue #649, by allowing developers to explicitly pass null as the value for TVPs in Dapper.

### DIFF
--- a/Dapper/SqlDataRecordListTVPParameter.cs
+++ b/Dapper/SqlDataRecordListTVPParameter.cs
@@ -32,7 +32,7 @@ namespace Dapper
 
         internal static void Set(IDbDataParameter parameter, IEnumerable<Microsoft.SqlServer.Server.SqlDataRecord> data, string typeName)
         {
-            parameter.Value = (object)data ?? DBNull.Value;
+            parameter.Value = (object)data;
             if (parameter is System.Data.SqlClient.SqlParameter sqlParam)
             {
                 sqlParam.SqlDbType = SqlDbType.Structured;
--- a/Dapper/SqlDataRecordListTVPParameter.cs
+++ b/Dapper/SqlDataRecordListTVPParameter.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Linq;
 
 namespace Dapper
 {
@@ -32,7 +33,7 @@ namespace Dapper
 
         internal static void Set(IDbDataParameter parameter, IEnumerable<Microsoft.SqlServer.Server.SqlDataRecord> data, string typeName)
         {
-            parameter.Value = (object)data;
+            parameter.Value = data != null && data.Any() ? data : null;
             if (parameter is System.Data.SqlClient.SqlParameter sqlParam)
             {
                 sqlParam.SqlDbType = SqlDbType.Structured;


phildude99:
I would like to see this change merged as well. We are trying to pass a TypeTable with criteria/search information in it, but want to see all records when the TypeTable is empty.

fuglede:
Yep, this one is rather annoying. For what it's worth, if one passes in the table as an empty `DataTable` rather than an `IEnumerable<SqlDataRecord>`, then it does work as expected.

Would it be a problem if empty `data` were also to cause `parameter.Value` to be `null`, so that one doesn't have to carry out that transformation explicitly? That would certainly simplify the API.

GoldenCrystal:
@fuglede I updated the PR to take into account empty enumerables, but I am not sure that this is really worth it. I fear this will cause unnecessary allocations, especially when you are already using the optimized pattern for ````IEnumerable<SqlDataRecord>````:

````csharp
public static IEnumerable<SqlDataRecord> ToSqlDataRecords(this IEnumerable<Item> items)
    => items.Any() ?
        items.ToSqlDataRecordsInternal() :
        null;

private static IEnumerable<SqlDataRecord> ToSqlDataRecordsInternal(this IEnumerable<Item> items)
{
    var record = new SqlDataRecord
    (
        new SqlMetaData("Id", SqlDbType.Int, false, true, SortOrder.Ascending, 0),
        new SqlMetaData("Name", SqlDbType.NVarChar, 100, false, true, SortOrder.Ascending, 1)
    );

    foreach (var item in items)
    {
        record.SetInt32(0, item.Id);
        record.SetString(1, item.Name);

        yield return record;
    }
}
````

This code is what you might use together with regular ````SqlParameter```` (no Dapper).
Reusing this code with Dapper calling ````Enumerable.Any()```` will be redundant, and the ````items```` enumerable will be enumerated three times for only streaming the results once.


@NickCraver Could we have your input on this, please ?


For reference, the behavior of ````SqlParameter```` is that:
* ````DBNull```` maps to the ````NULL```` SQL value
* ````null```` maps to ````DEFAULT```` (For stored procedures, this has the same effect as not passing the parameter at all)
* For table-valued parameters (TVP):
  * Only ````DataTable```` and ````IEnumerable<SqlDataRecord>```` are allowed
  * SQL ````NULL```` are not allowed, while ````DEFAULT```` is allowed

fuglede:
@GoldenCrystal Thanks. And yep, maybe I should mention that my own motivation for wanting something like this on the other side of the API was exactly that I otherwise have to do something amounting to the `items.Any() ? items.ToSqlDataRecordsInternal() : null` pattern by hand for every invocation. But of course you're right that any existing code that already does this will end up with an additional enumeration.

NickCraver:
Trying to catch up on some PRs for the next release - thanks for the fix!

PR #762, opened 2017-05-01
-  SqlMapper: MultiMapImpl: Only yield results when a mapper returns a value

jmorgannz:
There is a use case where you may want to omit rows from a query during mapping. Especially in one-to-many situations.

This change allows you to return null in a map function, and not have dapper yield those nulls out into the resulting IEnumerable<>

I couldn't test the unit test, as I don't have a local copy of MSSQL. Here's hoping the automated builds show up if I bungled it for this fork.

### DIFF
--- a/Dapper.Tests/MultiMapTests.cs
+++ b/Dapper.Tests/MultiMapTests.cs
@@ -464,6 +464,91 @@ Order by p.Id";
             }
         }
 
+        private class TestMultiMapRemovesNullMapResults_User
+        {
+            public Int32 UserId { get; set; }
+            public List<TestMultiMapRemovesNullMapResults_Post> Posts;
+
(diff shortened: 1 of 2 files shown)


NickCraver:
I appreciate the PR, but at the same time: I don't think this belongs in Dapper. This behavior shouldn't be for Dapper to decide, it should be for a user to make their determinations afterwards, e.g. `.Where(i => i != null)` or in your iterator.

Reasons:
- What if someone uses this behavior to check for `nulls`? e.g. a null is a data error we need to track down.
- What about value types? This example assumes reference types in several ways, and isn't safe. It's also inconsistent behavior, e.g. nulls rendering a `default(T)` for `default(int) == 0` would return based on the return type where a null of a reference type would not.

PR #764, opened 2017-05-04
Readme.md: Fixed link to test suite

perlun:
The file being referred to in the previous link is non-pertinent.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -428,7 +428,7 @@ Dapper has no DB specific implementation details, it works across all .NET ADO p
 
 Do you have a comprehensive list of examples?
 ---------------------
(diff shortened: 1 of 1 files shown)


NickCraver:
Thanks!

perlun:
No probs @NickCraver, thanks for a great lib. Btw, is there a way to rename fields if they have different names in the db and the POCO class? I looked but couldn't find any obvious way to do it, but perhaps I just missed something obvious.

NickCraver:
@perlun Today you'd do that in the SQL side, e.g. `SELECT MyCol as PropertyName`, but I have a proposal to add much better support on the code side for this, check out https://github.com/StackExchange/Dapper/issues/722

perlun:
@NickCraver thanks, looks good. Can use the workaround for now, until the discussion in that issue has been settled etc.

PR #772, opened 2017-05-10
Fix #771. Bumps System.Data.SqlClient

### LINKED ISSUE #771: Re: Microsoft Security Advisory 4021279
Hi,

re https://github.com/dotnet/corefx/issues/19535. 

The latest available Dapper nuget (1.50.2) has a dep on  `System.Data.SqlClient/4.1.0` , which as a transient dep to  `System.Net.Security/4.0.0` - which is mentioned in the above security annoncement. 

I see master is on `System.Data.SqlClient/4.3.0`, which also looks affected.

Should Dapper be updated..? Want a PR?

johnkors:


NickCraver:
Hmm, we haven't fixed this in master yet - we actually don't want to do this, we want to remove the dependency which is in error, see https://github.com/StackExchange/Dapper/commit/637158f5ce98f3b6d32a0fa6ee7a091e535724ff - I need to pull that over to make merging easier for tuples.

johnkors:
Ah, ok. :+1:  Closing :)

NickCraver:
@johnkors thanks for the heads up on this - we'll get it fixed today :)

Issue #295, opened 2015-06-06
Reading nullable DateTime fails when the column has a non-null value


ngeor:
I'm using MySql database and the NuGet version of Dapper. I'm reading from the database like this:

```
return _connection.Query<Person>(
    "SELECT * FROM Person WHERE PersonId=@PersonId",
    new { PersonId = 42 }).SingleOrDefault();
```

The class contains a nullable DateTime property. When the value is NULL in the database, it works fine. When it has a value, it fails complaining:

```
Error parsing column 14 (LastCommentedAt= - Object)
```

I managed to solve this by installing my own TypeHandler for `DateTime?`. I saw how Dapper.NodaTime is doing it so I created my own handler:

```
public class NullableDateTimeHandler : SqlMapper.TypeHandler<DateTime?>
{
    protected NullableDateTimeHandler()
    {
    }

    public static readonly NullableDateTimeHandler Default = new NullableDateTimeHandler();

    public override void SetValue(IDbDataParameter parameter, DateTime? value)
    {
        if (value.HasValue)
        {
            parameter.Value = value.Value;
        }
        else
        {
            parameter.Value = DBNull.Value;
        }
    }

    public override DateTime? Parse(object value)
    {
        if (value == null)
        {
            return null;
        }

        if (value is DateTime)
        {
            return (DateTime)value;
        }

        return Convert.ToDateTime(value);
    }
}
```

and I make sure I install this handler before anything else happens:

```
SqlMapper.AddTypeHandler(NullableDateTimeHandler.Default);
```

With this in place, I can read nullable dates correctly whether they have a NULL value or an actual value. I think this should be built-in functionality provided by Dapper, so I'm raising this ticket.

The stacktrace I get without my workaround looks like this:

```
MESSAGE:
System.Data.DataException : Error parsing column 14 (LastCommentedAt= - Object)
  ----> System.InvalidCastException : Cannot cast from source type to destination type.
+++++++++++++++++++
STACK TRACE:
at Dapper.SqlMapper.ThrowDataException (System.Exception,int,System.Data.IDataReader,object) <0x002e7>
at (wrapper dynamic-method) object.Deserializedf762666-acf8-4d08-832f-f40f4dc76cd7 (System.Data.IDataReader) <0x0105e>
at Dapper.SqlMapper/<QueryImpl>d__61`1<Person>.MoveNext () <0x006c3>
[...]

PR #774, opened 2017-05-12
Fix MySQL tests

bgrainger:
The MySQLTests were being run against `TestBase.connection`, which is a `SQLConnection`, not the local variable `conn`, which is the `MySqlConnection`.

Once this was fixed, two tests started failing because #295 (`InvalidCastException` is thrown when `AllowZeroDateTime=True` is present in the MySQL connection string) is still open.

Finally, I added a minor optimisation to not lazily-create `TestBase.connection` in `TestBase.Dispose`; this is primarily useful for the MySQLTests that don't ever need to open a `SQLConnection` to run.

### DIFF
--- a/Dapper.Tests/Providers/MySQLTests.cs
+++ b/Dapper.Tests/Providers/MySQLTests.cs
@@ -26,9 +26,9 @@ namespace Dapper.Tests
         [FactMySql]
         public void DapperEnumValue_Mysql()
         {
-            using (var connection = GetMySqlConnection())
+            using (var conn = GetMySqlConnection())
             {
-                Common.DapperEnumValue(connection);
+                Common.DapperEnumValue(conn);
             }
         }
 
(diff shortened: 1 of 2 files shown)


NickCraver:
Well, damn. I wish you made this 3 days ago before I started the `netstandard2.0` branch :) Thanks for the fixes! I'll get them merged over into the 2.0 stuff as well (broken for other reasons).

bgrainger:
Well, if *you* had tweeted *this* three days ago, I might have started earlier 😜: https://twitter.com/Nick_Craver/status/862862874619805696

I author [MySqlConnector](https://github.com/mysql-net/MySqlConnector/) so your tweet had got me thinking I should start support for `netstandard2.0` and tackle items such as mysql-net/MySqlConnector#182.

But first I had to see if Dapper's MySQL tests could even run against the current version of my library, which is when the yak shaving began...

NickCraver:
=P fair enough, thanks for all the work and testing here, it's much appreciated!

PR #779, opened 2017-05-15
Update ReadMe to help developers install this great library.

gtzinos:


NickCraver:
Forgive the bluntness, but why add MySQL here? Why not every other provider? Dapper is agnostic, this would seem to imply that Dapper is built for MySQL...

mgravell:
I'm aware that there are *multiple* adapters available for some backends, and that some work better than others with tools like dapper. If that is the motivation here, I'd say that is reasonable - but ideally with clearer copy. For example:

---

Installation
--------
 
Nuget: https://www.nuget.org/packages/Dapper
 
`PM> Install-Package Dapper`

Dapper works against most ADO.NET providers. Sometimes multiple ADO.NET providers are available for a particular database; it is the experience of us / our users that the following work particularly well:

- SQL Server: `SqlClient` (the inbuilt .NET SQL Server provider)
- MySQL: https://www.nuget.org/packages/MySql.Data/

etc...

---

thoughts?

NickCraver:
Yep - I agree with that. I think we should list many or none, one has had implications and shows undue favoritism we'd want to avoid. There are cases in which there are better MySQL drivers in the .NET Core world as well, so it's already problematic to only point to the official :)

gtzinos:
Sorry that i help you to have some documentation for just one of them. If you are a contributor and don't just a downloader, you can do the same for another provider. Time is yours.

NickCraver:
Closing this out per discussion above - it's also worth nothing that the version history is both pre-release recommendations (not really a fan on this) and the situation remains that the current stable release of MySQL's ADO.NET implementation is async over sync. It's the worst provider you can use for async performance right now (MySQL Connection is far better). So combining the *only* choice being also the worst performing one, I think we'd want to do a very different direction as Marc has above and list caveats with some of these...something we don't have time to keep updated at the moment as well.

PR #782, opened 2017-05-20
AddDynamicParameters repeated should work  while paramtype is DbString 

### LINKED ISSUE #731: DynamicParameters.Add cannot 'override' the property of template if value is DbString
I am using an object as parameters template,  and use the DynamicParameters.Add method to 'override' a string property to DbString so that it can use SQL Server index. 
But i got the exception : The variable name has already been declared. 

```csharp
        public static void Run()
        {
            Test(1);  //ok
            Test("1"); //ok
            Test(DateTime.Now); //ok
            Test(new DbString() { Value = "1" });  //System.Data.SqlClient.SqlException: 'The variable name '@Value' has already been declared.
        }

        public static void Test<T>(T value)
        {
            using (SqlConnection conn = new SqlConnection("Data Source=.;Initial Catalog=master;Persist Security Info=True;"))
            {
                var template = new { Value = value };
                var dp = new DynamicParameters(template);
                dp.Add("Value", value);
                conn.Execute("SELECT 1 WHERE @Value IS NOT NULL", dp);  
            }
        }
```

jiabiao:
fix #731 

This is a original test case.  Replacing 123 with DbString should work as well.
```csharp
        [Fact]
        public void Test_AddDynamicParametersRepeatedShouldWork()
        {
            var args = new DynamicParameters();
            args.AddDynamicParams(new { Foo = 123 });
            args.AddDynamicParams(new { Foo = 123 });
            int i = connection.Query<int>("select @Foo", args).Single();
            i.IsEqualTo(123);
        }
```

### DIFF
--- a/Dapper.Tests/ParameterTests.cs
+++ b/Dapper.Tests/ParameterTests.cs
@@ -1021,6 +1021,18 @@ SELECT value FROM @table WHERE value IN @myIds";
         }
 
         [Fact]
+        public void Test_AddDynamicParametersRepeatedIfParamTypeIsDbStiringShouldWork()
+        {
+            var foo = new DbString() { Value = "123" };
+
+            var args = new DynamicParameters();
+            args.AddDynamicParams(new { Foo = foo });
+            args.AddDynamicParams(new { Foo = foo });
+            int i = connection.Query<int>("select @Foo", args).Single();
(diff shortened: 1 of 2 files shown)


NickCraver:
Good fix, thank you!

PR #792, opened 2017-05-30
Allow custom ISqlAdapter to be used

tekook:
Allow to extend or overwrite the AdapterDictionary to implement custom ISqlAdapter's, for when the DefaultSqlAdapter is not useful.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -76,6 +76,22 @@ namespace Dapper.Contrib.Extensions
                 ["fbconnection"] = new FbAdapter()
             };
 
+        /// <summary>
+        /// Add or set a Custom ISqlAdapter to the AdapterDictionary to use by <see cref="GetFormatter(IDbConnection)"/>
+        /// </summary>
+        /// <param name="connectionname">Name of the connectionname</param>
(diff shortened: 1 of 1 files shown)


NickCraver:
I don't think this is something we want to add *yet*. 2.0 will have major breaking changes to this to resolve a lot of issues with integer assumptions and related bits. Anyone implementing an `ISqlAdapter` now would be burned. Note that is isn't *actually* exposed in any way a user could interface at the moment; this would change that and that's the negative impact.

tekook:
Okay I biggie. Just dir clarification:

The reason behind my pull request was the following. 
I'm using an OdbcCinnection connecting to an access database. 
Since there is no adapter for odbcconnection the default SqlAdapter gets used. 
This one does not work for access instead you need the SqlCEAdapter. 

Will 2.0 help with that Problem?

NickCraver:
Closing this out (to cleanup so many open issues, many related). We plan to overhaul this in v2 so changes in this particular PR won't make any sense. I'd say file an issue for this being in V2 specifically. There are still downsides to exposing it which we should discuss and see if it's worth it. For example, we can't ship new methods on the interface where are getting a lib update for fixes otherwise would be trivial.

Please open an issue if still important with the use cases and let's see which route we should go.

PR #795, opened 2017-06-04
Update Readme.md

cihadsevim:
[indirim kodu](https://indirimkodu.gen.tr/) prodly uses Dapper.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -435,6 +435,6 @@ Who is using this?
 ---------------------
 Dapper is in production use at:
 
-[Stack Overflow](https://stackoverflow.com/), [helpdesk](https://www.jitbit.com/web-helpdesk/)
+[Stack Overflow](https://stackoverflow.com/), [helpdesk](https://www.jitbit.com/web-helpdesk/), [indirim kodu](https://indirimkodu.gen.tr/)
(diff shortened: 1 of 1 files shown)


NickCraver:
I don't like closings things like this, so I'll expound my reasoning: 
- We can't maintain a list like this, Dapper just has too many consumers now.
- There's an uneasy feeling I have on linking to for-pay content sites (including helpdesk that was on there before).
- All of the documentation is in English, expecting consumers to consume international languages to access usage is a low percentage ask.

As such, I have removed the other monetary site (leaving only Stack Overflow) in https://github.com/StackExchange/Dapper/commit/e645420c7b71f2160510c1cb4606aa56bee66620 as well, with the call for listings.

Dapper needs a lot more documentation work, which is something I want to get to this year (I've been doing it for our other libraries). If we want a return of a section for "Who uses it?" I think that would be a much better place than at the bottom of a README anyway. I'd welcome an issue discussing that if people think it's important.

Thoughts? I don't want to close this with zero discussion (that's very uncool), and I'm happy you find Dapper to be useful.

NickCraver:
No activity here, closing out to cleanup.

PR #800, opened 2017-06-12
Markdown tables

Ffloriel:
Markdown tables instead of html
“Can be faster” in correct row


NickCraver:
Good improvement, thanks!

Issue #722, opened 2017-03-14
V2: [Column] and [Table] Attributes


NickCraver:
This would be part of the overall Dapper v2 project (#688)

There's obviously been a lot of interest in controlling column and table name mappings in a coherent and fully featured way. But all of the PRs we've seen are all over the map as to how to go about this. They're also Dapper.Contrib restricted, whereas we need something to cover things overall. For example calling `.Insert<T>()` and then calling `.Query<T>` afterwards and that not working isn't expected or awesome.

Basically, this can't be `Dapper.Contrib`, it needs to be global. This type, this member goes to this table, this column...in a controllable way. I'm picturing an API like this:

```c#
[Table("Users")]
public class User
{
    public int Id { get; set; }
    public string Name { get; set; }
    [Column("Address")]
    public string AddressSomethingCrazy { get; set; }
}
```
Now, this doesn't work for everyone. You can't add attributes everywhere and some people don't want to. They want conventions, or covering types they don't control  - so we need something else. I'm not sure what these are called, but basically we make the type and member to table and column mappings pluggable. For example, something like this:

```c#
public partial class SqlMapper
{
    public Func<Type, string> TableNameMapper { get; set; }
    public Func<Type, string> DefaultTableNameMapper { get; }
    public Func<Type, PropertyInfo, string> ColumnNameMapper { get; set; }
    public Func<Type, PropertyInfo, string> DefaultColumnNameMapper { get; }
}
```

So if your convention is init caps, underscores, emoji, or whatever really, you could define this in one place. You could call the default mapper as a fallback in your own function as well, if you just want to handle a few cases and fall back for the rest. Or, you could make your own attributes, and entirely base all mappings on whatever scheme you could come up with. As a simple example:

```c#
SqlMapper.TableNameMapper = type =>
{
	if (type == typeof(User)) return "Users";
	return SqlMapper.DefaultTableNameMapper(type);
};
```

All of these would apply across the board, to `.Query<T>`, `.Insert<T, TKey>`, `.Update<T>`, `.Get<T>`, and...well, you get the idea. They need to work everywhere, that's why it's needs to be in Dapper core and consumed all the way through `.Contrib`, etc.

Now, that deals with naming (I hope) - but there's more we can do here too. For example, is there interest in controlling defaults or something else in attributes? Naming is the biggest thing I've seen *by far*, but there may be some other areas we can solve along the same lines.

### Related Info
Related issues: #447, #463, #515, #605, #623, #670, #681, #572
[...]

PR #819, opened 2017-07-25
Fixed lowercase key problem when insert query

ismkdc:
Fixed lowercase key problem (its converts Id column to id) when insert query on postgres adapter


NickCraver:
This is a breaking change for current users (and a duplicate of several various PR versions). Please check out #722 for a discussion of what we're doing in V2 to address this.

PR #823, opened 2017-07-27
Dapper.Contrib - Overload for Update with expression for selecting columns to update

vvucetic:
Hi,

I've added new overloads for Update with expression for selecting columns to update. I don't know if it fits the idea of Dapper.Contrib. 

For example

`connection.Update(new Car() { Id = 1, Name = "Saab" }, t=> new { t=> t.Name, t.Age });`

it's a bit clumsy for arrays and lists, but I can live with it. I can't remember solution that has simpler implementation without breaking the idea of other update methods.

`connection.Update(cars, t=> new { t[0].Name });`

Appreciate comments and review.

Cheers

### DIFF
--- a/Dapper.Contrib/Readme.md
+++ b/Dapper.Contrib/Readme.md
@@ -14,7 +14,9 @@ IEnumerable<T> GetAll<T>();
 int Insert<T>(T obj);
 int Insert<T>(Enumerable<T> list);
 bool Update<T>(T obj);
+bool Update<T>(T obj, Expression<Func<T, object>> columnsToUpdate);
 bool Update<T>(Enumerable<T> list);
+bool Update<T>(Enumerable<T> list, Expression<Func<Enumerable<T>, object>> columnsToUpdate);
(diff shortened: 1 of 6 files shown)


NickCraver:
Hey @vvucetic I think this would be a good addition. The one "but" is a temporary one: we're going to have to change a lot of methods here in v2 for things like #722, and we'll need to patch this in after that's settled (likely simplifying things). Leaving this open and tagging for v2 inclusion!

jornhd:
I'd hoped to see this kind of possibility in Dapper.Contrib and was disappointed to only find this unmerged PR.
But I have a suggestion. What if you change the signature to `Update<T>(object tableKey, params Expression<Func<T, object>>[] updateExpressions)`
Then the syntax would be:
`connection.Update<Car>(1, car => car.Name = "NewName", car => car.Age = 3)`

newbe36524:
emmm, almost 3 years pass.

NickCraver:

PR #827, opened 2017-08-08
Micro optimizations for CreateParamInfoGenerator

verysimplenick:
BenchmarkDotNet results
Hi SO team!
I added some micro optimizations in your code:
1) Ordinal string comparison at property sorting and in checking for ValueTuple
2) avoid allocations for linq
3) .ToArray() replaced by list, in most cases in function
But I found some strange `if`, I did't understand why it's needed, I think we can move it up (before `foreach` for exclude `foreach` execution in some cases or just delete this `if`), because seems like already checked in coupe lines upper ( If I right understand ) Please check it.
Line about 2446

```csharp
            if (ctors.Length == 1 && propsList.Count == (ctorParams = ctors[0].GetParameters()).Length) // here we check ctors args len and count of properties
            {
                // check if reflection was kind enough to put everything in the right order for us
               // somework...
                if(ok)
                {
                    // pre-sorted; the reflection gods have smiled upon us
                    props = propsList;
                }
                else { // might still all be accounted for; check the hard way
                    var positionByName = new Dictionary<string,int>(StringComparer.OrdinalIgnoreCase);
                    foreach(var param in ctorParams) 
                    {
                        positionByName[param.Name] = param.Position; // param.Name can be not uniq? 
                    }
                    if (positionByName.Count == propsList.Count) // another check seems like before
                    {
...
```


verysimplenick:
@mgravell , @NickCraver any feedback (negative/positive)?

Dapper/SqlMapper.cs
```diff
@@ -2221,7 +2225,13 @@ public static object SanitizeParameterValue(object value)
 
         private static IEnumerable<PropertyInfo> FilterParameters(IEnumerable<PropertyInfo> parameters, string sql)
```
NickCraver: Would change this to a `List<PropertyInfo>` return so consume know it's more specific now.

Dapper/SqlMapper.cs
```diff
@@ -2419,19 +2429,27 @@ private static List<IMemberMap> GetValueTupleMembers(Type type, string[] names)
             il.Emit(OpCodes.Ldarg_0); // stack is now [command]
             il.EmitCall(OpCodes.Callvirt, typeof(IDbCommand).GetProperty(nameof(IDbCommand.Parameters)).GetGetMethod(), null); // stack is now [parameters]
 
-            var propsArr = type.GetProperties().Where(p => p.GetIndexParameters().Length == 0).ToArray();
+            var allTypeProps = type.GetProperties();
+            var propsList = new List<PropertyInfo>(allTypeProps.Length);
+			for(int i = 0; i < allTypeProps.Length; ++i)
```
NickCraver: Nit: formatting

Dapper/SqlMapper.cs
```diff
@@ -2463,13 +2481,17 @@ private static List<IMemberMap> GetValueTupleMembers(Type type, string[] names)
...
-                            Array.Sort(positions, propsArr);
-                            props = propsArr;
+                            props = propsList.ToArray();
+                            Array.Sort(positions, (PropertyInfo[]) props);
                         }
                     }
                 }
             }
-            props = props ?? propsArr.OrderBy(x => x.Name);
+            if(props == null)
```
NickCraver: Nit: formatting

Dapper/SqlMapper.cs
verysimplenick: Can we delete this function? I found only one reference in only one if.

Dapper/SqlMapper.cs
verysimplenick: default

In current Dapper code. Can you give `.editorconfig` or some formatting guides for Dapper code?

Dapper/SqlMapper.cs
verysimplenick: Ok. Missing whitespace

Dapper/SqlMapper.cs
NickCraver: Yep absolutely, we should add this. I'll try and get one in today.

Dapper/SqlMapper.cs
NickCraver: It may get more complex later...and the calling function is already fairly complex. I'd still favor leaving this separate. It'll get inlined via JIT, but actually inlining the C# will be harder to read/maintain.

NickCraver:
Updated comments - needs a rebase (all `netstandard2.0` work now merged in). If you have time to get to this today, I'll get it in the next release :)

verysimplenick:
@NickCraver ok, I do it.
What about `if` in my first post?

NickCraver:
@verysimplenick Don't worry about formatting, I'll clean it up globally and add .editoconfig at the same time, was just feedback for the future...which hopefully `.editoconfig` can obsolete for most cases.

NickCraver:
Pulling in before a formatting pass!

PR #843, opened 2017-09-06
Added From and DerivedFrom handlers

pimbrouwers:
What's up SO team?!

Have been using dapper for some time, only recently began using the SqlBuilder. In my travels I encountered several circumstances where having a ".From()" and ".DerivedFrom()" handler was quite useful.

Use case as follows:

```c#
var listBuilder = new SqlBuilder();
var pageBuilder = new SqlBuilder();

var list = listBuilder.AddTemplate("select /**select**/ /**from**/ /**where**/");
listBuilder.Select("*, row_number() over (order by CreatedAtUtc desc) as rn")
			.From("dbo.TableNeedingPaging")
			.Where("Deleted = 0");

var page = pageBuilder.AddTemplate("select /**select**/ /**derived**/ /**where**/");
pageBuilder.Select("*")
			.DerivedFrom(list, "tbl")
			.Where("rn between 0 and 20");

Console.WriteLine(page.RawSql);
```

Which ouputs:
```sql
select *
 FROM (select *, row_number() over (order by CreatedAtUtc desc) as rn
 FROM dbo.TableNeedingPaging WHERE Deleted = 0
) as tbl WHERE rn between 0 and 20
```

### DIFF
--- a/Dapper.SqlBuilder/SqlBuilder.cs
+++ b/Dapper.SqlBuilder/SqlBuilder.cs
@@ -149,5 +149,14 @@ namespace Dapper
 
         public SqlBuilder Having(string sql, dynamic parameters = null) =>
             AddClause("having", sql, parameters, "\nAND ", "HAVING ", "\n", false);
+
+        public SqlBuilder From(string sql) =>
+            AddClause("from", sql, null, "*", "FROM ", "", false);
+
+        public SqlBuilder DerivedFrom(Template sqlTemplate, string alias) =>
+            DerivedFrom(sqlTemplate.RawSql, alias);
+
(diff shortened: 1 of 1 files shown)


NickCraver:
Thank you for submitting this idea! And apologies for not getting to some of these in far, far too long as our time is stretched way thin. There are a lot of suggestions that make the SqlBuilder more complex, often _much_ more complex to understand the relationships in usages. I think the overall intent is probably to keep is simple, and if people need something far beyond that to move to something much more fully featured that's better maintained, better documented, and honestly just more intuitive/better. The class as it stands was just a utility method someone made on a whim that grew unexpectedly.

The next step on SqlBuilder is probably to split it into another repo, see #1658 for details on why we're splitting some of these out.

PR #850, opened 2017-09-27
Beginning RowVersion support, tidying, de-duplication.

RaringCoder:
I originally started making changes to support Optimistic Concurrency on Update using a new `RowVersionAttribute`. I then went down the rabbit hole and made the following changes:

 - De-duplication of code shared by the async and sync code paths.
 - Re-wrote the type property caching to a unified type, only needing to walk the properties once.
 - Standardised on arrays instead of lists for cached items, to reduce allocation sizes.
 - Simplified some code that was mutating key lists, so we don't need to take copies everywhere now.
 - Honoured a few primary method parameters such as `transaction` and `commandTimeout` (they were not always propagated downwards.

Some of the changes are groundwork for future changes I want to test:

 - Exception support for Optimistic Concurrency, where the Update statement checks if the version has changed before attempting to update and returns an error if it has, calling code can then react to this specifically.
 - `Span<T>` support for the list of `PropertyInfo`. This allows us to allocate one large array instead of multiple smaller arrays, and also may allow us to avoid using `Union` and `Except` if we create composite `Span<T>` objects.

Currently, the row version check is only on Update, though it can easily be added to Delete if desired.


RaringCoder:
Didn't intend for a future change-set to add itself to this PR.

PR #880, opened 2017-11-11
when inserting, check if type implements IEnumerable<T> to determine if it's a collection

bdrupieski:
`IDbConnection.Insert<T>` treats all generic types as collections when determining if it should insert a single entity or a list of entities. An exception is thrown when attempting to insert a single entity that has a generic type parameter.

Not all generic types are collections. This PR introduces an additional check to determine if a type is a collection by checking to see if it implements `IEnumerable<T>`.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -314,7 +314,7 @@ namespace Dapper.Contrib.Extensions
                 isList = true;
                 type = type.GetElementType();
             }
-            else if (type.IsGenericType() && type.GetTypeInfo().ImplementedInterfaces.Any(x => x.Name == "IEnumerable`1"))
+            else if (type.IsGenericType() && type.GetTypeInfo().ImplementedInterfaces.Any(ti => ti.IsGenericType() && ti.GetGenericTypeDefinition() == typeof(IEnumerable<>)))
             {
                 isList = true;
                 type = type.GetGenericArguments()[0];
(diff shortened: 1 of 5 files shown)


Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -314,7 +314,7 @@ private static string GetTableName(Type type)
                 isList = true;
                 type = type.GetElementType();
             }
-            else if (type.IsGenericType())
+            else if (type.IsGenericType() && type.GetTypeInfo().ImplementedInterfaces.Any(x => x.Name == "IEnumerable`1"))
```
NickCraver: Any reason to not use `.Any(ti => ti.IsGenericType && ti.GetGenericTypeDefinition() == typeof(IEnumerable<>))` for these checks?

bdrupieski: @NickCraver No. That's a better way of doing it. I've changed it.

`ti.IsGenericType` does not work in .NET Standard 1.3, but
`ti.IsGenericType()` does, so instead of
`.Any(ti => ti.IsGenericType && ti.GetGenericTypeDefinition() == typeof(IEnumerable<>))`
this needs to be
`.Any(ti => ti.IsGenericType() && ti.GetGenericTypeDefinition() == typeof(IEnumerable<>))`.

bdrupieski:
@NickCraver Is there any way I can help you get this PR approved and this fix deployed to NuGet? Would you like me to write additional tests?

I'm available to help you. I'm just not sure what to do. If you'd like me to write more or different tests, please let me know. Or, if there's something else, I'm happy to assist.

Thank you for contributing your time to this project and for helping me to do my job.

NickCraver:
Sorry for the delays, we're just understaffed on these projects. Looking good now, thanks for the contribution!

bdrupieski:
@NickCraver Thank you!

PR #886, opened 2017-11-26
safely cast IEnumerable<PropertyInfo> as IList instead of array to save array creation on every insert

bdrupieski:
In the `ISqlAdapter.Insert` and `ISqlAdapter.InsertAsync` methods in Dapper.Contrib the `IEnumerable<PropertyInfo> keyProperties` parameter is safely cast to `PropertyInfo[]`, and if it isn't `PropertyInfo[]` a new array is created. 

At least in this package, a `List<PropertyInfo>` is always passed. This means the safe cast to `PropertyInfo[]` will return null and a new array is created on every insert. By changing this to safely cast to `IList<PropertyInfo>` we can prevent allocating this array.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -339,8 +339,8 @@ public partial class SqlServerAdapter
         if (first == null || first.id == null) return 0;
 
         var id = (int)first.id;
-        var pi = keyProperties as PropertyInfo[] ?? keyProperties.ToArray();
-        if (pi.Length == 0) return id;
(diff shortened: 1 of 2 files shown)


NickCraver:
In this case, the root issue isn't intuitive (this whole section really needs a rewrite...). While this is an optimization, it's also a breaking change. Unfortunately, some things in the pipe *modify* that collection (several issues around this) and that's why .ToList() or arrays are created in various places. We need to change how all this works (but haven't had a lot of time for Contrib, or the main lib for a while honestly). But, we can't put this change in...it'll result in all sorts of fun behavior :(

Hope that helps, and sorry we've been under to much OSS water to keep anywhere near up to date on PRs :(

bdrupieski:
Ah, OK. Thanks for responding with an explanation!

PR #887, opened 2017-11-26
when updating and deleting, check if type implements IEnumerable<T> before using generic type as entity's type

bdrupieski:
This is a follow-up to PR #880. I didn't know the same issue affects Update and Delete. 

I'm sorry.


bdrupieski:
I updated this to account for #930. 

It requires the fix for Insert/InsertAsync so the tests pass. When https://github.com/StackExchange/Dapper/pull/948 is merged I'll rebase on master so that it works.

bdrupieski:
I rebased. This should be good to go now.

NickCraver:
Getting this in for the next release - thanks!

PR #899, opened 2017-12-13
Fix case sensitive problem of key properties during insert on PostgresSQL

smlk:
Without quote on key column, it will be treated as lower case in case of Postgres.
So, I add double quote to column names and remove ToLower() call in order to use camel spelling in both of code and databse.


NickCraver:
This is a duplicate of several other issues/PRs - note that this is a breaking change. We plan to address this in v2 via the #722 path though. We just cannot change it in this way as it'll break everyone depending on non-case-sensitive behavior today. Please take a look at #722 for how we plan to address this!

PR #900, opened 2017-12-14
Fixed issue with empty sqldatarecord list as a table valued parameter.

hstefanov:
System.ArgumentException is thrown when passing empty SqlDataRecord list as an anonymous object or as table-valued parameter to Dapper. Added two unit tests.


hstefanov:
@mgravell, can u please review the changes and comment if there is any other way of avoiding this exception. Thank you.

NickCraver:
@hstefanov I see what we're trying to avoid and agree it'd be nice not to throw. On the fix: this changes the behavior for the `null` case (an equivalent `else` that does it would not have this difference). I'm not familiar enough with TVP behavior to know if that's a breaking change or not...but since we can not break, we should strive not to.

arinwt:
As it was before, `null` becomes `DbNull` which _does_ throw an exception anyway.
It would be nice to get this fix in place; I'm fine with `null` throwing an exception, but I don't want an empty collection throwing.  Right now we are working around it with a similar fix to this PR.

NickCraver:
@hstefanov I'm just catching up here - is there any chance you could merge master in (I don't want to squash and loose your name on the contribution). The issue itself was fixed in a duplicate PR but your tests here are great to ensure we don't regress and I've love to pull them in if you're willing! Or a new PR with tests simply pasted in...whichever you find easier I'd very much welcome!

hstefanov:
@NickCraver thank you for the comments. I think the PR is ready for merge.

PR #906, opened 2017-12-19
PostgreSQL: make table names case-sensitive

henriquemotaesteves:
To make identifiers (table, column, etc) case-sensitive, PostgreSQL request double quotes around them. Currently, the adapter only make columns case-sensitive. This pull request add support to make tables case-sensitive too.

The test suite was executed against the following databases with success:

Database | Version
-- | --
SQLServer | 2016 SP1
PostgreSQL | 9.6.3
MySQL | 5.7.20
SQLite | 3.21.0


NickCraver:
Unfortunately, this would be a breaking change (which we can't do in V1) - check out #722 for discussion around plans for V2 here. I think we need to make this configurable in an easy way and would appreciate thoughts there.

PR #913, opened 2018-01-04
Make DapperRow implement IReadOnlyDictionary<string, object>

enarax:
.NET 4.5 Introduced `IReadOnlyDictionary<string, object>` which is kind of a subset of `IDictionary<string, object>`. This interface is really useful especially in immutable object models. However, for legacy reasons `IDictionary `does not inherit from `IReadOnlyDictionary `unfortunately. 

This pull request **makes the internal DapperRow class implement IReadOnlyDictionary<string, object>** so we can simply pass a row into an API taking that kind of parameter, without the performance penalty of copying the row into a new Dictionary.

### DIFF
--- a/Dapper/SqlMapper.DapperRow.cs
+++ b/Dapper/SqlMapper.DapperRow.cs
@@ -10,6 +10,7 @@ namespace Dapper
         private sealed class DapperRow
             : System.Dynamic.IDynamicMetaObjectProvider
             , IDictionary<string, object>
+            , IReadOnlyDictionary<string, object>
         {
             private readonly DapperTable table;
             private object[] values;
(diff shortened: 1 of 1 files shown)


Dapper/SqlMapper.DapperRow.cs
```diff
@@ -208,6 +209,41 @@ private object SetValue(string key, object value, bool isAdd)
...
+            }
+
+            object IReadOnlyDictionary<string, object>.this[string key]
+            {
+                get { TryGetValue(key, out object val); return val; }
+            }
+
+            IEnumerable<string> IReadOnlyDictionary<string, object>.Keys
+            {
+                get { return this.Select(kv => kv.Key); }
```
mgravell: possibly `return table.FieldNames;` ? although this would expose the field-names to abuse... slightly probematic

Dapper/SqlMapper.DapperRow.cs
```diff
@@ -208,6 +209,41 @@ private object SetValue(string key, object value, bool isAdd)
...
+            }
+
+            IEnumerable<object> IReadOnlyDictionary<string, object>.Values
+            {
+                get { return this.Select(kv => kv.Value); }
```
mgravell: possibly `return values;`, although similar concerns to `Keys` - probably best to leave this "as is", too!

Dapper/SqlMapper.DapperRow.cs
mgravell: edit: yeah, this (my code, not yours) also wouldn't take dead values into account; probably best to leave it alone

enarax: Not a bad idea. I pretty much copied this from the existing IDictionary implementation, however so it really might be best to leave alone, for the sake of consistency.
(The only difference in the implementation is me skipping the .ToArray() at the end because for this interface it is enough to return an IEnumerable)

NickCraver:
Merging in :)

PR #922, opened 2018-01-15
Update dapper contrib library with ignore update attribute

MichaelZaslavsky:
Hi Dapper Team,

The idea is to ignore properties while calling update method and not ignore them while calling insert method.

Hope you will consider including this ASAP.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -211,7 +211,8 @@ namespace Dapper.Contrib.Extensions
             var allProperties = TypePropertiesCache(type);
             keyProperties.AddRange(explicitKeyProperties);
             var computedProperties = ComputedPropertiesCache(type);
-            var nonIdProps = allProperties.Except(keyProperties.Union(computedProperties)).ToList();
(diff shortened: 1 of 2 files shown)


NickCraver:
Unfortuantely, this would further complicate the columns issue overall by adding yet another attribute (and not covering the non-attribute use case). Could you please add feedback/use cases to #722?

MichaelZaslavsky:
I already saw this issue, but it will be only on dapper v 2.0 which will not happen soon.
It would be nice if meantime we could use some solution to this problem.
Could this code be accepted for now, until dapper v 2.0 releases?

NickCraver:
@MichaelZaslavsky I do not want to do that because it's not a good upgrade path for users. If we're doing ignore update then it's not "change X to Y" as upgrade instructions, it's a complete inversion of the model for what's included. IMO, that is a very confusing thing and not good from a upgrade instructions point of view. We want to keep breaks at a minimum for v2, the more added before, the more we break on users just after investment which is also not awesome.

NickCraver:
Closing out to cleanup - will factor this into #722, but unsure if we'll add it as a mechanism in-box or likely let users extend something as an opt-out path (using Table/Column attributes by default). The extensibility will be there to support it, though.

Issue #933, opened 2018-02-01
'Null object cannot be converted to a value type.' when using nullable type on interface


JaseRandall:
Using Dapper.Contrib (for the change tracking) and SQLite.

I have an interface and class... here's a cut down version:
```
public interface IExample
{
   public int Id { get; set; }
   DateTime? Dob { get; set; }
}

public class Example : IExample
{
   public int Id { get; set; }
   public DateTime? Dob { get; set; }
}
```

**This works:**
```
public IEnumerable<IExample> GetAll()
{
    return Connection.GetAll<Example>();
}
```

**This does not work:**
```
public IEnumerable<IExample> GetAll()
{
    return Connection.GetAll<IExample>();
}
```

It throws 'Null object cannot be converted to a value type.' which obviously isn't right since I'm using a nullable type.  I've removed all nulls from the database and it made no difference.  If I change DateTime? to DateTime then it works.  I could work around it by assinging DateTime.MinValue however I'd prefer not to.

Is this a bug in Dapper.Contrib or something I'm doing wrong?

NickCraver:
When you're trying to deserialize into `IExample`, what is the type of the object in the list? When you call `.GetType()` on an object, what would you expect the result to be?

I ask the above to illustrate why deserializing to an interface (or an abstract class) wouldn't really work...because you can't instantiate it. Does that make sense?

creative999:
I've had more time today to look into this and have worked out why it occurs when I use the interface.

What you've said does make sense however the issue isn't the interface.  Dapper.Contrib has code to create an object from an interface:

```
if (!type.IsInterface()) return connection.Query<T>(sql, null, transaction, commandTimeout: commandTimeout);

var result = connection.Query(sql);
var list = new List<T>();
 foreach (IDictionary<string, object> res in result)
 {
    var obj = ProxyGenerator.GetInterfaceProxy<T>();
    foreach (var property in TypePropertiesCache(type))
    {
        var val = res[property.Name];
        property.SetValue(obj, Convert.ChangeType(val, property.PropertyType), null);
     }
     ((IProxy)obj).IsDirty = false;   //reset change tracking and return
     list.Add(obj);
}
```
I've never used reflection, am pretty junior in terms of programming experience and I don't program professional however, I think the issue is caused by:
` property.SetValue(obj, Convert.ChangeType(val, property.PropertyType), null);`
which can't set the value of a nullable type.

[...]

PR #936, opened 2018-02-03
Change to GetAll to allow the usage of a nullable type in T when T is an interface… #933

JaseRandall:
This pull request is in response to an issue that I raised:

> 'Null object cannot be converted to a value type.' when using nullable type on interface #933

Basically I had DateTime? and was using an interface so that I could use the change tracking in Dapper.Contrib.

Without this modification, the code throws 

> System.InvalidCastException: 'Invalid cast from 'System.String' to 'System.Nullable`1[[System.DateTime, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]'.'

OR

>  'Null object cannot be converted to a value type.' 

depending on whether the value from the (SQLite) database has a value or is null.

My suggested change:
checks for null and if so, continues;
otherwise if the type is nullable, it obtains the underlying property and assigns the value to the underlying property;
otherwise assign the value to the property as per the original code.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -249,9 +249,9 @@ namespace Dapper.Contrib.Extensions
                 foreach (var property in TypePropertiesCache(type))
                 {
                     var val = res[property.Name];
+                    if (val == null) continue;
                     if (property.PropertyType.IsGenericType() && property.PropertyType.GetGenericTypeDefinition() == typeof(Nullable<>))
                     {
-                        if (val == null) continue;
                         var genericType = property.PropertyType.GetGenericArguments()[0];
                         property.SetValue(obj, Convert.ChangeType(val, genericType), null);
                     }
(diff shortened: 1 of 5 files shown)


NickCraver:
Can you please add a test for this to ensure we don't regress? Happy to merge in after we make sure we're all covered from a recurrence :)

Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -202,7 +202,16 @@ private static PropertyInfo GetSingleKey<T>(string method)
                 foreach (var property in TypePropertiesCache(type))
                 {
                     var val = res[property.Name];
-                    property.SetValue(obj, Convert.ChangeType(val, property.PropertyType), null);
+                    if (val == null) continue;
+                    if (property.PropertyType.IsGenericType() && property.PropertyType.GetGenericTypeDefinition() == typeof(Nullable<>))
+                    {
+                        var genericType = property.PropertyType.GetGenericArguments()[0];
```
NickCraver: Since .GetGenericArguments() allocates an array every time, instead use: `Nullable.GetUnderlyingType(property.PropertyType)` to do this much cheaper. Same below :)

JaseRandall:
@NickCraver I realised that I missed including the async Get & GetAll methods. I've added the additional code to those methods and have added a couple of tests. I'll admit that this is the first time I've used tests (I'm such an amatuer!!) and so while I think they are adequete, I'm not certain of that.  Hopefully it's all up to scratch now but if not, please let me know.

NickCraver:
@JaseRandall looking awesome, thank you so much for your first contribution! Excellently done :)

PR #942, opened 2018-02-08
Dapper.Contrib - a new DbManager wrapped class for  coding experience like DbContext of EntityFramework and MyBatis 3

BeanHsiang:
Hi,

I've added a new DbManager wrapped class in Dapper.Contrib.
The description markdown file is : [DbManager for dapper](https://github.com/BeanHsiang/Dapper/blob/master/Dapper.Contrib/Readme%20-%20DbManager.md)

This is my first pull request on github. Appreciate comments and review.


NickCraver:
While we really do appreciate the effort here, I discussed this with Marc and we're both in agreement this should be in a separate project/NuGet. It's simply not in the goals of Dapper to be a data context or similar. But, some people do want such a thing! It just doesn't align with the project goals of being simple and fast. It's a level above where we always aim to be.

If you do publish this on NuGet as another lib please point us at it so we can keep an eye on progress :)

PR #944, opened 2018-02-12
fix equal tests

DominikHerold:


NickCraver:
hahaha my fault - the catch and fix is really appreciated!

PR #952, opened 2018-02-20
Allow parameters passed by ICustomQueryParameter/IDynamicParameters to be reused

jamescrowley:
**Background**

We wrap read-only Dapper calls in Polly to allow for retry strategies. We also use the IDynamicParameters and ICustomQueryParameter interface. Because we create our own db parameters, when re-trying, you get an exception "The SqlParameter is already contained by another SqlParameterCollection" unless you re-create the SqlParameters within the retry code block.

EF had a similar issue, and took this approach: https://github.com/aspnet/EntityFramework6/pull/142, fixing https://github.com/aspnet/EntityFramework6/issues/81

**This PR**

Prior to disposing of any command that is created, this PR now calls clear on the parameters collection. I believe this to be safe, given the library is about to dispose of the command in all these instances. 

The new tests do not hit all the impacted methods, but have the basic requirement - not sure how you'd want to approach that?


NickCraver (approved):
Looking good here - thanks for the fix and the pointers at EF!

Issue #958, opened 2018-03-01
Default Parameter Type for DynamicParameters


roend83:
I would like to globally override the default parameter type used for DynamicParameters queries when the parameter value is NULL.  It seems to be defaulting to NVarChar (DbType.String) in Sql Server right now, and I would like it to default to VarChar (DbType.AnsiString).  I'm happy to send a PR to support this, but I want to make sure this is something you'd actually want first.

mgravell:
There are already options to control the string format at both the
per-parameter level (DbString) and globally (there's a static property
somewhere). Do these not work for your scenario?

roend83:
No, I'm already using the static property to control this for all strings.  The issue that I'm running into is unique to using the DyanamicParameters class to pass parameters.  Something like this:
```
string myValue = null;
DynamicParameters param = new DynamicParameters();
param.Add("myValue", myValue);
```
Because myValue is NULL and we can't do NULL.GetType(), the DbType on the command parameter will not be set and thus default to NVARCHAR in Sql Server.  I know that I can go in and change all my code that uses DynamicParameters to explictly pass a DbType when I pass strings, but I'd rather not have to do that now or remember that going forward.
I can think of 2 ways to gracefully solve this problem:
1. Add a static property somewhere to allow me to set the default parameter type when it can't be inferred from a null value
2. Change DynamicParameters.Add to be a generic method (i.e. void Add\<T\>(string name, T value, ...) and store the type away to do the DbType lookup later.
Does this make sense?

mgravell:
Option 2 has some appeal and I could "get with". Option 1 sounds risky. If
you have ideas on implementing option 2, it sounds viable.

roend83:
I took a stab at this.  It seems to work in my application.  I can't run the tests on my machine though.  Are there instructions for the test setup somewhere?

NickCraver:
See #963 for progress here.

rodri-oliveira-dev:
I ran into a related version of this problem, particularly when the CLR value is null and therefore doesn’t carry enough information to express the intended SQL Server contract.

Instead of changing Dapper’s global default, I experimented with making the SQL Server type explicit at the call site:

new
{
    Code = SqlParam.VarChar(null, 30)
}

Even though the CLR value is null, the resulting SqlParameter is still explicitly configured as SqlDbType.VarChar with Size = 30.

I extracted this approach into an external library:

https://github.com/rodri-oliveira-dev/Dapper.TypedParameters

[...]

PR #963, opened 2018-03-07
Generic Add Method for Dynamic Parameters

roend83:
Addresses #958

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -100,6 +100,27 @@ namespace Dapper
         /// <param name="dbType">The type of the parameter.</param>
         /// <param name="direction">The in or out direction of the parameter.</param>
         /// <param name="size">The size of the parameter.</param>
+        public void Add(string name, object value, DbType? dbType, ParameterDirection? direction, int? size)
+        {
+            parameters[Clean(name)] = new ParamInfo
+            {
+                Name = name,
+                Value = value,
+                Type = value?.GetType(),
+                ParameterDirection = direction ?? ParameterDirection.Input,
+                DbType = dbType,
+                Size = size
+            };
+        }
+
+        /// <summary>
+        /// Add a parameter to this dynamic parameter list.
+        /// </summary>
+        /// <param name="name">The name of the parameter.</param>
+        /// <param name="value">The value of the parameter.</param>
+        /// <param name="dbType">The type of the parameter.</param>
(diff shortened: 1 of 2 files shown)


NickCraver:
Unfortunately this would be a breaking change. While we can *add* the generic versions, removing the previous ones is not an option until the next major release. Can you please add them back and a test that shows these fixing the issue?

roend83:
We actually noticed an issue with the missing non-generic Add methods and added them back.  

I'm fine adding a test for the generic methods.  Can you tell me what I have to do to run the tests?

NickCraver:
@roend83 Ah yes, git is just doing a horrible job of diffing here - to run the tests just push :) Take advantage of the build server running them for you on this PR :)

roend83:
@NickCraver I added a couple tests that call the 2 new generic methods.  Let me know if these look ok.

Dapper/DynamicParameters.cs
```diff
@@ -128,6 +150,32 @@ public void Add(string name, object value = null, DbType? dbType = null, Paramet
...
+        /// <param name="size">The size of the parameter.</param>
+        /// <param name="precision">The precision of the parameter.</param>
+        /// <param name="scale">The scale of the parameter.</param>
+        public void Add<T>(string name, T value = default(T), DbType? dbType = null, ParameterDirection? direction = null, int? size = null, byte? precision = null, byte? scale = null)
+        {
+            parameters[Clean(name)] = new ParamInfo
+            {
+                Name = name,
+                Value = value,
+                Type = value?.GetType() ?? typeof(T),
```
NickCraver: It *feels* like this should be `typeof(T)` always, since we're explicitly supplying the type information. In the case of a base type and a child type, `T` can be the base type and `value` can be of the child type in a valid generic call...so these can differ.

Thoughts?

dude0001:
@NickCraver & @mgravell Are there any plans to pull this in soon?

ghost1face:
Hey all, I just ran across this PR which looks to resolve an issue I've just run into:

```cs
DateTime? dateParam = default;
var params = new {  dateParam = dateParam };

var result = connection.ExecuteScalar<object>("SELECT 1 FROM table WHERE date = @dateParam", params);
```

I'm using a Postgres database, and receive an exception:  `could not determine data type of parameter $1`.   

Of course this is a silly example, but it gets the point across.  Are there any plans for this PR to make it into the code base soon?

Patman0021:
@NickCraver @mgravell 
ping?

Issue #976, opened 2018-03-19
Proposal: Allow Query() to support polymorphic loads


pjhuck:
Dapper has documented support for [Type Switching Per Row](https://github.com/StackExchange/Dapper#type-switching-per-row), however, this is somewhat cumbersome to use. What I'd really like is the ability to use the standard `Query<>()` api to load a list of polymorphic objects. This would also allow me to use these types with things like [multi mapping](https://github.com/StackExchange/Dapper#multi-mapping). The code I'd like to write would look like:

```c#
IEnumerable <Shape> shapes = connection.Query<Shape>("SELECT * FROM Shapes"); 

```

We've hacked a fork of Dapper to allow this -- much of the code can be done outside the core dapper classes, however, it requires 2 changes to SqlMapper: 

1. Support for "custom deserializers" -- The ability to register a custom `Func<IDataReader, object>` for a specific base type. This Func would look at the row and produce the deserializer for the concrete type. This requires a change in [`GetDeserializer`](https://github.com/StackExchange/Dapper/blob/07fe543f1530b33818c9ce989bf0672e145c954d/Dapper/SqlMapper.cs#L1785) to check for the existance of the custom deserializer.
2. Stop using `CommandBehavior.SequentialAccess` -- because the custom deserializers need to examine the discriminator column out of order to determine which subclass to deserialize.  Not sure the impact this change has, but if it's significant, perhaps this could be made configurable. 

Here's an example of custom deserializer which looks at a discriminator column to produce a deserializer for the proper subclass:

```c#
internal Func<IDataReader, object> GetShapeDeserializer(Type type, IDataReader reader, int startBound, int length, bool returnNullIfFirstMissing)
{
    var typeMap = new Dictionary<object, Type> {
        [ShapeType.Circle] = typeof(Circle),
        [ShapeType.Square] = typeof(Square),
        [ShapeType.Triangle] = typeof(Triangle)
    };
    var deserializers = typeMap.Values.ToDictionary(
        t => t,
        t => SqlMapper.GetTypeDeserializer(t, reader, startBound, length, returnNullIfFirstMissing));

    if (length == -1)
        length = reader.FieldCount - startBound;

    return r => {
        int idx = r.GetOrdinal("Type");
        object val = r.GetValue(idx);
        if (val == DBNull.Value)
            return null;

        Type childType = typeMap[val];
        return deserializers[childType](r);
    };
}
```

[...]

PR #977, opened 2018-03-20
proof of concept for polymorphic query

pjhuck:
Furthering the discussion from #976, here's a proof of concept of the changes. There are simple tests for proving it works and supports horizontal partitioning. 

In theory, the only public API change could be the SqlMapper.RegisterPolymorphicLoader method.

### DIFF
--- a/Dapper.Tests/PolymorphicTests.cs
+++ b/Dapper.Tests/PolymorphicTests.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Linq;
+using Xunit;
+
+namespace Dapper.Tests
+{
+    public class PolymorphicTests : TestBase
+    {
+        class Shape
+        {
+            public int Type { get; set; }
+            public int NumberOfSides { get; set; }
+        }
+
+        class Circle : Shape
+        {
+            public double Radius { get; set; }
+        }
+
+        class Triangle : Shape
+        {
+            public double Angle { get; set; }
+        }
+
+        static PolymorphicTests()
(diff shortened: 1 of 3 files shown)


NickCraver:
I'm noticing at a glance here that all of the `SequentialAccess` is removed - this would be a performance regression since several primary providers have specialized low-allocation code paths that are taken when this is a known behavior. For example, [here's `SqlClient -> SqlDataReader`](https://github.com/dotnet/corefx/blob/c01cee026c08661ae6014bbdf04b415339279216/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlDataReader.cs#L1515).

IMO, this needs to be done without such a regression in play, and I'm not sure how to do that barring (in the provider or on-our-side) intermediate collections for rows for a very minority use case. Otherwise it's a performance burden on everyone for the benefit of very few...that's a hard API tradeoff to make.

pjhuck:
yeah, I figured they were there for performance reasons. Agree that you don't wan't to take that hit for all. I can see a few ways to avoid it:

1. As @mgravell noted, we could require the discriminator column to be first. I'd like to avoid this, as it makes the API much less usable since you'd have to be very careful with your SQL.
2. We could make the make the CommandBehavior somehow configurable for the type. This PR already introduces the "custom deserializer" as state for a given type. Seems reasonable to keep custom CommandFlags there. So the performance hit would only be taken for the necessary types.

If your open to the concept of 2, then I can update this proof of concept with that.

PR #981, opened 2018-03-24
GetTableName() method improvement.

DevelAx:
Handling English plurals ending in "-es" in `GetTableName()` method.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -312,8 +312,8 @@ namespace Dapper.Contrib.Extensions
                 }
                 else
                 {
-                    if (TableNameMappingOptions.HandleEsPluralEnding)
-                        name = type.Name + (_esPluralsEnding.Any(ends => type.Name.EndsWith(ends, StringComparison.OrdinalIgnoreCase)) ? "es" : "s");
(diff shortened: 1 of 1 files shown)


DevelAx:
Something like additional `static bool` field in `SqlMapperExtensions` class?

mgravell:
I don't normally dip in the "contrib" package much, so I can't advise on
the "right" way to toggle it without looking, but: it can't just change
what happens unilaterally - that would break anyone upgrading.

DevelAx:
Ok, you're right, I'll try to make it another way.

NickCraver:
@DevelAx we want to fix this globally in v2, so I'm hesitant to add a bunch of options here (especially in a new way, when they're moving *libraries*, and mismatched upgrades are confusing to move about since both may exist at once). Can you please take a look at #722 and see if the proposal does what you want?

DevelAx:
@NickCraver As far as I could understand there is a dispute about attributes and delegates but not about improving/correction the default plurals ending in "-es", although this pull-request is referenced there (not sure what it means).

NickCraver:
@DevelAx the default algorithm for table names (in absence of specific attributes) would be user-settable. We could have a few default `Func<>` available, or you could do whatever you wanted when specifying your own. Let's focus there and make sure the overall solution works. Right now we've got code, issues, and some hacks in dozens of places...v2 is time to clean house and make this much more usable, intuitive, and simpler.

PR #984, opened 2018-03-29
Removing unjustified use of AppendFormat in StringBuilder class

DevelAx:


NickCraver:
@DevelAx yep, very unrelated things :) If you push to the same branch, it adds to the PR.

DevelAx:
@NickCraver so, what better way to split it now?

NickCraver:
@DevelAx likely easiest to start a new branch and put only the `StringBuilder` changes in. Alternatively, you can make a new branch and git cheery pick the single commit over...and then open a PR with that branch.

Generally speaking, for PRs (to avoid this) you'll want to branch immediately and do your work there instead of on your master, unless you *really* intended on ever making 1 change. That makes it easier to update your master later as well.

DevelAx:
@NickCraver I created 2 separate branches PRs:
https://github.com/StackExchange/Dapper/pull/988
https://github.com/StackExchange/Dapper/pull/989

PR #987, opened 2018-03-30
Tweaking error messages and supporting IDbConnection implementations …

benmccallum:
…that return a DbCommand.


benmccallum:
Extending Nick's error messages with some suggestions. Addresses: https://github.com/StackExchange/Dapper/issues/757

NickCraver:
Fixed merge, yanking in cleaner create call :)

PR #988, opened 2018-03-30
Invalid cast from 'System.Int32 to Enum

DevelAx:
Splitted from 
https://github.com/StackExchange/Dapper/pull/984


NickCraver:
Can you please add some unit tests illustrating what this is attempting to fix?

NickCraver:
If we want to revisit this please ping - right now this is a change without any context that shifts behavior.

PR #989, opened 2018-03-30
Removing unjustified use of AppendFormat in StringBuilder class

DevelAx:
Splitted from https://github.com/StackExchange/Dapper/pull/984


NickCraver:
Good fix, thanks!

PR #991, opened 2018-04-02
Discarding duplicate fields when mapping

DevelAx:
The issue description is here:
https://stackoverflow.com/questions/49601732/dapper-query-incorrectly-maps-the-entity-id

It fixed the issue but I haven't delved deeply into that code, so I cannot be absolutely sure that it will not lead to any bugs.

I think it's better to fix it anyway, because sometimes unintentionally anyone can just forget to limit the `SELECT` and get a false result without even knowing about it. And it will be great if it is prevented by default. Also, it reduces the number of mappings of unnecessary fields.

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -3097,7 +3097,7 @@ namespace Dapper
                 throw MultiMapException(reader);
             }
 
-            var names = Enumerable.Range(startBound, length).Select(i => reader.GetName(i)).ToArray();
+            var names = Enumerable.Range(startBound, length).Select(i => reader.GetName(i)).Distinct().ToArray();
 
             ITypeMap typeMap = GetTypeMap(type);
 


DevelAx:
Probably this fix will break cases with real joining when there could be similar fields.

ajisactive:
I personally think the way Dapper is working is correct. Distinct(). (the change you made) will fetch one field out of multiples. But this will create more trouble than solution. Which field is selected? Why that? Why not other? Too many if-and-but.

The better solution I can imagine of is that Dapper should throw an exception if duplicate column name is found. That way, user know in advance what wrong is going on. User then can correct the code in whatever best way for given situation.

DevelAx:
@ajisactive I agree with any solution which will be better than a quiet setting of the incorrect value.

NickCraver:
This is a breaking behavior change that also breaks many tests. I don't agree with the behavior change either, but regardless it's not something we can take without a major version change.

If you want to map multiple types, query for multiple types with a proper `splitOn` (which defaults to `Id`). You should be using `Query<PostTag, PostTagIndex, WhateverResultYouWant>` instead of `Query<T>`.

Closing this out to cleanup :)

PR #994, opened 2018-04-04
Change Insert<T>(...) return type from [long] to [int]

DevelAx:
Couldn't find any reason why the return type of [Insert<T>(...)](https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.cs#L324) method is `long`, while the values obtained from the internal [adapter.Insert(...)](https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.cs#L381) & [connection.Execute(...)](https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.cs#L388) calls are of type `int`. 
The return type of the similar synchronous  [InsertAsync<T>(...)](https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.Async.cs#L139) method is `int` too.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -321,7 +321,7 @@ namespace Dapper.Contrib.Extensions
         /// <param name="transaction">The transaction to run under, null (the default) if none</param>
         /// <param name="commandTimeout">Number of seconds before command execution timeout</param>
         /// <returns>Identity of inserted entity, or number of inserted rows if inserting a list</returns>
-        public static long Insert<T>(this IDbConnection connection, T entityToInsert, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
+        public static int Insert<T>(this IDbConnection connection, T entityToInsert, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
         {
             var isList = false;
 


NickCraver:
This is a breaking API change, as stated in other PRs...we can't just do those. The methods will be changing to generics in v2 anyway, but at the moment: this breaks a lot of people and can't be done without a major version change.

DevelAx:
@NickCraver  Ok, I just thought that people call it with casting to `int` anyway.
BTW, aren't you going to implement something like this?
```
public static async T Insert<T>(this IDbConnection connection, T entity, IDbTransaction transaction = null, int? commandTimeout = null, ISqlAdapter sqlAdapter = null) where T : class
{
    PropertyInfo keyProperty = SingleKeyProperty<T>();
    int id = connection.Insert(entity, transaction, commandTimeout, sqlAdapter);
    keyProperty.SetValue(entity, id);
    return entity;
}
```
In order to set `Id` to the entity automatically when necessary and return the entity instead of returning just `Id`.

NickCraver:
> Ok, I just thought that people call it with casting to int anyway.

I know it seems like that wouldn't break much on the surface, but methods in any lib using this would break - there are source breaks (which this still is) and binary breaks (which it definitely is).

For v2, I'm thinking something like `TKey Insert<T, TKey>(`, and additionally having a singular `long Insert<T>` for convenience that's just a wrapper that calls through for easier upgrades. I'm going to close this out to cleanup but I'll write up an issue on v2/v3 methods for generic rewrites here.

PR #1003, opened 2018-04-15
Column properties

marygabry:
Added functionalities to use the ColumnAttribute, if defined, from net451 in Get, Insert, Update and Delete extensions methods.


Dapper.Contrib/SqlMapperExtensions.Async.cs
```diff
@@ -24,18 +24,34 @@ public static partial class SqlMapperExtensions
         /// <returns>Entity of T</returns>
         public static async Task<T> GetAsync<T>(this IDbConnection connection, dynamic id, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
         {
+            String keyName = "";
```
chriswolf-nrg: You don't need this variable.

Dapper.Contrib/SqlMapperExtensions.Async.cs
```diff
@@ -24,18 +24,34 @@ public static partial class SqlMapperExtensions
...
+
             if (!GetQueries.TryGetValue(type.TypeHandle, out string sql))
             {
                 var key = GetSingleKey<T>(nameof(GetAsync));
                 var name = GetTableName(type);
+                keyName = key.Name;
+                String keyColumnName = GetDBColumnName(key);
 
-                sql = $"SELECT * FROM {name} WHERE {key.Name} = @id";
+                sql = $"select * from {name} where {keyColumnName} = @" + keyName;
```
chriswolf-nrg: You can just do this:
`sql = $"select * from {name} where {keyColumnName} = @{key.Name}";`

marygabry: This has been received and checked in. Thank you!

NickCraver:

PR #1004, opened 2018-04-16
[Delete] & [Update] methods fixed for generic entities.

DevelAx:
`Delete` & `Update` methods threw the exception _"Entity must have at least one [Key] or [ExplicitKey] property"_ for a generic entity like this:
```
    public class GenericUser<TKey>
         where TKey : IEquatable<TKey>
    {
        [Key]
        public TKey Id { get; set; }
        public int Age { get; set; }
    }
```
I'm not sure that it's a good idea to expose a new method from the `SqlMapper`, probably it needs to be refactored: https://github.com/StackExchange/Dapper/pull/1004/files#diff-48000d1adc3fd2af62b908b6d189dc59R511

P.S.
I didn't check the `Insert(Async)` methods and others from the `Dapper.Contrib`, there is a chance they have the same issue too.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -219,7 +219,7 @@ namespace Dapper.Contrib.Extensions
             {
                 type = type.GetElementType();
             }
-            else if (type.IsGenericType())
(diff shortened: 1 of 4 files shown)


NickCraver:

PR #1007, opened 2018-04-18
Normalize postgresql names (table, where paremeter...)

Karnaukhov:


NickCraver:
Unfortunately, this isn't something we can do. See issue #722 for discussion. In short: if we take this change, it will not affect some users and break others. People who have names that would match today upgrade their library and their queries no longer run.

We want to offer that control in v2, but it's a bad breaking change we can't just implement - I'd love thoughts added to #722. I'm hoping we get to spend some time on it soon.

Issue #1016, opened 2018-04-27
Dapper.Contrib does not pass Turkey Test


yusufozturk:
In the following code:

https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.cs#L117

There is a string.equals operation to find "id" in the all properties which fails in Turkey test. So operations are getting failed in all Turkish locale servers. "id" is "İD" or "ıd" in Turkish, so basically it fails for the "ID" column.

We have a custom "ToLower" extension which handles this issue. I don't know if there is another easy solution to work around.

canertosuner:
(+1)
Having the same problem. Any fix about it ?

rhubley:
If you changed it to compare to "ID" instead of "id" does it work for you? Apparently ToUpper works better for Turkey, don't know if it fails for other cultures though.

yusufozturk:
@rhubley 

ToUpper can make it "İD", so it will fail again :)

I think the best way is adding culture to string.equal operation. For now, we wrote a custom convert method to change ıd to id, İD to ID etc.

canertosuner:
@yusufozturk 
Could you share your ext. method ? or what does it exactly do for it ?

Adondriel:
We should probably address this by using whatever locale is used in the model object, by changing the hardcoded "id" to grabbing the variable name from the object itself. This would mean that if they define the model in turkish(is that the name of the language?) that should give us the correct variable to query against.

PR #1023, opened 2018-05-14
Issue #1016

AdemCatamak:
Issue #1016

During the string comparison instead of 'CurrentCultureIgnoreCase', 'OrdinalIgnoreCase' is used. As a result, the case of unit tests in Turkish systems resulting with a failure is resolved.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -114,7 +114,11 @@ namespace Dapper.Contrib.Extensions
 
             if (keyProperties.Count == 0)
             {
-                var idProp = allProperties.Find(p => string.Equals(p.Name, "id", StringComparison.CurrentCultureIgnoreCase));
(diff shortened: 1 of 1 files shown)


rhubley:
How does this handle a Database that is a different culture than the application? 
Could an English application connect to a Turkish Database and still work properly?

Just curious if there are unknown consequences with making this change.

AdemCatamak:
Actually, this error not about Db Collations. This bug occurs when using `string.Equals` with 'CurrentCulture'. 

ScreenShot which is below show us, properties found successfully. However, "Id" is not equal to "id" when ignore case sensitivity in Turkish Culture. Thus, string comparison should be Invariant Culture Case Insensitive.

NickCraver:
This would be a good addition to .Contrib for the next release still, but please see #1658 - Dapper.Contrib is now located at https://github.com/DapperLib/Dapper.Contrib (separate from this repository). 

Part of the split is so that things like this can do in a major/breaking version release of .Contrib, independent of Dapper. So, we need to re-target this to that repo...but the change is sound :)

Issue #953, opened 2018-02-21
Option to set the default ISqlAdapter


oferns:
If I want to use something like MiniProfiler with MySQL and Dapper.Contrib then I need to wrap the MySQlConnection. This causes Dapper.Contrib to use the default (wrong) ISqlAdapter. Please either add an argument to the UpdateAsync & Update methods to accept an ISqlAdapter (like the insert does) or allow us to specify the default adapter

NickCraver:
You can currently control the adapter like this:
```c#
SqlMapperExtensions.GetDatabaseType = conn => "MySqlConnection";
```

I agree this isn't awesome, and want to look at this harder with #722 and related issues in v2, but it does work in current code.

PR #1030, opened 2018-05-29
#953 Optional ISqlAdapter parameter for Insert, Update, Delete, UpdateAsync

klym1:
This PR adds optional `ISqlAdapter` parameter for mentioned extension methods. #953


NickCraver:
I left a solution on #953 - `SqlMapperExtensions.GetDatabaseType` exists for this today (and need not be passed on every method call). The bigger issue though, is this would be a breaking change (any library using these methods would throw a `MethodNotFound` exception). Going to close this out since there's a solution in place :)

PR #1041, opened 2018-06-13
Tweaks to Contrib SqlMapperExtensions

nzcoward:
Updated package dependencies (in tests). Shifted some into .NETCoreApp 1.0 and .NETCoreApp 2.0 conditionals.

Breaking Change: GetTableName returns the name surrounded by Square Brackets, so no need to wrap keyword table names manually (through TableNameMapper or Attribute). Unsure if all supported DBs support this notation, but there was an added [] in the Insert<T> method, which would have broken supported DBs if that was the case.

Updated the tests to not wrap table names where they were. All SQL Server tests pass.

Breaking Change: Changed ordering of GetTableName conditions. I feel that because an attribute is closer to an object, it should be used over other mechanisms (e.g. TableNameMapper). So now conditions evaluate:

1. TableAttribute
2. TableNameMapper
3. TypeName with an 's' (there's actually a good pluralizer somewhere, but perhaps that is better to include as a dependency if the developer uses TableNameMapper).

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -391,7 +391,7 @@ namespace Dapper.Contrib.Extensions
             else
             {
                 //insert list of entities
-                var cmd = $"insert into [{name}] ({sbColumnList}) values ({sbParameterList})";
+                var cmd = $"insert into {name} ({sbColumnList}) values ({sbParameterList})";
                 returnVal = connection.Execute(cmd, entityToInsert, transaction, commandTimeout);
             }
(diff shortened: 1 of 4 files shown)


NickCraver:
Just catching up here - unfortunately this isn't a set of changes we can accept. It would simply break every *other* provider (and the test runs themselves). Dapper is not for SQL Server only, it supports most ADO.NET implementations. 

I'd like to better understand the problem you're trying to solve though - can you please chime in on #722 where we want to re-write all the table and column bits to better support this? I do want to make sure your issue is addressed in v2 where this is planned.

PR #1045, opened 2018-06-18
build.ps1 is not handling pre-release build number correctly.

alexwiese:
The -Contains operator doesn't do substring comparisons.

> -Contains
>Description: Containment operator. Tells whether a collection of reference values includes a single test value.

Should be using -Match operator instead.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -30,7 +30,7 @@ function CalculateVersion() {
         Exit 1
     }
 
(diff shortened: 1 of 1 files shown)


NickCraver:
Good catch - we've replaced the versioning elsewhere and need to here for the next release. Folding this in for the interim :)

PR #1047, opened 2018-06-18
extend Connection.Query<T> from 7 to 15 mapping types

faradzen:
move all Query<TFirst,...,TSeventh, TResult> methods to tt template

### DIFF
--- a/Dapper/Dapper.csproj
+++ b/Dapper/Dapper.csproj
@@ -27,4 +27,16 @@
     <PackageReference Include="System.Runtime.InteropServices" Version="4.3.0" />
     <PackageReference Include="System.Xml.XmlDocument" Version="4.3.0" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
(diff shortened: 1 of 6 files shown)


NickCraver:
Hey! First, thanks for the PR. I appreciate the effort. 

Unfortunately, this isn't a change we'd want to make (it's something that's come up a few times in the past few years). Having template for code generation adds complication, isn't automated in a regular build, and the usefulness of more than 7 overloads doesn't balance the cost of maintaining it. But more directly to the issue: there already exists an overload for the case of more than 7 items (though I'd re-visit overall design at that point):
```c#
public static IEnumerable<TReturn> Query<TReturn>(
	IDbConnection cnn,
	string sql, 
	Type[] types,
	Func<object[], TReturn> map,
	object param = null,
	IDbTransaction transaction = null,
	bool buffered = true,
	string splitOn = "Id",
	int? commandTimeout = null,
	CommandType? commandType = null)
```

faradzen:
yeah, i know that tt is sucks) but we need this on our project and need quickly. thx for response)

PR #1048, opened 2018-06-19
Fix #418 in `UpdateAsync`

### LINKED ISSUE #418: explicitKey bug

alexwiese:
Fixes #418 - `Update` was fixed but `UpdateAsync` was not fixed.


NickCraver:
This area needs a total rewrite in v2, but totally agree on this fix for now. Better to allocate the list than to fail in other ways - we'll optimize it after. Thanks for the fix!

alexwiese:
@NickCraver yeah I did raise that in #1049

PR #1050, opened 2018-06-19
upd comment for query t7 method - need more, use query with Type[]  p…

faradzen:
…arams


NickCraver:
👍 Good comment change, thanks!

Issue #879, opened 2017-11-11
Sysbase dapper.contrib support


giorgiozoppi:
I have a package for supporting sybase in dapper contrib. Currently work in our enviroment, how can I contribute to back to the community?

PR #1053, opened 2018-06-19
Add SQLAnywhere support

xiaoningmzhang:
I also had a solution to support SQL Anywhere (see issue #879).

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -181,7 +181,7 @@ namespace Dapper.Contrib.Extensions
             for (var i = 0; i < allPropertiesExceptKeyAndComputed.Count; i++)
             {
                 var property = allPropertiesExceptKeyAndComputed[i];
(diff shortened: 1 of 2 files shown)


xiaoningmzhang:
My company depends on Sybase, and I would like to get this work pulled in.  Is there any additional contributor guidelines I need to know about or more work that is needed to be done?

mgravell:
for context, I'm not historically very involved in contrib; reviewing this, it looks like this is a breaking change (changes a public interface) which caller's could have implemented themselves and passed in via the optional parameter on `InsertAsync`. Not sure how we can rationalize that.

Note: there's also a *lot* wrong with this API generally, but that isn't your fault - I just looked at it right now, and want to weep gently :(

FWIW, I think what we *should* have used here is an abstract base class API which might have made this more extensible, but: I think we should add this into the "major breaking changes" bucket

NickCraver:
Indeed we can't pull this in as it is a breaking change. We really need to revisit Dapper.Contrib as a whole - maybe this is a sabbatical project?

NickCraver:

PR #1056, opened 2018-06-24
Added "delete" clause

fixer-m:
Wondering why this wasn't added before...


NickCraver:
I think the intent of SqlBuilder was to only be for non-mutation operations. For example, `INSERT` and `UPDATE` also don't exist. I don't have strong opinions on this, but everywhere I've seen this used thus far was with Delete in the template itself to be *very* clear about what's being modified (same with the other verbs).

fixer-m:
@NickCraver I was already thinking of contributing `INSERT` and `UPDATE` operations :)
Ok, then I will use `DELETE` right in the template.

NickCraver:
Closing out to cleanup :)

PR #1062, opened 2018-07-01
Sql columnNames to class property ColumnAttributes

nancykaystudios:
Added the ability to map Sql column names to class property ColumnAttributes. This works by adding the ColumnAttribute to the class property, setting its name the same as the column name in the database.
Example - this will map the database column named "person_id" to the class property named "ContactPersonId":
    [Column("person_id")]
    public int ContactPersonId { get; set; }
This requires references to System.ComponentModel.Annotations - Version="4.5.0" and System.ComponentModel.DataAnnotations.


nancykaystudios:
Not sure my changes are in this request.

PR #1063, opened 2018-07-01
Sql columnNames to class property ColumnAttributes

nancykaystudios:
Added the ability to map Sql column names to class property ColumnAttributes. This works by adding the ColumnAttribute to the class property, setting its name the same as the column name in the database.
Example - this will map the database column named "person_id" to the class property named "ContactPersonId":
    [Column("person_id")]
    public int ContactPersonId { get; set; }
This requires references to System.ComponentModel.Annotations - Version="4.5.0" and System.ComponentModel.DataAnnotations.


nancykaystudios:
The build notes contained this: **No artifacts found matching '.nupkgs\*.nupkg' path**
I'm not sure if I need to commit anything else. I had to install a NuGet package for System.ComponentModel.Annotations v4.5.0, and a reference for this was added to the .csproj files, but do I need to commit any other files for this?

NickCraver:
@nancykaystudios Thanks for taking a stab at this! Unfortunately though, this would be a breaking change. Any users with these properties but not using them that way with Dapper would suddenly have a behavior change.

But! We have other plans for this (that conflict with this specific approach) in v2, please see #722 for discussion.

NickCraver:
Going to close this out as #722 is the path there (we will use the types in this PR, but implementation would be very different).

Issue #718, opened 2017-03-06
Microsoft.Data.Sqlite and Guid column


tidyui:
Hi there! I'm using dapper in my .net core library and have previously tested with SqlServer & LocalDb. I just switched to Sqllite and Guid id columns seem to not be working with the message:

`Unable to cast object of type 'System.Byte[]' to type 'System.Guid'`

See full stack trace below:

    InvalidOperationException: Error parsing column 0 (Id=System.Byte[] - Object)
      Dapper.SqlMapper.ThrowDataException(Exception ex, int index, IDataReader reader, object value)
      Deserialize45afa75f-34ca-4996-b0bd-653743fbf96d(IDataReader )
      Dapper.SqlMapper+<QueryImpl>d__124.MoveNext()
      System.Collections.Generic.List..ctor(IEnumerable<T> collection)
      System.Linq.Enumerable.ToList<TSource>(IEnumerable<TSource> source)
      Dapper.SqlMapper.Query<T>(IDbConnection cnn, string sql, object param, IDbTransaction transaction, bool buffered, Nullable<int> commandTimeout, Nullable<CommandType> commandType)

Regards

erikpowa:
https://github.com/StackExchange/Dapper/issues/642#issue-190093216
https://github.com/aspnet/Microsoft.Data.Sqlite/issues/204#issue-122845487
https://github.com/aspnet/Microsoft.Data.Sqlite/pull/210#issuecomment-166770975

AFAIK It's provider's design, it's pick a fitting type first

mgravell:
this is a more nuanced one than you would think; historically, indeed
dapper has let the downstream provider choose the types that it returns.
Most ado.net providers know how to identify and cast a guid column from
their data store. We are aware that some providers don't do this correctly.
We investigated automatically casting between guid and byte[] in the cases
where it didn't happen, but then another problem came up: endianness. It
turns out that different backend stores use different endianness to store
guids. This means that there is no reliable endianness choice that dapper
can choose or infer that will work correctly in all cases. Because of this
I'm not happy for dapper to provide a probably-wrong conversion.

so: you'll either need to use a provider that understands guids, or use
byte[] and perform your own conversion (checking the results)

tidyui:
[...]

Issue #461, opened 2016-02-15
TypeHandler is not called for parameterised constructor


JoeStead:
We've had to create a TypeHandler to handle DateTimeOffset because SQLite has no concept of date/time, so it is saved as a string, which is fine. When querying the data back out, the `Parse` method is only called if it is being serialised to a type with a parameterless constructor. If a constructor exists which initialises all the properties the type handler is not called, which results in a "InvalidOperationException" with the following 

> A parameterless default constructor or one matching signature (System.String Id, System.String SomeValue, System.String SomeDateValue) is required for DapperTypeHandler.ParameterisedTypeConstructor materialization

Note how it is looking for SomeDateValue to be a string still.

I'm not sure if I'm explaining this the best way, I haven't been using Dapper for very long, so this may not be the best way of solving this problem. Below is a small, reproducible example using an in memory SQLite database.

```csharp
using System;
using System.Data;
using System.Data.SQLite;
using Dapper;

namespace DapperTypeHandler
{
    class Program
    {
        static void Main(string[] args)
        {
            SqlMapper.AddTypeHandler(new DateTimeOffsetHandler());

            var dbConnection = new SQLiteConnection("Data Source=:memory:");
            dbConnection.Open();
            dbConnection.Execute(@"CREATE TABLE SomeTable (
                                      Id                VARCHAR,
                                      SomeValue         VARCHAR,
                                      SomeDateValue     DATETIMEOFFSET
                                    )");

            dbConnection.Execute(
                "INSERT INTO SomeTable (Id, SomeValue, SomeDateValue) VALUES (@Id, @SomeValue, @SomeDateValue)",
                new
                {
                    Id = "id",
                    SomeValue = "what up?",
                    SomeDateValue = new DateTimeOffset(2016, 02, 15, 16, 0, 0, TimeSpan.Zero)
                });

            var parameterlessWorks = dbConnection.Query<ParameterlessTypeConstructor>("SELECT * FROM SomeTable");

            //throws about not being able to find constructor (It expects the DateTime field to be a string still)
            var parameterDoesNot = dbConnection.Query<ParameterisedTypeConstructor>("SELECT * FROM SomeTable"); 

        }
    }

    class DateTimeOffsetHandler : SqlMapper.TypeHandler<DateTimeOffset>
    {
        public override void SetValue(IDbDataParameter parameter, DateTimeOffset value)
        {
            parameter.Value = value.ToString();
        }

[...]

PR #1082, opened 2018-07-25
Handling classes with Guid type properties and persisting them to a sqlite database.

basicdevelopment:
This simple Guid handler should fix issue #447 (and duplicate issue: #718). It implements the method suggested by numerous other users but leaves the byte order (endian) to the .net layer.
It also fixes the guid equivalent issue raised in #461.

Note: this fix has only been tested against .net core 2.1.

### DIFF
--- a/Dapper.Tests.SQlite/ClassWithGuidPropertyWithDefaultConstructor.cs
+++ b/Dapper.Tests.SQlite/ClassWithGuidPropertyWithDefaultConstructor.cs
@@ -1,0 +1,40 @@
+﻿using System;
+
+namespace Dapper.Tests.SQlite
+{
+    /// <summary>
+    /// ClassWithGuidPropertyWithDefaultConstructor: simple class with a Guid property and a default constructor.
+    /// </summary>
+    public class ClassWithGuidPropertyWithDefaultConstructor : IEquatable<ClassWithGuidPropertyWithDefaultConstructor>
+    {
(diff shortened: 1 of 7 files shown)


basicdevelopment:
Not to state the obvious but the error message "Sqlite is unavailable: Unable to load DLL 'sqlite3': The specified module could not be found." that you see in the CI tests (that target net252) can be fixed by copying the appropriate versions (x86 and x64) of the sqlite3.dll to Dapper.Tests\bin\Debug\net452\x86 (and x64) folders.

NickCraver:
First, thanks for the contribution! Unfortunately, there's a bit of history on why this functionality isn't in Dapper...

We have decided not to support this scenario (and are encouraging the ADO.NET team to fix it at the shared layer). In short: Dapper is mostly provide agnostic, with *very* little special casing code. Due to bit order issues, what Dapper gets as a GUID vs. what anyone else gets as a GUID (e.g. Linq2Sql or EntityFramework or 20 others) depends on how you interpret the bits. Nothing is agreed upon here, it's per-provider.

See https://github.com/StackExchange/Dapper/issues/637 for context here...we'e been down this road before and regretted it.

All that being said, let's see if we can possibly address this with type mapping in v2.0 where *you* can choose, rather than us choosing, how to store or query such a thing. That's really the only thing that's going to work for everyone depending on how they stored the data in the first place.

PR #1091, opened 2018-08-08
Make `SqlBuilder.AddClause()` method `public`

jboarman:
In order to extend `SqlBuilder` with additional methods such as `InsertInto()` or `Update()`, one currently would have to derive from `SqlBuilder` and create a less obvious `SqlBuilderEx` type of class for use in a project.

A simpler approach would instead be to create only an extension method that calls `AddClause()`.  This requires that the `AddClause()` method be made public.

This update appears to be super low-risk with the super upside of inviting creative solutions based on `SqlBuilder` to flourish in the wild and eventually make their way back into this shared library.

### DIFF
--- a/Dapper.SqlBuilder/SqlBuilder.cs
+++ b/Dapper.SqlBuilder/SqlBuilder.cs
@@ -102,7 +102,7 @@ namespace Dapper
         public Template AddTemplate(string sql, dynamic parameters = null) =>
             new Template(this, sql, parameters);
 
-        protected SqlBuilder AddClause(string name, string sql, object parameters, string joiner, string prefix = "", string postfix = "", bool isInclusive = false)
+        public SqlBuilder AddClause(string name, string sql, object parameters, string joiner, string prefix = "", string postfix = "", bool isInclusive = false)
         {
             if (!_data.TryGetValue(name, out Clauses clauses))
             {


NickCraver:
There is one heavy downside here: `AddClause()` would be in intellisense for all users, and would be a strange thing indeed compared to all others. Given the purpose of the class and the intent to be intuitive and mapping to SQL, that's a fair big downside IMO.

That being said, I'm open to making it protected so it can be used by any inheritors...

jboarman:
Thanks @NickCraver for hearing me out...


On one hand it's hard for me to see that exposing `AddClause()` via IntelliSense is a problem, but on the other hand, I also can see how you really don't have time to be faced with more questions from end users on how it's supposed to work.

So, I will do my best to lay out a convincing case for why making this method public could actually be a good thing.  :)

*Case # 1 -- Users Already Look to the Source and are aware of `AddClause()`*
There are scant examples or tutorials online for how or why one would use `SqlBuilder`.  There are only a handful of bonafide questions on Stack Overflow that even cover Dapper's `SqlBuilder`.  And at least two of those questions have answers that basically point back to the source code as the one true source of documentation.[[1]] [[2]]  As a result, with `SqlBuilder`'s source code being the primary source of `SqlBuilder` documentation, its users are already savvy of how `SqlBuilder` operates under-the-hood.

*Case # 2 -- Extending `SqlBuilder` is Necessary, So Making that _Easy_ for Users Would be Super Awesome*
If one uses `SqlBuilder` any period, it's not too long before one starts to desire extra methods.  For example, passing a list of strings to `Select()` instead of having to format a comma-separated list for each call, or similarly, passing simply `dynamic` parameters to `Where()` without having to enumerate and format the list of parameters on each call, or to fulfill the need for `InsertInto()` or `Update()` methods -- these are all nice to have and could easy to add with extension methods.

*Case # 3 -- Methods Can Be Hidden from IntelliSense*
Supposedly, according to a couple Stack Overflow answers, editors will hide methods when this attribute is added to a method: `[EditorBrowsable(EditorBrowsableState.Never)]`.[[3]]  Maybe I should have led with this solution, or maybe this just seems hacky.  The upside is that IntelliSense won't hide the method when the method is local to the project where it is defined.  It only hides methods defined in external libaries.  Is this a reasonable compromise?

[1]: https://stackoverflow.com/a/34211696
[2]: https://stackoverflow.com/a/32792233
[3]: https://stackoverflow.com/a/9086345

mdockal:
The method is already protected, I have extended it via inheritance.   However, it would be nice to extend it via an extension method.   I do think that SQLBuilder is lacking features in some areas.    Perhaps if those features were added there would be no need to extend it.

NickCraver:
Apologies, I thought I had responded earlier...but I guess not :(

I have an idea here to hopefully service both needs. What if we make the method internal, and have an extension method (public) that exposes it. But that extension method would be in a `Dapper.Extensions` namespace. So while anyone `using Dapper;` wouldn't see it, extra functionality for extending Dapper would still be available and hopefully suit the needs laid out here.

What do you think? I'm considering this approach for several things...this would just be the first.

NickCraver:
Note that we intend to split Dapper.SqlBuilder into a separate repo potentially (see #1658), but if anyone wants to take a crack at the proposed `Dapper.Extensions` idea above - would gladly accept that PR (or we can modify this one to such an approach).

PR #1094, opened 2018-08-11
Cross target .netcoreapp2.1

iamcarbon:
This allows us to avoid importing the System.Reflection.Emit.Lightweight assembly and it's transitive dependencies on .NETCOREAPP2.1 projects.


NickCraver:
Can you explain a bit more what you're after here? What issue is this actually causing?

I'll go ahead and say: I don't think we want to add another build here - if anything we'll narrow down in the next release eliminating `net451` and `netstandard1.x`. We're hard and heavy on StackExchange.Redis for a bit, then hopefully getting back to some Dapper v2 action here.

iamcarbon:
Was hoping to simplify the dependency graph on our NETCORE2.1 apps -- but all is currently well.

Aside: loading up on targets now might also not be wise with the upcoming .NETCOREAPP2.2, .NETSTANDARD2.1, NETCOREAPP3.0, and .NET4.8 releases in the pipeline.

Issue #1080, opened 2018-07-20
Is Dapper.Rainbow not supporting SQLite?


cnxy:
When I use Dapper.Rainbow for SQLite, the following message appears: System.Data.SQLite.SQLiteException: "SQL logic error No such table: INFORMATION_SCHEMA.TABLES"
Is Dapper.Rainbow not supporting SQLite?
The code show as below:
`            string fileName = "MyDatabase.db";
            if (!File.Exists(fileName)) SQLiteConnection.CreateFile(fileName);
            SQLiteConnection connection = new SQLiteConnection($"Data Source = {fileName}");
            MyDatabase db = MyDatabase.Init(connection, 2000);
            db.Execute("CREATE TABLE IF NOT EXISTS Cat(Id integer PRIMARY KEY AUTOINCREMENT, Name text NOT NULL)");
            db.Cats.Insert(new Cat { Name = "Tom" });`

mdockal:
I have the same issue, so this is a bug.   SQLite doesn't have a table called INFORMATION_SCHEMA.TABLES updated the code to detect if it is a SQLite connection.   Created Pull Request #1095

gerdus:
Ran into this today, not nice.
Rainbow seems to do same job as Dapper.Contrib? Which is also unmaintained.
Please just add a note to the readme stating it only works with SQL Server.

PR #1095, opened 2018-08-12
Fix Issue #1080

mdockal:
Sqlite doesn't have a table called INFORMATION_SCHEMA.TABLES, instead you must use SQLITE_MASTER table.   Updated the code to determine if it IsSqliteConection and append the correct table name.   Also SQLITE_MASTER does not have a column name called "TABLE_NAME".

### DIFF
--- a/Dapper.Rainbow/Database.cs
+++ b/Dapper.Rainbow/Database.cs
@@ -325,9 +325,18 @@ namespace Dapper
                 }
             }
 
-            var builder = new StringBuilder("select 1 from INFORMATION_SCHEMA.TABLES where ");
-            if (!string.IsNullOrEmpty(schemaName)) builder.Append("TABLE_SCHEMA = @schemaName AND ");
-            builder.Append("TABLE_NAME = @name");
+            bool isSqliteConnection = _connection.GetType().Name.ToLower().Equals("sqliteconnection");
+            var builder = new StringBuilder("select 1 from ");
+            if (isSqliteConnection)
+            {
+                builder.Append("SQLITE_MASTER where NAME = @name");
+            }
+            else
+            {
+                builder.Append("INFORMATION_SCHEMA.TABLES where ");
+                if (!string.IsNullOrEmpty(schemaName)) builder.Append("TABLE_SCHEMA = @schemaName AND ");
+                builder.Append("TABLE_NAME = @name");
+            }
 
             return _connection.Query(builder.ToString(), new { schemaName, name }, _transaction).Count() == 1;
         }


NickCraver:
I understand the issue here (agree we need a fix), but historically the detection mechanism hasn't worked out well.

An aside/FYI on the technical side: instead of creating a string with every comparison with `.ToLower()`, instead you can do:
```c#
string.Equals(_connection.GetType().Name, "sqliteconnection", StringComparison.OrdinalIgnoreCase) 
```
...but the main issue is how the detection works in general. For example, you can wrap a `SqliteConnection` in...whatever you want really. MiniProfiler and other libs wrap the connection to track which connections are issues. In the MiniProfiler example, the type of `DbConnection` would be `ProfiledDbConnection`...and we'd still fail.

We have a similar problem in `Dapper.Contrib` with the SQL adapters so we need to find a better solution. Some ideas I've considered:
- Go into a "mode" for a provider based on exceptions thrown, e.g. "SqliteException" - we listen for these and change the type.
  - But that has to be stored somewhere, and what if you change connections (in Contrib) or have multiple types? Not safe.
- `try`/`catch` here, and upon failure (that exception type...we could switch on it), we run the correct one.
  - But that's pretty slow and wasteful

@mgravell any thoughts here?

mdockal:
@NickCraver  I was not aware of the String.Equals() Method Overload.

I believe something like this would work because I am using it in another project I am working on:
```
         
                var connection = new DerivedSqliteConnection(@"Data Source=test.db;Cache=Shared"); 
                var sqlFlavor = SqlFlavor.For(connection);  //Returns SqliteFlavor
                var builder = new StringBuilder("select 1 from ");              
                builder.Append(sqlFlavor.SchemaTableNameString);                
                if (!string.IsNullOrEmpty(schemaName)) builder.Append("TABLE_SCHEMA = @schemaName AND ");
                builder.Append("TABLE_NAME = @name");
```

The problem is you need to look at the Type.BaseType and see if it inherits from one of the Framework connections.

```
 public interface ISqlFlavor
    {
        string SchemaTableNameString { get; }
        string SchemaTableColumnNameString { get; }
    }

    public abstract class SqlFlavor : ISqlFlavor
    {

        private static readonly Dictionary<string, ISqlFlavor> LanguageDictionary = new Dictionary<string, ISqlFlavor>
        {
            ["sqlconnection"] = new SqlServerFlavor(),
            ["sqlceconnection"] = new SqlServerCeFlavor(),
            ["npgsqlconnection"] = new PostgreSqlFlavor(),
            ["sqliteconnection"] = new SqliteFlavor(),
            ["mysqlconnection"] = new MySqlFlavor(),
        };

        public virtual string SchemaTableNameString => "INFORMATION_SCHEMA.TABLES where ";

        public virtual string SchemaTableColumnNameString => "TABLE_NAME = @name";


        public static ISqlFlavor For(string name, bool throwExceptionIfNotFound = false)
        {
            if (LanguageDictionary.ContainsKey(name)) return LanguageDictionary[name];
            if (throwExceptionIfNotFound)
            {
                throw new InvalidOperationException(string.Format("The language {0} does not exist.", name));
            }
            return null;           
        }


        public static ISqlFlavor For(IDbConnection connection)
        {           
            var language = default(ISqlFlavor);
            var connectionType = connection.GetType();
            // Traverse connection type and base types to find a connection.
            while (connectionType != null)
            {
                language = For(connectionType.Name.ToLower());
                if (language != null) break;
                connectionType = connectionType.BaseType;
            }
            return language;           
        }
        
    }

    public class SqliteFlavor : SqlFlavor
    {
        public override string SchemaTableNameString => "SQLITE_MASTER where ";

        public override string SchemaTableColumnNameString => "NAME = @name";
    }

    public class SqlServerFlavor : SqlFlavor
    { }

    public class PostgreSqlFlavor : SqlFlavor
    { }

    public class MySqlFlavor : SqlFlavor
    { }

    public class SqlServerCeFlavor : SqlFlavor
    { }
```

PR #1102, opened 2018-08-22
Force static constructor calls for DefaultTypeMap _fields types

MathiasPius:
Static constructors are not guaranteed to be called when instantiated using Reflection. I encountered this bug when attempting to register TypeHandlers in the static constructor of a generic function, and then having DataExceptions thrown by Dapper when attempting to query a database for objects containing fields of the generic type.

The following code example demonstrates the issue: https://gist.github.com/MathiasPius/cd6e7d3b10aca8d0212e8b9b3435d852#file-program-cs

When run, it produces the following exception: 
```
InvalidCastException: Unable to cast object of type 'System.Guid' to type 'ID`1[DapperStaticConstructorBug.Program+ComplexObject]'.
```
Because the static constructor of `ID<ComplexObject>` is never called, and the TypeHandler for `ID<ComplexObject>` is never added.

This pull request modifies only the `DefaultTypeMap` and makes it explicitly call the static constructor of any _field types, if they exist. The `RunClassConstructor` function also guarantees that the static constructor is only ever called once, and will not re-run static constructors, if they have already been run.

Not too familiar with the Dapper codebase, so let me know if there's a better way to go about this.

### DIFF
--- a/Dapper/DefaultTypeMap.cs
+++ b/Dapper/DefaultTypeMap.cs
@@ -23,6 +23,12 @@ namespace Dapper
                 throw new ArgumentNullException(nameof(type));
 
             _fields = GetSettableFields(type);
+			
+			// when using reflection, there's no guarantee that static constructors are called
+			// before object instantiation, so do it manually for all types (whose constructors haven't been called)
+			foreach(var field in _fields)
+                System.Runtime.CompilerServices.RuntimeHelpers.RunClassConstructor(field.FieldType.TypeHandle);
+			
             Properties = GetSettableProps(type);
             _type = type;
         }


NickCraver:
Ultimately, I don't think we should make this change. The situation is unfortunate (but still resolvable on the user side), the workaround may be worse though.

The reason this isn't a safe change is it starts invoking code all over someone's codebase which may not be so friendly. Things which work today because dependencies are instantiated or statically configured in the right order (as an example, let's say a connection string is set before it's used)...would suddenly go haywire. And users wouldn't understand why. And *if they did* there's nothing they could do about it in most cases, save downgrading Dapper.

I am happy to discuss further, and I'm glad you raised the issue for us to be on the lookout for and others to find. On the fix side though: this isn't something we should add into the library, I hope that's understandable. I don't like turning down PRs (especially first PRs!), but in this case that's the situation.

MathiasPius:
Very sound reasoning, I agree completely. I was able to work around the issue by overriding the SqlMapper.TypeMapProvider and calling the static constructor on all the properties of a a new DefaultTypeMap instance:

```C#
SqlMapper.TypeMapProvider = (t) =>
{
    var map = new DefaultTypeMap(t);
    foreach (var property in map.Properties)
        System.Runtime.CompilerServices.RuntimeHelpers.RunClassConstructor(property.PropertyType.TypeHandle);

    return map;
};
```

Thank you for taking the time to explain your reasoning.

PR #1105, opened 2018-08-25
Support splitting using the type name as a prefix alias

dallasbeek:
Allows using the type name as an alias when returning records that may have columns that exist in both columns or database doesn't allow selecting a column name twice.

SELECT A.*, B.Id AS TableBId, B.Name AS TableBName
FROM
Table A
JOIN TableB B ON.....

Split ON "Id" (or split on TableBId)

### DIFF
--- a/Dapper.Tests/MultiMapTests.cs
+++ b/Dapper.Tests/MultiMapTests.cs
@@ -597,7 +597,7 @@ Order by p.Id
                      cast('01 Feb 2013' as datetime) as CreateDate,
                      'ghi' as Name, 'def' as Phone",
             (T, P) => { T.Author = P; return T; },
-            null, null, true, "ID,Name").Single();
+            null, null, true, "ID,ID").Single();
 
             Assert.Equal(123, result.ID);
             Assert.Equal("abc", result.Title);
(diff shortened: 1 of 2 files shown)


NickCraver:
Hey, thanks for the commit/interest!

Unfortunately, this would be a breaking change in behavior - so not something we could take. More so though, a user can just specify this in options already bu using the string, so I'm not sure changing the behavior here (or making it more magical) really makes sense. Generally, when something matches *only one* naming scheme, it's better to be very explicit about that thing.

dallasbeek:
This pull is pretty specific to Sqlite https://github.com/mapbox/node-sqlite3/issues/443 not returning duplicate columns.  This pull doesn't "change" behavior but I'm indifferent as to whether this is accepted as I'm not using SQLite and I'm sure those that do probably just alias the column and handle it via their own mapper.

NickCraver:
Unfortunately it's not scoped to that provider (and honestly it's be slightly worse if it was...making it inconsistent). If we need to change behavior here, I'm open to finding an approach (like we are in #722 with `[Table]`, `[Column]`, etc.) but we need to be less magical and backwards compatible in whatever approach that is.

dallasbeek:
Funny you mention that because that is how this came to light for me https://github.com/StackExchange/Dapper/issues/722#issuecomment-410855632.  [Dapper.Database](https://github.com/dallasbeek/Dapper.Database)

dallasbeek:
Trully it's not terribly important to me but I do recall PetaPoco seemed a bit more intelligent in it's ability to split results.

Issue #1115, opened 2018-09-05
sporadic "Field already exists"


alex-jitbit:
When adding fields to a dynamic result-set (which is `DapperTable` with `DapperRows`) inside a `foreach` loop - a sporadic "InvalidOperationException: Field already exists" is thrown.

Like this.

```
var results = connection.Query("SELECT blablah FROM blahblala");
foreach(var o in results)
     o.NewProperty = xxx;
```

Exception:

```
[InvalidOperationException: Field already exists: NewProperty ]
Dapper.DapperTable.AddField(String name) +179
Dapper.DapperRow.SetValue(String key, Object value, Boolean isAdd) +52
Dapper.DapperRow.SetValue(String key, Object value) +11
```

Getting this exception like once a month in a heavily used .NET app, right after deploying (i.e. when the views are recompiled). I guess the compiler/runtime/whatever parallelizes the foreach in some way?... 

Anyway, the `SetValue` method in `DataRow` in not thread safe. Adding a `lock` fixes this nicely.

I will suggest a PR that fixes this shortly.

NickCraver:
I'm hesitant to make changes here since the rest of the type isn't thread safe either (nor does it purport to be). I'm very curious how this is happening though (it's often proved to be a bad idea to fix a thing we don't understand).

For your case: are there no statics in play, or concurrency above? Dapper and the runtime would not parallelize on the same object here...

For what it's worth, generally we try to avoid the dynamic `.Query()` overloads and use concrete types. It also works for value tuples as an easy way to get this going without creating types only used once, for example `.Query<(string NewProperty)>("Select NewProperty From MyTable");` (note this maps by *position* and not name).

alex-jitbit:
Well, the one truly reliable way to reproduce this error would be:

```
var results = connection.Query("SELECT blablah FROM blahblala");
Parallel.ForEach(results, r => {
    r.NewProperty = xxx;
});
```

I'm also very curious how this is happening without `Parallel` :) and spent a good part of the week trying to reliably replicate this... The only way I found involves some black magic, like "restart Visual Studio two times then rerun a Selenium-powered unit-test that launches IIS-Express" :)) 

I have no statics in my code, it's just a plain action-method in an MVC controller in NET 4.5 app.

Anyways, feel free to close the issue, as this can be easily fixed via a `lock` outside your code. Lets just keep the issue in the archive, so people can google it up (I saw some searchbox suggestions on this, meaning other people DO search for this stuff)

[...]

PR #1116, opened 2018-09-05
proposed fix for #1115

alex-jitbit:
making SetValue thread safe


NickCraver:
Discussing on #1115 - this is the first time this has happened so I'm remiss to change behavior (on introduce locks for all) that should likely be upstream.

PR #1123, opened 2018-09-10
Benchmarks:  updating libs and overall runner

NickCraver:
This is prepping a run we can drop into the README easily. And more info on allocations and such along with it. It's finishing off the benchmark port getting it into a good output state (something we can copy/paste). It removes Simple.Data because the package was debug and causes a lot of issues.

This doesn't output 100% like we want yet (fastest to slowest). BenchmarkDotNet joins summary results after ordering...we need to post-order that. I'm asking the maintainers how to do so.

Also note Belgrade was not parameterizing before, giving it a huge unfair advantage. This has been fixed to be apples:apples.


Dapper.Tests.Performance/Benchmarks.cs
```diff
@@ -44,15 +43,20 @@ public Config()
...
-                .WithTargetCount(5)
-                .WithRemoveOutliers(true)
+                .WithInvocationCount(50)
+                .WithIterationCount(BenchmarkBase.Iterations)
+                .WithMaxIterationCount(BenchmarkBase.Iterations)
+                .WithOutlierMode(OutlierMode.All)
             );
         }
     }
-}
```
adamsitnik: This should make it work:

```
Add(new DefaultOrderer(SummaryOrderPolicy.FastestToSlowest));
SummaryPerType = false;
```

NickCraver: I don't see an `Add()` overload for `IOrderer`, am I missing a namespace somewhere? Had it working at the `.Run()` call site over in Program.cs, but happy to move here if it works better.

adamsitnik: @NickCraver Sorry, it should be `Set`

PR #1126, opened 2018-09-11
Added Dashing Benchmarks

markjerz:
Seeing as you're in the middle of updating the benchmarks :-)


NickCraver:
Pulled in, thank you!

PR #1132, opened 2018-09-20
readonly struct

SergerGood:
added for comparison

### DIFF
--- a/Dapper.Rainbow/Dapper.Rainbow.csproj
+++ b/Dapper.Rainbow/Dapper.Rainbow.csproj
@@ -6,8 +6,6 @@
     <Description>Trivial micro-orm implemented on Dapper, provides with CRUD helpers.</Description>
     <Authors>Sam Saffron</Authors>
     <Copyright>2017 Sam Saffron</Copyright>
-    <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
-    <LangVersion>latest</LangVersion>
     <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
     <!-- TODO: Docs -->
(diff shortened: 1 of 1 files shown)


SergerGood:
rollback .net franework target

NickCraver:
Looking much better! I need to see which version of C# this actually needs (latest has issues with build servers and such) and use that version, e.g. 7.3 is safe. In any case, we should locate that in Directory.Build.props instead of each project. properties in there apply across all projects :)

Pinging @mgravell for breaking change eyes - I'm not 100% on the rules here.

mgravell:
The change to add `in` on the `public` methods is absolutely a breaking change; fine for the non-public, but no-go for the public, until the next time we're doing a hard break.

Context: https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxASwDYB8ACBmAAgTggBMB7AOwwE8DYEBXAYxgIDFzyCBvIkitTppKbYBAQBuAgF8AsACh8BHACYCAYV6KCulYRwAWAgEEoANQiVMGCAApO3AGZcAlLxkEA9F4IA3CAxGOBgaAAc4Di4CF3IdPWUjUygASUo7ESjnNw9vXwBtEQBdf0Dg0IisgDIYrnjdROMzACU4JztiJyza8nceTx9SoJDwyMca2MUZIA

mgravell:
bits of this look fine, but to emphasize we can't take things like the change to add `in` on `public` methods like:

``` c#
public static int Execute(this IDbConnection cnn, in CommandDefinition command) => ExecuteImpl(cnn, command);
```

Do you want to revisit? or should I pick and choose pieces?

SergerGood:
done, but there are problems with WIP

mgravell:
Assigning to self; will re-review and get back

PR #1141, opened 2018-10-17
[ADD] Dapper.Contrib: Insert, Update, Delete Non-generic with param o…

SvetiStephan:
…bject entity

Adding non-generic functions to Dapper.Contrib: Insert, Update, Delete, because some might need the runtime type of an instance (.GetType()) and not typeof(T).


NickCraver:

Issue #569, opened 2016-07-22
Dapper WHERE IN statement fails with ODBC


mgravell:
See http://stackoverflow.com/questions/38527197/dapper-where-in-statement-with-odbc

mgravell:
Fixed in 1.50.2

PR #1146, opened 2018-10-19
Fixed issue #569 with IN and other condition.

andresrsanchez:
Hello,

On issue #569 with IN statement and other condition Query() always fails because 'Replace' in 'PassByPosition' method (SqlMapper.cs) doesn't iterate over '?' character, and Clear() statement in 'firstMatch' condition delete all IN parameters. The result is 'wrong number of parameters' logically.

I created an issue a couple months ago #1025 but without any test, so i closed it.
Thanks for all and sorry for my english.

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -2101,11 +2101,11 @@ namespace Dapper
                             else
                             {
                                 var sb = GetStringBuilder().Append('(').Append(variableName);
-                                if (!byPosition) sb.Append(1);
+                                if (!byPosition) sb.Append(1); else sb.Append(namePrefix).Append(1).Append(variableName);
                                 for (int i = 2; i <= count; i++)
                                 {
                                     sb.Append(',').Append(variableName);
-                                    if (!byPosition) sb.Append(i);
+                                    if (!byPosition) sb.Append(i); else sb.Append(namePrefix).Append(i).Append(variableName);
                                 }
                                 return sb.Append(')').__ToStringRecycle();
                             }
--- a/Dapper.Tests/Providers/OLDEBTests.cs
+++ b/Dapper.Tests/Providers/OLDEBTests.cs
@@ -82,6 +82,22 @@ namespace Dapper.Tests
         }
 
         [Fact]
+        public void Issue569_SO38527197_PseudoPositionalParameters_In_And_Other_Condition()
+        {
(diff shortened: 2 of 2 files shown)


Dapper/SqlMapper.cs
```diff
@@ -2101,11 +2101,11 @@ public static void PackListParameters(IDbCommand command, string namePrefix, obj
                             else
                             {
                                 var sb = GetStringBuilder().Append('(').Append(variableName);
-                                if (!byPosition) sb.Append(1);
+                                if (!byPosition) sb.Append(1); else sb.Append(namePrefix).Append(1).Append(variableName);
```
mgravell: a little confused here - it looks like we'll have `{variableName}{namePrefix}1{variableName}` because of the previous line...? (sorry for delay)

andresrsanchez: Oh yeah, i'll try to explain myself because i'm confused too haha

With the test I added I think that the "problem" is easily to view, in concrete is in this method https://github.com/StackExchange/Dapper/blob/master/Dapper/SqlMapper.cs#L1730 and this line 
```csharp
cmd.CommandText = pseudoPositional.Replace(cmd.CommandText, match =>
```

If you debug the above line you will see that cmd.CommandText contains 

```sql
where s1.value in (?,?,?,?) and s2.value = ?score?
```
And the problem with that is that Replace doesn't iterate around '?' character, so at the end of the method with or without my commit cmd.CommandText is gonna be 

```sql
select s1.value as id, s2.value as score 
                        from string_split('1,2,3,4,5',',') s1, string_split('1,2,3,4,5',',') s2
                        where s1.value in (?,?,?,?) and s2.value = ?
```
but without my commit cmd.Parameters has one parameter, not five because cmd.CommandText with my commit looks like the above image and Replace does its job!

```sql
where s1.value in (?ids1?,?ids2?,?ids3?,?ids4?) and s2.value = ?score?
```

Thanks for all Marc!!

mgravell:
OK, I think that makes sense, thanks; merging

akshaykhalatkar:
Hi @mgravell  @rhubley
I am currently using dapper 1.60.6 version and facing similar issue while querying the data using similar regex provided in this example. 
I think that this change will help me as well. So should i upgrade my dapper to the latest 2.0.4 one. I saw that this fix is tagged in 1.50 release, so not sure if we need to upgrade our package or not.

PR #1149, opened 2018-10-27
Mention how to use user defined variables

### LINKED ISSUE #620: Documentation should mention how to use user defined variables
Because it DOES work! :-)

I finally found this link, which helped me: http://stackoverflow.com/questions/9186983/non-parameter-sql-variables-with-mysql-connector-net-and-dapper

Conclusion: enabled it on the connectionstring and don't provide Dapper with a property to map.

yvesmh:
Fixes #620

Added a section to indicate Dapper works well with User Defined Variables.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -370,6 +370,14 @@ using (var reader = connection.ExecuteReader("select * from Shapes"))
 }
 ```
 
+User Defined Variables
+---------------------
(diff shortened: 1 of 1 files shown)


mgravell:
I'm OK with this, but could you please add "with mysql" (with whatever
casing) to the title? This seems specific to one provider.

yvesmh:
Sure, added it now.

PR #1154, opened 2018-11-06
Improve Exception message

Wixely:
In cases where the object has a property named ID, the property will automatically be assigned as [Key], even if [Computed] is added manually in an attempt to override it. This can cause confusion, this exception will be thrown despite the code looking fine and there only being one [ExplicitKey] defined. I've added this extra line of information to help diagnosis.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.cs
@@ -153,8 +153,7 @@ namespace Dapper.Contrib.Extensions
             var explicitKeys = ExplicitKeyPropertiesCache(type);
             var keyCount = keys.Count + explicitKeys.Count;
             if (keyCount > 1)
-                throw new DataException($"{method}<T> only supports an entity with a single [Key] or [ExplicitKey] property."+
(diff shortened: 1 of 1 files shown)


Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -153,7 +153,8 @@ private static PropertyInfo GetSingleKey<T>(string method)
             var explicitKeys = ExplicitKeyPropertiesCache(type);
             var keyCount = keys.Count + explicitKeys.Count;
             if (keyCount > 1)
-                throw new DataException($"{method}<T> only supports an entity with a single [Key] or [ExplicitKey] property");
+                throw new DataException($"{method}<T> only supports an entity with a single [Key] or [ExplicitKey] property."+
+                " [Key] Count: ${keys.Count}, [ExplicitKey] Count: ${explicitKeys.Count}");
```
NickCraver: This isn't actually parameterized in C# - what you want here is:
```c#
throw new DataException($"{method}<T> only supports an entity with a single [Key] or [ExplicitKey] property. [Key] Count: {keys.Count}, [ExplicitKey] Count: {explicitKeys.Count}");
```

PR #1166, opened 2018-12-07
support aliyun mysql rds read/write connection LAST_INSERT_ID()

witskeeper:
support aliyun mysql rds read/write connection LAST_INSERT_ID()

### DIFF
--- a/Dapper.Contrib/Dapper.Contrib.csproj
+++ b/Dapper.Contrib/Dapper.Contrib.csproj
@@ -8,6 +8,9 @@
     <TargetFrameworks>net451;netstandard1.3;netstandard2.0</TargetFrameworks>
     <!-- TODO: Docs -->
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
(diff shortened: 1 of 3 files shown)


bgrainger:
MySQL is "spoken" by a lot of different backends (MySQL, MariaDB, Amazon Aurora, Azure Database for MySQL, TiDB, etc.) so it's difficult to say for sure that this is a completely safe change without testing them all. However, it seems like it should be safe; that's valid [MySQL comment syntax](https://dev.mysql.com/doc/refman/8.0/en/comments.html) and should be ignored by any other MySQL server.

It might raise questions if users are looking at MiniProfiler or server query logs or similar, and wondering where this new comment suddenly came from.

NickCraver:
I'm going to close this out because I just realized, this PR seems to be intended for a fork at: https://www.nuget.org/packages/NetCorePal.Dapper.Contrib/

witskeeper:
> I'm going to close this out because I just realized, this PR seems to be intended for a fork at: https://www.nuget.org/packages/NetCorePal.Dapper.Contrib/

that's me，haha.

PR #1175, opened 2019-01-09
Hana provider

Mosheh:
The SAP Company  does not provide lib through the nuget repository, I needed to include it as a folder in the project.

Download Hana Database (VM Oracle Virtual Box or VMWare): https://www.sap.com/cmp/ft/crm-xu16-dat-hddedft/index.html


NickCraver:
I'm going to close this out as-is, since it'll add more binaries to the repo here. If there's a NuGet release people can use I'm happy to take another look.

Issue #798, opened 2017-06-08
Get type which was added with AddTypeMap


CassandraDied:
I can add a custom type mapping with AddTypeMap method, but how can I get it back if I want to check that a mapping is really there?

It seems like GetTypeMap is the right thing to call, but it works with a different collection.

1. How can I check that my mapping was added?

2. Is it correct that GetTypeMap and AddTypeMap  are not working with the same collection?

3. As I can see, SetTypeMap will add my custom mapping and I can retrieve it with GetTypeMap later. Is it the correct way to accomplish my goal? I want to map C# DateTime type to DateTime2 SQL Server type.

namphuongtran:
Hi @CassandraDied , 
Did you resolve DateTime to DateTime2 SQL server issue? i got same issue when mapping DateTime parameter from c# to store which is usage User Defined Type Table.  

I dont know why when i pass parameter is type of DateTime then i got message "System.Data.SqlClient.SqlException: 'Conversion failed when converting date and/or time from character string.
". However i use call insert or update with datetime type it working fine.
Anybody help me? 
THanks

justintoth:
@namphuongtran Did you figure it out? I have the same issue.

tangdf:
Hi @CassandraDied 
You can call  `LookupDbType` method.
```cs
            SqlMapper.AddTypeMap(typeof(DateTime), DbType.DateTime2);

#pragma warning disable 618
            var dbType = SqlMapper.LookupDbType(typeof(DateTime), null, false, out var handler);
#pragma warning restore 618


            Assert.Equal(DbType.DateTime2,dbType );
```
But the method  is abandoned, only use  for internal .
https://github.com/StackExchange/Dapper/blob/a213b2aeee66ac0bff327ff19e70516fb629abe3/Dapper/SqlMapper.cs#L371-L419

jamesmh:
Same here -> I have cases where I need to map certain types temporarily, so I need to get the current mapped `DbType`, set a new one, run some queries, then put the original one back.

Using an internal "obsolete" method doesn't jive well with me... Would be nice to have a non-obsolete way to do this.

PR #1188, opened 2019-02-06
Fix Issue 798: Public Way To Get Currently Mapped DbType

jamesmh:
Potential fix for #798.

This simply exposes one of the internal "obsolete" methods via a wrapper method.

This would allow someone to (for example) get the currently mapped DbType for some type, Add a new MapType and run some queries. Then, "reset" the map back to the original in a more dynamic way (without having to know what the original mapped type was).

### DIFF
--- a/Dapper.Tests/TypeHandlerTests.cs
+++ b/Dapper.Tests/TypeHandlerTests.cs
@@ -736,5 +736,29 @@ namespace Dapper.Tests
             public string SomeValue { get; }
             public Blarg SomeBlargValue { get; }
         }
+
+        [Fact]
+        public void Issue798_TestChangingDefaulTypeMappingUsingDbTypeLookup()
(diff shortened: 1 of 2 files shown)


NickCraver:
Thanks a lot for information on the use case here. Agreed it's an issue, but for the approach we'd need something safer (this is static and introduces a race for the case).

Overall, we're looking at an `IDapperConfig` ambient and passable context to a method - that would allow type map changing and fetching (we can be more permissive about access to it, since you'll be in control of the scope and lifetime of the config being used). I'll try and open an issue in the coming weeks with the ideas we've been spitballing in hangouts to hopefully address this and may more issues.

luca-domenichini:
@NickCraver any plan on this? Did you create the issue about `IDapperConfig`?

PR #1198, opened 2019-02-24
Bump version to 1.60.0

NickCraver:
This bumps and locks at 1.60.0 to cope with whatever binding policies .NET Core is enforcing - then we can stick at this until 2.x I guess.


NickCraver:
Reviewed - pushing this up

PR #1199, opened 2019-02-25
Gets timeout from connection

ismaelgasparin:
Condition to obtain connection timeout, when it is not informed.

Reason: I have many connection strings, with the timeouts defined in the string, which change depending on the environment

### DIFF
--- a/Dapper/CommandDefinition.cs
+++ b/Dapper/CommandDefinition.cs
@@ -123,6 +123,10 @@ namespace Dapper
             {
                 cmd.CommandTimeout = SqlMapper.Settings.CommandTimeout.Value;
             }
+            else 
+            {
+                cmd.CommandTimeout = cnn.CommandTimeout;
+            }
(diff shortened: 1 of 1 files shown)


NickCraver:
@ismaelgasparin First, thanks for the first PR :)

On the bad news side: these values aren't (and don't mean) the same thing, though - they should not be intertwined. One is "how long does it take to establish a connection to a server?", the other is "how long should this command take, on an established connection?" I'm not sure how to best phrase: this just isn't a valid thing to do.

For example, in a local LAN I'd often timeour connections to my server at 1 second, because it should be nearly instantaneous. But queries could take any amount of time. On SQL Server for example, this would lower all queries from a 30 second threshold to a 1 second threshold and break a lot of people :)

ismaelgasparin:
Tks @NickCraver .

Oh, I did not know that. Thanks for the reply :). Now I know they are different things.

Best regards

PR #1203, opened 2019-03-07
[Dapper Contrib] Enable optimistic concurrency using Row Versioning

### LINKED ISSUE #851: Handling RowVersion for optimistic concurrency during update
Hi,

The database we are operating against uses optimistic concurrency via the MS SQL `TIMESTAMP` column type. This derails using `Update` without some complete faff of custom SQL to handle the situation.

I have forked the repo and done a first-pass implementation of supporting a new `RowVersion` attribute here: https://github.com/RaringCoder/Dapper

It is bundled in with some aesthetic changes, an update to how property caching works, and some code de-duplication.

Before I go for a full PR, I want some advice on how better to handle a concurrency collision other than simply not updating any rows (the current implementation). My prevailing theory is something akin to how `Insert` currently works in that, per adaptor, we do a query to select any database-generated Ids. We _could_ do something similar and select a flag out when checking if the version changed:

    SELECT COUNT(1) AS VersionHasChanged FROM Table WHERE Key = @key AND Version <> @version

I will be using my fork locally to make progress on the app I'm working on, but I'd like to get this baked in to Contrib properly.

nehmebilal:
I added a [RowVersion] attribute that can be used on an integer column for versioning rows and to enable support for optimistic concurrency. Updating a row will only succeed if the version in the given entity matches the version of the row in the database.

The resulting query is as follows:

```
update myTable set  Version = Version + 1, `Foo` = @Foo, `Bar` = @Bar  where `Id` = @Id and `Version` = @Version
```

I tested it in my application and it works well. I will add the appropriate unit tests if you're willing to merge this feature.

Resolves #851.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -208,63 +208,13 @@ namespace Dapper.Contrib.Extensions
         /// <returns>true if updated, false if not found or not modified (tracked entities)</returns>
         public static async Task<bool> UpdateAsync<T>(this IDbConnection connection, T entityToUpdate, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
         {
-            if ((entityToUpdate is IProxy proxy) && !proxy.IsDirty)
+            if (entityToUpdate is IProxy proxy && !proxy.IsDirty)
             {
                 return false;
             }
 
-            var type = typeof(T);
-
-            if (type.IsArray)
-            {
-                type = type.GetElementType();
-            }
-            else if (type.IsGenericType())
-            {
-                var typeInfo = type.GetTypeInfo();
-                bool implementsGenericIEnumerableOrIsGenericIEnumerable =
-                    typeInfo.ImplementedInterfaces.Any(ti => ti.IsGenericType() && ti.GetGenericTypeDefinition() == typeof(IEnumerable<>)) ||
-                    typeInfo.GetGenericTypeDefinition() == typeof(IEnumerable<>);
-
-                if (implementsGenericIEnumerableOrIsGenericIEnumerable)
-                {
-                    type = type.GetGenericArguments()[0];
-                }
-            }
-
-            var keyProperties = KeyPropertiesCache(type).ToList();
-            var explicitKeyProperties = ExplicitKeyPropertiesCache(type);
-            if (keyProperties.Count == 0 && explicitKeyProperties.Count == 0)
-                throw new ArgumentException("Entity must have at least one [Key] or [ExplicitKey] property");
-
-            var name = GetTableName(type);
-
-            var sb = new StringBuilder();
-            sb.AppendFormat("update {0} set ", name);
-
-            var allProperties = TypePropertiesCache(type);
-            keyProperties.AddRange(explicitKeyProperties);
-            var computedProperties = ComputedPropertiesCache(type);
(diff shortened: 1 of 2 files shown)


nehmebilal:
Actually the Etag row doesn't necessary need to be automatically updated by the database and it's value can be set by the caller. To make this work, I can add a Boolean parameter to the Etag attribute specifying whether the row is auto-updated by the database or not.

nehmebilal:
Hello?

NickCraver:
Just getting to this - there are a lot of things to consider when doing optimistic concurrency. What is a version for this column? Some people use a GUID, some people use a date time, or a shared sequence, or a SQL Server `ROWVERSION`, etc. The implementation here assumes a simple integer or long implementation which only covers one of many use cases. That's a design that isn't well suited for a core feature of a shared library.

I think the goal here would be in v2 would be to perhaps make the SQL generation in the few .Contrib methods a bit pluggable so that users can override a bit and handle additional columns like this if we wanted or some such. This would allow a big of flexibility in how this operates (e.g. not just setting a column but referencing a sequence, etc.) Does that make sense?

nehmebilal:
@NickCraver I would agree with you if this was being added to the main Dapper API as opposed to Dapper Contrib. To me Dapper Contrib provide simple common APIs that are easy to use out of the box but are not very customizable. For instance, insert does not support a where clause, update only works if you provide the full row, only int ids are allowed, etc. With that in mind, I think a default optimistic concurrency API would align well with what Dapper Contrib currently provides, i.e. out of the box easy to use apis. Not to mention that several databases like Azure Table for instance have chosen one mechanism to support optimistic concurrency (in the case of Azure table they use Etags), probably to make it an easy to use feature.

I agree that if Dapper Contrib APIs were more customizable maybe there wouldn't be a need for this, but they currently aren't and I feel that having a plan to maybe (or *perhaps* to use your wording :)) make them pluggable in the future is not too concrete to stop us from adding an additional features to what we currently have, that would probably be useful to most users.

NickCraver:
@nehmebilal We're planning out Dapper v2 now, all the things we want to do later in the year. Adding customization to those methods is intended...so adding an API here I'd have 100% intention of removing with no in-box replacement in v2.0 isn't great for users. All APIs have weight, and this would be dropped off in the move - it's just not a thing we aim to support. Either we would support it in a flexible way that worked for many people...or we wouldn't ship it in box. The goal is not to ship this in v2, it's to enable *someone else to plug it in* in v2.

The arguments here of a single inflexible implementation do align with the status quo of Dapper.Contrib...however, that status quo is bad (and there are many issues and PRs showing this, e.g. all of the column issues and limitations). Part of making this immensely better in v2 is to not make it worse and even harder to port (and further break people) in for v2.

nehmebilal:
@NickCraver When do you think V2 would be shipped? :)

NickCraver:
The goal is this year - we have a long list and we'll see how it goes. We're talking a lot about what changes need to go in for vNext here while away from work at an event with some time to do so.

nehmebilal:
Ok. I was hoping we don't have to fork the repo to unblock ourselves, but I see no other choices at the moment.

rubenruvalcabac:
Hi,

@NickCraver As v2 has been released, I would like to know if the changes you mentioned were implemented and how could use it to enable the optimistic concurrency checking.

@nehmebilal the [RowVersion] attribute makes sense as a very straightforward way to solve this important feature. Did you publish your solution?

PR #1204, opened 2019-03-07
Extend support for Proxy interfaces - move ProxyGenerator Code to Dapper

OneLineTwoBugs:
Summary:
- Support Change Tracking per field (Dapper.Contrib had Change Tracking per object) if type is an interface
- Dapper's `Query*<T>` returns a dynamic proxy object if `typeof(T).IsInterface()`. All properties that have been set by `Query*<T>` are marked as dirty in the result.
- Dapper.Contrib's `Insert<T>` and `Update<T>` functions only add dirty fields to the sql statement.
- This allows for "Partial Updates" and "Partial Inserts" as described here:  https://stackoverflow.com/questions/54978664/dapper-dapper-contrib-partial-updates

Impacts: 
- The code is fully backward compatible, all existing test-cases pass without any modification to them.
- new testcases have been added to test the new possible usecases of Partial Updates and Partial Inserts.
- Added Overhead for existing code is minimal. Dapper's `Query*<T>` functions execute one additional `if (type.IsInterface())` per call.
- The class `ProxyGenerator` is moved from Dapper.Contrib to Dapper.
- The Interface IProxy from Dapper.Contrib (now in Dapper) is extended by one additional property `HashSet<string> DirtyFields` and one method `void MarkAsClean()`.
- The Dirty Fields are exposed to the client. The returned dynamic proxy objects implement `ProxyGenerator.IProxy`. Casting the result to this type allows access to properties `readonly bool IsDirty`, `readonly HashSet<string> DirtyFields` and `void MarkAsClean()`.

Why is this good?
1. Client knows which fields have been set by Dapper's `Query*<T>`. This is especially neat if you want turn the result into a JSON without transmitting all default values.
2. Client knows whether a DB field is actually default(T) or if it was just not a part of the SELECT statement.
3. It allows combining Dapper's `Query<T>` and Dapper.Contrib's `Update<T>` statement to work on the same object. Consider this:
```
var res = conn.Query<A>("SELECT id, ColA, ColB, ColC FROM A WHERE id = @id", new { id = 5 });
res.ColA = "b";
res.ColB = 4;
conn.Update<A>(res);
```
Before, this caused havoc in the database. All Columns except id, ColA, ColB and ColC of the row with id = 5 were set to default(T). Now, if we modify `A` to be an interface (`IA`), this does what one initially thinks it does.
4. Dapper.Contrib's `Insert<T>` can make use of the defaults set in the database.

Things to note:
Dapper.Contrib's `Get<T>` and `GetAll<T>` return a clean object with all fields set. This is not exactly consistent with Dapper's `Query*<T>`, which returns a dirty object. But since `Get<T>` and `GetAll<T>` would set all fields to dirty by default, the client would want to mark them as clean in most cases, so he gets a shorter update statement afterwards. Also, `Get<T>` and `GetAll<T>` returned clean objects historically, so in order to be backwards compatible, this needs to stay this way.


NickCraver:
I want to say thanks for the contribution, but this isn't a direction we want to go with Dapper. If you want full change tracking or updates for objects I would recommend one of the ORMs that accepts these additional costs associated with change tracking like Entity Framework. It's just not something we want in the core Dapper library.

OneLineTwoBugs:
I'm sorry to hear this. Please consider my two counter arguments:
1. As mentioned above, the additional cost (in terms of runtime) are absolutely minimal. That is, if you DON'T use interfaces. If you use them, there may be some overhead.

2. Most of this code is already included in the Dapper project, especially the very hard to read and maintain code for generating the proxy classes out of interfaces by using IL Emit. Really, most of what I did is extend the IL code slightly and move part of that code from Dapper.Contrib to Dapper.

If you don't want Dapper to use track changing, then why add the functionality to Dapper.Contrib in the first place?

Last but not least: Would you consider a pull request which would only add the per-field track changing to Dapper.Contrib? (as an enhancement to the current Dirty-tracking per object)

NickCraver:
This remains a direction Dapper is not headed in, so I'm going to close this our rather than let it linger. We've split Dapper.Contrib into another repository to make that delineation clearer (the ambiguity of purpose you call out is legit, hopefully this helps). I suggest if people are interested in change tracking in general, then EntityFramework Core is a better path to investigate overall - it's quite efficient and well maintained here.

PR #1206, opened 2019-03-08
Updated Hand Coded benchmarks

yesmey:
Hand coded version did not operate under the same circumstances as others.

- Add CommandBehavior.SingleResult and CommandBehavior.SingleRow. All other framworks does this under the hood.
- Run SqlCommand.Prepare() (perhaps it should be benchmarked as "Compiled"?)
- Use GetNullableValue because it's easier to read and slightly faster
- _table.Rows was incrementally added, which made the benchmark run slower the longer it executed. Added _table.Rows.Clear() to prevent the growth.
- Changed select to use * like other benchmarks


NickCraver:
Thanks for catching this! I'm away on travel this week but I'll grab this branch and re-run the numbers when on the same desktop next week.

PR #1229, opened 2019-04-05
fix #1228 - rev sqlclient to 4.6.0

### LINKED ISSUE #1228: Update System.Data.SqlClient to 4.6.0
Dapper is currently set to using `System.Data.SqlClient` 4.4.0 which doesn't work with `TransactionScope`'s.

e.g.

```
using (var scope = new TransactionScope())
{
    using (var db = new SqlConnection(connectionString))
    {
        var result = db.Query......
    }
}
```
and same goes with `async` which requires the Transaction Flow Options enabled.

This was fixed in _I think_ 4.5.0. but 4.6.0 is now the most recent _stable_

image

This was also mentioned in Issue #1026 .

I'm happy to do a PR - if you are all happy with the update.

NOTE: As mentioned by @NickCraver in [this comment](https://github.com/StackExchange/Dapper/issues/1026#issuecomment-390473120) from that issue, we can of course _manually_ NuGet install the package and we're good to go. But the worry is that, this fix is not obviously known by consumers/developers. When we first try the queries with Dapper (as above) we will get an exception and it's not obvious that the 'fix' is to _manually_ install the more recent version.

So yeah .. this would be a nice fix IMO.

Thoughts, lovely repo org members?

mgravell:
See #1228

### DIFF
--- a/Dapper.Tests.Contrib/Dapper.Tests.Contrib.csproj
+++ b/Dapper.Tests.Contrib/Dapper.Tests.Contrib.csproj
@@ -19,7 +19,7 @@
     <PackageReference Include="Microsoft.Data.Sqlite" Version="1.1.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
     <PackageReference Include="MySqlConnector" Version="0.44.1" />
-    <PackageReference Include="System.Data.SqlClient" Version="4.4.0" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.6.0" />
     <PackageReference Include="xunit" Version="$(xUnitVersion)" />
     <PackageReference Include="xunit.runner.visualstudio" Version="$(xUnitVersion)" />
   </ItemGroup>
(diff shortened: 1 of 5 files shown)


Dapper.Tests.Contrib/Dapper.Tests.Contrib.csproj
```diff
@@ -19,7 +19,7 @@
     <PackageReference Include="Microsoft.Data.Sqlite" Version="1.1.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
     <PackageReference Include="MySqlConnector" Version="0.44.1" />
-    <PackageReference Include="System.Data.SqlClient" Version="4.4.0" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.6.0" />
```
NickCraver: Hmmm, we shouldn't actually need this ref in anything with a Dapper project ref - should only need it on Dapper and Dapper.StrongName, right? And delete the other lines here?

mgravell: You're right - pushed fix

PR #1233, opened 2019-04-07
[WIP] support F# anonymous types

enricosada:
Support F# anonymous types in query

```fsharp
let res = db.QuerySingle<{| Tid: int; Name: string; X:string; Y: int |}>("select Tid=20, Name='fdsf', X='hfds', Y=15")  
```

- [X] The search for constructor fallback to search by names sorted, if not found by position.
- [ ] The instantiation of the constructor must respect order or constructor argument

To initialize the constructor argument correctly, i tried to reoder the il fields doing the read from datareader to match constructor arguments position, not datareader name position, but i got:

```
System.InvalidOperationException : Invalid attempt to read from column ordinal '0'.  With CommandBehavior.SequentialAccess, you may only read from column ordinal '2' or greater.
```

so cannot be done like this, because read need to be sequential by column index

To minimize changes i'll try to:

- [ ] leave reading by sequential name index
- [ ] if costructor need it, instead of pushing to stack the constructor args use local var (`stloc`)
- [ ] put constructor args to stack from local vars, based on constructor arguments position

fix https://github.com/StackExchange/Dapper/issues/1226


Dapper/DefaultTypeMap.cs
```diff
@@ -77,6 +77,32 @@ internal static List<FieldInfo> GetSettableFields(Type t)
...
+            return FindConstructorBySortedNames(names, types);
+        }
+
+        private ConstructorInfo FindConstructorBySortedNames(string[] names, Type[] types)
+        {
+            if (names.Length != types.Length)
+                return null;
+
+            var byName = names.Zip(types, (name, type) => new { name, type })
+                              .OrderBy(x => x.name)
```
enricosada: String.Comparison.Ordinal?

Dapper/SqlMapper.cs
```diff
@@ -3187,9 +3187,21 @@ private static LocalBuilder GetTempLocal(ILGenerator il, ref Dictionary<Type, Lo
...
+            bool reorderConstructorParameters = false;
+            List<IMemberMap> members;
+            if (IsValueTuple(type))
+            {
+                members = GetValueTupleMembers(type, names);
+            }
+            else if (specializedConstructor != null)
+            {
+                members = names.Select(n => typeMap.GetConstructorParameter(specializedConstructor, n)).ToList();
+                reorderConstructorParameters = members.Select((m, i) => i != m.Parameter.Position).Any(x => x);
```
enricosada: check empty constructor because valid

Dapper/DefaultTypeMap.cs
enricosada: done, `StringComparer.OrdinalIgnoreCase`

Dapper/SqlMapper.cs
enricosada: done

enricosada:
done, @mgravell @NickCraver ready to review

enricosada:
@mgravell anything i can help with for this PR?

enricosada:
@NickCraver @mgravell any specific issue with this PR i need to address?
happy to help to have it merged

mgravell:
I totally understand. It isn't personal - I simply haven't had chance to look yet - I'm juggling time between multiple projects.  The existence of tests in the PR is great. I'll try and look for the next drop. Sorry for delays.

enricosada:
Np at all @mgravell , i know it may happen, it was just to check if there was something todo

meanwhile i'll rebase it

thanks a lot!

Dapper/DefaultTypeMap.cs
```diff
@@ -77,6 +77,27 @@ internal static List<FieldInfo> GetSettableFields(Type t)
         /// <param name="types">DataReader column types</param>
         /// <returns>Matching constructor or default one</returns>
         public ConstructorInfo FindConstructor(string[] names, Type[] types)
+        {
+            return FindConstructorByColumns(names, types) // first search constructor using the same position of names
+                ?? FindConstructorBySortedNames(names, types); // if not found, search by names sorted (F# anonymous records)
+        }
+
+        private ConstructorInfo FindConstructorBySortedNames(string[] names, Type[] types)
```
mgravell: isn't most of this method just: `Array.Sort(names, types);` ? that would be a **lot** more efficient, if so

NickCraver:
@enricosada did you see Marc's comment above? We'd love to get this in, but it could be a fair bit more efficient if intent matches.

enricosada:
@NickCraver no, i missed that sry, let me check. thx for ping

Dapper/DefaultTypeMap.cs
enricosada: I tried to understand how to use `Array.Sort(names, types)`, but dunno how to use it here, because `names` is not the keys for `types` array.

`names` contains the ctor arg name, `types` is the ctor arg type.
So `names[0]` is the name of the first arg, and `types[0]` is it's expected type

That's why i do

```csharp
var byName = names.Zip(types, (name, type) => new { name, type })
```

to pair each name with his type, so i can order

```
`.OrderBy(x => x.name, StringComparer.OrdinalIgnoreCase`
```

these maintaining the sync.

i'll later split it two array again

```csharp
var namesByName = byName.Select(x => x.name).ToArray();
var typesByName = byName.Select(x => x.type).ToArray();
```

because i want to reuse the existing implementation who find the constructor

```
return FindConstructorByColumns(namesByName, typesByName);
```

enricosada: I can maybe use `Array.Sort` but i need anyway to:

- create the paired array (to maintain sync)
- sort by ordinal the `names` array into `namesSorted`
- `Array.Sort` the paired array with the `namesSorted` array and with a custom comparer (to extract the name pair)

i'll also need to use a tuple instead of anynoumn type, if i need a custom comparer i think.

So will be two sort (instead of just one) and the same number of temp array instantiations.

enricosada: The only enhancement i can think of is to replace

```csharp
var namesByName = byName.Select(x => x.name).ToArray();
var typesByName = byName.Select(x => x.type).ToArray();
```

with

```csharp
string[] namesByName = new string[ byName.Length ];
Type[] typesByName = new Type[ byName.Length ];

for (int i = 0; i < byName.Length; ++i) {
   namesByName[i] = byName[i].name;
   typesByName[i] = byName[i].type;
}
```

who should be i think a bit faster, but less readable.

what do you think @mgravell ?

enricosada:
@mgravell i added some comments, because i not understood your feedback
@NickCraver happy to fix, but dunno how to improve this atm, it fix the bug atm and i didnt see a big perf hit for F#

mgravell:
Hi @enricosada - I got nudged by a mutual friend to remind me to look at this again; sorry for delay. Can I check - the confusion we're talking about - is this the zip vs dual sort vs arrays thing?

My understanding - and please do correct me if I'm wrong here - is that we have two arrays; one of the names, one of the types, matching by index; so if we have 3 columns, we might have

names: "def", "ghi", "abc"
types: Int32, String, Single

as I read the code, the intent here is to sort by the names, preserving the respective positions in both arrays, so we end up with:

names: "abc", "def", "ghi"
types: Single, Int32, String

is that correct? If so, that's exactly what the dual array version of Array.Sort *does*:

``` c#
        string[] names = { "def", "ghi", "abc" };
        Type[] types = { typeof(int), typeof(string), typeof(float) };

        Array.Sort(names, types);

        Console.WriteLine($"names: {string.Join(", ", names)}");
        Console.WriteLine($"types: {string.Join(", ", types.Select(x => x.Name))}");

        /* output:
            names: abc, def, ghi
            types: Single, Int32, String
         */
```

Have I misunderstood?

enricosada:
> Have I misunderstood?

hi @mgravell

Maybe i didnt understood what the dual array version of Array.Sort does.
I tried locally when you mentioned it but didnt understood.

I think that's exactly what i want, sort two arrays based on values of first array and maintaining the sync. 

Let me rebase and use Array.Sort

mgravell:
> sort two arrays based on values of first array and maintaining the sync

to confirm, yes: that is an accurate description of what `Array.Sort` does with the overload that takes two arrays

enricosada:
@mgravell applyed your feedback, it's now using `Array.Sort` instead of `zip+sort`
I tested locally with other cases and seems fine too.
So if appveyor is green, should be ok

Thanks a lot btw @mgravell , TIL about Array.Sort overload :D

TravisLeithVeloqx:
Can I ask about the status of this? I would like to use this feature in my project and can clone this branch for now, but only if it is expected to make it into master.

atlemann:
I just got hit by this as well. Would be awesome if it made it in a new release!

enricosada:
@mgravell i think this is ready for review and merge.

happy to do any change required

TravisLeithVeloqx:
I keep coming up against this issue. Is there a release schedule in mind?

mgravell:
will re-review; sorry for dropping between cracks

mgravell:
Code seems good; checking the tests locally for regressions - seems fine; you're happy that it is working from F#?

TravisLeithVeloqx:
@mgravell @enricosada 
Until now I have not actually attempted to use this.
Today I pulled this branch and built it. I added the Dapper.dll to my the project I am currently working on and attempted to use an anonymous record. I get the same `parameterless default constructor` error. 

The offending code is
``` F#
let parameters = Map ["ids", box ids; "dates", box dates]
use conn = new MySqlConnection(connectionString)

conn
|> dapperMapParametrizedQuery<{|id:int; date:DateTime; variance:float|}> sql parameters
|> Seq.groupBy (fun x -> x.date)
|> Seq.map (fun (dt, sq) -> dt, sq |> Seq.map (fun x -> {id = x.id; variance = x.variance})) |> Ok
```

So either I have built it wrong, or am using it wrong.

enricosada:
Let me check your repro

From mobile

--
Enrico Sada
@enricosada
________________________________
Da: TravisLeithVeloqx <notifications@github.com>
Inviato: Thursday, April 16, 2020 8:59:59 AM
A: StackExchange/Dapper <Dapper@noreply.github.com>
Cc: Enrico Sada <enrico@sada.io>; Mention <mention@noreply.github.com>
Oggetto: Re: [StackExchange/Dapper] support F# anonymous types (#1233)


@mgravell<https://github.com/mgravell> @enricosada<https://github.com/enricosada>
Until now I have not actually attempted to use this.
Today I pulled this branch and built it. I added the Dapper.dll to my the project I am currently working on and attempted to use an anonymous record. I get the same parameterless default constructor error.

The offending code is

let parameters = Map ["ids", box ids; "dates", box dates]
use conn = new MySqlConnection(connectionString)

conn
|> dapperMapParametrizedQuery<{|id:int; date:DateTime; variance:float|}> sql parameters
|> Seq.groupBy (fun x -> x.date)
|> Seq.map (fun (dt, sq) -> dt, sq |> Seq.map (fun x -> {id = x.id; variance = x.variance})) |> Ok

So either I have built it wrong, or am using it wrong.

ChechyLevas:
@enricosada is this working?

NickCraver:
I updated this branch to master - still prepared to merge but we don't use F# and would really appreciate some community feedback on whether this works as intended.

enricosada:
@NickCraver thanks a lot.

I’ll check tomorrow morning

brase:
@enricosada  How can I test it? Should it work with unordered ctor parameters?

realvictorprm:
pew, has been a while. Maybe I should take a look into that.

realvictorprm:
> @enricosada How can I test it? Should it work with unordered ctor parameters?

So general suggestion from my side, I would have gone down the route and added another test folder with an F# project that depends on Dapper and uses it with anon records to make sure that the feature is functional. Anything against going that far?
Alternatively an F# library project could be added, that contains an anon record with a type alias, such that it can be consumed by C# and used for directly piping into the C# functions to test.

realvictorprm:
@NickCraver are you still open for having this change? If yes, could we reiterate on what you want to see from your side to be changed to get it merged, considering how long it has been since work has been done on this PR & topic?

Issue #1213, opened 2019-03-22
Dapper.Contrib character case bug


moabtools:
Hi! 
I have table in postgres created with field names in quotes (i.e. case sensitive):
```
create table "buttons" (
"Id" SERIAL PRIMARY KEY,
"UserId"  INTEGER,
"Phone"  VARCHAR,
"NumberType" INTEGER,
...
```
and c# class also with case sensitive properties:
```
    public class Button
    {
        public Int32 Id { get; set; }
        public Int32 UserId { get; set; }
        public String Phone { get; set; }
        public NumberType NumberType { get; set; }
        ...
```

When I try to insert record with InsertAsync, it throws an error
```
42703: column "id" does not exist
```
as you can see, column "id" is written here in lower case, because the function InsertAsync skips id-column during query creation.

UpdateAsync works perfectly because it uses "Id" in where clause.

But when I change the name of column in database to "id" - InsertAsync works good, but UpdateAsync fails by the same reason.

Please help.

Renerick:
Can confirm the same behaviour with postgresql. No matter what name it is if you apply the `Key` attribute, Dapper.Contrib will use it in lower case. For example, it got this error 
```
PostgresException: 42703: column "fooid" does not exist
```
with this class
```cs
[Table("bar")]
public class Bar
{
    [Key]
    public long FooId { get; set; }
    ...
}
```
This does not apply to other properties though.

Renerick:
So, basically, the reason is in the way Postgres treats identifiers. If you create the table in double quotes you will have you use double quotes always (and Dapper.Contrib does not). If you will use names without quotes, it will be case insensitive the problem must be solved

rclarke2050:
I found a way around this field name problem for the time being.

in my Repository Constructor i used:

DapperExtensions.DapperExtensions.SqlDialect = new DapperExtensions.Sql.PostgreSqlDialect();

and in the Insert Method:   db.Insert(item); 
and in the Updaet Method: db.Update(item);

i did not use the SqlMapper approach, nor did I use the SqlMapperExtensions.Insert()

Ive tried on .Net Core 2.1 Lambda, and AWS Aurora 10.7.

Dapper v2.0.30
Dapper.Contrib 2.0.30
DapperExtensions.DotnetCore 1.0.1

I hope this helps others for now....in the interim.

goforgold:
Hi @rclarke2050 

Does it mean that we have to include DapperExtenssions library to achieve this?

[...]

PR #1236, opened 2019-04-11
fix inconsistent Contrib column name formatting in postgres and test

nathan-c:
I realise this is technically a change in behaviour but it actually represents a standardisation in how postgres column names are treated. 

This is intended to add quotes to all column names when using postgres. The current behaviour omits quotes for id column on get and also on insert when getting an id back but uses quotes for all other operations and columns.

My personal preference would be to support snake_case names and not use quotes at all as this is more in line with postgres standard practise. I realise this has been talked about in #722 

Note that this change does not affect the fact the Contrib currently does not quote table names so there is still an inconsistency between table and column name handling.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -25,12 +25,18 @@ namespace Dapper.Contrib.Extensions
         public static async Task<T> GetAsync<T>(this IDbConnection connection, dynamic id, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
         {
             var type = typeof(T);
+            var adapter = GetFormatter(connection);
+
             if (!GetQueries.TryGetValue(type.TypeHandle, out string sql))
             {
                 var key = GetSingleKey<T>(nameof(GetAsync));
                 var name = GetTableName(type);
 
-                sql = $"SELECT * FROM {name} WHERE {key.Name} = @id";
+                var sb = new StringBuilder("select * from ");
+                sb.Append(name);
+                sb.Append(" where ");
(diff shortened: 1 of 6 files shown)


NickCraver:
We'll be addressing this in #722 (v2) as mentioned above...but we cannot take such a change in v1. It's absolutely a breaking change for many users...we can't ship this change in a minor version. Honestly, I wouldn't support doing it in a major version. I think we'll maintain the current behavior in v2 but allow users to have full control easily - most people using 1-2 lines of config to use the rules they want (some in-box).

nathan-c:
To confirm. Currently `Get` and `Insert` expect lower case keys but `Delete` and `Update` require a case sensitive key column. I have updated the branch to show all the failing tests if behaviour stays the same.

I didn't expect this behaviour during a MSSQL -> PG migration. 

How can I stay up to date (or even help) with v2 work?

nathan-c:
This behaviour is also noted in a recent issue #1213

rclarke2050:
I found a way around this field name problem for the time being.

in my Repository Constructor i used:

DapperExtensions.DapperExtensions.SqlDialect = new DapperExtensions.Sql.PostgreSqlDialect();

and in the Insert Method:   db.Insert(item); 
and in the Updaet Method: db.Update(item);

i did not use the SqlMapper approach, nor did I use the SqlMapperExtensions.Insert()

Ive tried on .Net Core 2.1 Lambda, and AWS Aurora 10.7.

Dapper v2.0.30
Dapper.Contrib 2.0.30
DapperExtensions.DotnetCore 1.0.1

I hope this helps others for now....in the interim.

NickCraver:

PR #1242, opened 2019-04-20
Fixes mapping for C# tuples with more than seven elements

### LINKED ISSUE #982: Mapping to a ValueTuple with 8 or more elements only retrieves some columns
I'm not sure if mapping to ValueTuples is supported 100%, but I noticed that if the tuple has 8 or more elements, only the first 7 are mapped and the rest are filled with default values. My guess is that tuples with more items are defined as an 8-tuple with a TRest parameter that is another tuple.

jnm2:
Fixes #982. I think I made the minimum change possible.

/cc @mgravell

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -1239,7 +1239,7 @@ namespace Dapper
         }
 
         /// <summary>
-        /// Perform a multi-mapping query with 2 input types. 
+        /// Perform a multi-mapping query with 2 input types.
         /// This returns a single type, combined from the raw types via <paramref name="map"/>.
         /// </summary>
         /// <typeparam name="TFirst">The first type in the recordset.</typeparam>
@@ -1259,7 +1259,7 @@ namespace Dapper
             MultiMap<TFirst, TSecond, DontMap, DontMap, DontMap, DontMap, DontMap, TReturn>(cnn, sql, map, param, transaction, buffered, splitOn, commandTimeout, commandType);
 
         /// <summary>
-        /// Perform a multi-mapping query with 3 input types. 
+        /// Perform a multi-mapping query with 3 input types.
         /// This returns a single type, combined from the raw types via <paramref name="map"/>.
         /// </summary>
         /// <typeparam name="TFirst">The first type in the recordset.</typeparam>
@@ -1280,7 +1280,7 @@ namespace Dapper
             MultiMap<TFirst, TSecond, TThird, DontMap, DontMap, DontMap, DontMap, TReturn>(cnn, sql, map, param, transaction, buffered, splitOn, commandTimeout, commandType);
 
         /// <summary>
-        /// Perform a multi-mapping query with 4 input types. 
+        /// Perform a multi-mapping query with 4 input types.
         /// This returns a single type, combined from the raw types via <paramref name="map"/>.
         /// </summary>
         /// <typeparam name="TFirst">The first type in the recordset.</typeparam>
@@ -1302,7 +1302,7 @@ namespace Dapper
             MultiMap<TFirst, TSecond, TThird, TFourth, DontMap, DontMap, DontMap, TReturn>(cnn, sql, map, param, transaction, buffered, splitOn, commandTimeout, commandType);
 
         /// <summary>
-        /// Perform a multi-mapping query with 5 input types. 
+        /// Perform a multi-mapping query with 5 input types.
         /// This returns a single type, combined from the raw types via <paramref name="map"/>.
         /// </summary>
         /// <typeparam name="TFirst">The first type in the recordset.</typeparam>
@@ -1325,7 +1325,7 @@ namespace Dapper
             MultiMap<TFirst, TSecond, TThird, TFourth, TFifth, DontMap, DontMap, TReturn>(cnn, sql, map, param, transaction, buffered, splitOn, commandTimeout, commandType);
 
         /// <summary>
-        /// Perform a multi-mapping query with 6 input types. 
+        /// Perform a multi-mapping query with 6 input types.
         /// This returns a single type, combined from the raw types via <paramref name="map"/>.
         /// </summary>
         /// <typeparam name="TFirst">The first type in the recordset.</typeparam>
@@ -1374,7 +1374,7 @@ namespace Dapper
             MultiMap<TFirst, TSecond, TThird, TFourth, TFifth, TSixth, TSeventh, TReturn>(cnn, sql, map, param, transaction, buffered, splitOn, commandTimeout, commandType);
 
         /// <summary>
-        /// Perform a multi-mapping query with an arbitrary number of input types. 
+        /// Perform a multi-mapping query with an arbitrary number of input types.
         /// This returns a single type, combined from the raw types via <paramref name="map"/>.
         /// </summary>
         /// <typeparam name="TReturn">The combined type to return.</typeparam>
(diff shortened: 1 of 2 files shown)


Dapper/SqlMapper.cs
```diff
@@ -3097,6 +3068,121 @@ private static LocalBuilder GetTempLocal(ILGenerator il, ref Dictionary<Type, Lo
...
+            else
+            {
+                GenerateDeserializerFromMap(type, reader, startBound, length, returnNullIfFirstMissing, il);
+            }
+
+            var funcType = System.Linq.Expressions.Expression.GetFuncType(typeof(IDataReader), returnType);
+            return (Func<IDataReader, object>)dm.CreateDelegate(funcType);
+        }
+
+        private static void GenerateValueTupleDeserializer(Type valueTupleType, IDataReader reader, int startBound, int length, bool returnNullIfFirstMissing, ILGenerator il)
```
jnm2: @mgravell (For review, before I forget) The parameter `returnNullIfFirstMissing` is unused. Should I be using it, or can I rule that out and remove the parameter since the deserialization type is a value type?

Dapper/SqlMapper.cs
```diff
@@ -3097,6 +3068,121 @@ private static LocalBuilder GetTempLocal(ILGenerator il, ref Dictionary<Type, Lo
...
+
+            var enumDeclareLocal = -1;
+
+            for (var i = 0; i < languageTupleElementTypes.Count; i++)
+            {
+                var targetType = languageTupleElementTypes[i];
+
+                if (i < length)
+                {
+                    LoadReaderValueOrBranchToDBNullLabel(
```
jnm2:
```suggestion
                    LoadReaderValueOrBranchToDbNullLabel(
```

jnm2:
@mgravell Would you be able to set my expectations for an ETA? I hate to bug you.

mgravell:
You're right, I need to find the time. I'll see if I can review on Tuesday
(long family weekend here).

Dapper/SqlMapper.cs
mgravell: yeah, I think you're fine to remove it... but could always just add a test that adds a `(int,int)?` type and zero rows?

Dapper/SqlMapper.cs
```diff
@@ -3097,6 +3068,121 @@ private static LocalBuilder GetTempLocal(ILGenerator il, ref Dictionary<Type, Lo
...
+        private static void GenerateValueTupleDeserializer(Type valueTupleType, IDataReader reader, int startBound, int length, bool returnNullIfFirstMissing, ILGenerator il)
+        {
+            var currentValueTupleType = valueTupleType;
+
+            var constructors = new List<ConstructorInfo>();
+            var languageTupleElementTypes = new List<Type>();
+
+            while (true)
+            {
+                var arity = int.Parse(currentValueTupleType.Name.Substring("ValueTuple`".Length));
```
mgravell: needs to be invariant?

mgravell: (meaning: culture)

Dapper/SqlMapper.cs
```diff
@@ -3097,6 +3068,121 @@ private static LocalBuilder GetTempLocal(ILGenerator il, ref Dictionary<Type, Lo
...
+
+                foreach (var field in currentValueTupleType.GetFields(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly))
+                {
+                    if (field.Name == "Rest")
+                    {
+                        restField = field;
+                    }
+                    else if (field.Name.StartsWith("Item", StringComparison.Ordinal))
+                    {
+                        var elementNumber = int.Parse(field.Name.Substring("Item".Length));
```
mgravell: ditto

mgravell:
I've given it a once over - looks kinda great, which is nice; I'm cautious about the `int.Parse` because: culture, so that might need a check; it is also possible that I'm wrong and it works fine already. I want to read it from scratch a second time, and look through the tests - but: looking very promising.

Dapper/SqlMapper.cs
jnm2: Oops, this is something I keep forgetting to do. Thanks!

jnm2:
Awesome! I'll write that nullable tuple test and fix the culture sensitivity either late tonight or sometime tomorrow and ping you when I'm finished.

After you get a chance to do the final review, I'll be curious if there's a prerelease feed.

mgravell:
Preview feed on myget via appveyor:
https://www.myget.org/feed/stackoverflow/package/nuget/Dapper

Dapper/SqlMapper.cs
jnm2: @mgravell `connection.QuerySingleOrDefault<(int, int)?>("select null where 0 = 1");` does not result in `returnNullIfFirstMissing` becoming true. Should I be trying something with multi-mapping? (Looks like the parameter was added in https://github.com/StackExchange/Dapper/commit/e11257fdf82a4307523e09859e31da3d495877a6)

jnm2: I wrote this test against master to prove that nullable C# tuples have never been supported:

```cs
[Fact]
public void NullableTupleMultiMap()
{
    var outer = connection.Query<Foo, (int, int)?, Foo>(
        "select * from (values (42, null, null), (42, 43, 44)) as Foo (A, B1, B2)",
        (foo, b) =>
        {
            foo.B = b;
            return foo;
        },
        splitOn: "B1");

    Assert.Equal(42, outer.First().A);
    Assert.Null(outer.First().B);

    Assert.Equal(42, outer.Last().A);
    Assert.Equal(43, outer.Last().B.Value.Item1);
    Assert.Equal(44, outer.Last().B.Value.Item2);
}

private sealed class Foo
{
    public int A { get; set; }
    public (int, int)? B { get; set; }
}
```

And therefore `returnNullIfFirstMissing` is not meaningful yet and I can remove it.

jnm2:
@mgravell Ready for the second round!

jnm2:
The force pushes above are [adding `CultureInfo.InvariantCulture`](https://github.com/StackExchange/Dapper/compare/31bf76f1eb4c9b5897ac1b43dd559413d6efe934..93245d2a390aa7c2064f644be7e60305f0011d76) and [removing `returnNullIfFirstMissing`](https://github.com/StackExchange/Dapper/compare/93245d2a390aa7c2064f644be7e60305f0011d76..f61c3ce6ec1dc165fa4c6ff4827c666eae3547d6), addressing both your comments.

Do you think you might have a chance coming up next week to review again? I appreciate the time you've been able to give.

jnm2:
@mgravell The workaround at the moment is to introduce new private throwaway structs, so this keeps coming up. Is this PR a very risky one? Would it help if I added more tests?

mgravell:
let's do this...

jnm2:
We'll start using the MyGet feed right away! Thank you so much! 🎉

mgravell:
No: thank you. It was your contribution, and it is very much appreciated. Delay here was mostly my time juggling. Thanks again.

jnm2:
As an OSS maintainer, I know PRs are demanding. I appreciate you reviewing and getting this through.

Dapper/SqlMapper.cs
```diff
@@ -3400,9 +3391,115 @@ private static LocalBuilder GetTempLocal(ILGenerator il, ref Dictionary<Type, Lo
...
+            }
+        }
+
+        private static void LoadReaderValueOrBranchToDBNullLabel(ILGenerator il, int index, ref int enumDeclareLocal, int? valueCopyLocal, Type colType, Type memberType, out Label isDbNullLabel)
+        {
+            isDbNullLabel = il.DefineLabel();
+            il.Emit(OpCodes.Ldarg_0); // stack is now [...][reader]
+            EmitInt32(il, index); // stack is now [...][reader][index]
+            il.Emit(OpCodes.Dup);// stack is now [...][reader][index][index]
+            il.Emit(OpCodes.Stloc_0);// stack is now [...][reader][index]
```
jnm2: This line introduced bug https://github.com/StackExchange/Dapper/issues/1279, fixed in https://github.com/StackExchange/Dapper/pull/1280.

PR #1250, opened 2019-04-30
KeyPropertyNameMapper for overriding "id" assumption for key property name

ingeniousclown:
We have a different naming convention for our identity columns and POCOs than Dapper.Contrib assumes.  There is a mechanism in place for resolving table names but not one for resolving the name of the ID column.


pjt33:
Why only allow mapping the key property's name? Going the other way, `ITypeMapper` allows mapping arbitrary column names to arbitrary `MemberInfo`s.

PR #1254, opened 2019-05-06
Fix Rainbow Database.Dispose() so that it never throws and not overrides the active exception (if any)

dmytro-gokun:
https://github.com/StackExchange/Dapper/issues/1252


NickCraver:
Good points! I cleaned up the implementation slightly here because we'd still only want to take action if the connection isn't already closed. Once the build runs, we'll get this merged in.

PR #1255, opened 2019-05-06
Add ability to use an external transaction with Rainbow database

dmytro-gokun:
+ several small improvements to the transaction management.

### DIFF
--- a/Dapper.Rainbow/Database.Async.cs
+++ b/Dapper.Rainbow/Database.Async.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Data;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Dapper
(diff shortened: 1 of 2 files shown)


mgravell:
This is an API-breaking change; not sure we'll be able to find a way to merge it

dmytro-gokun:
@mgravell Okay, i see. My main point was to add SetTransaction method. The rest was cosmetic. Do you think creating a new PR with just SetTransaction added makes sense?

dmytro-gokun:
@mgravell Also, i'm not quite sure why these changes are API-breaking. I mean, i do not see how this may brake any existing working code. Can you elaborate?

NickCraver:
I can help a little here: changing the return type from `void` to a type is a breaking change, e.g. a library that was compiled against this one would break until it was recompiled. That's a binary break, rather than a code break.

See the IL difference here between the method calls (the `pop` vs. the `nop`):
image

That's the API compat break, but the other scenarios would break in unexpected ways. For example, this inadvertently (I think) drops support for nested transactions.

dmytro-gokun:
@NickCraver You are right. But there was a major release several days ago, where some other of my PRs were included. Some of them were binary incompatible also, i guess. Nobody expects binary compatibility between major releases. Source compatibility is not broken though. May be this could be included in some v3 release in 2025 :)

I do not get your point about nested transactions though. Could you elaborate?

mgravell:
Exactly one API was not binary compatible, and that was an API that I expect *very very few* people to have used. The 2.* release was a grudging "technically this isn't binary compatible, but in reality: you're going to be fine". The `BeginTransaction` API, on the other hand, is *very* likely to be commonly used.

> Nobody expects binary compatibility between major releases. 

Oh yes they do. They might not be *correct* to expect it, but: trust me, they do. Sometimes it is *necessary*, but we need to be careful to manage expectations then. We know we're going to *shatter* compatibility in v3, for example - and we'll be shouting from the hills about it. Changing the return type of `BeginTransaction` would be perfectly fine *at that point*. But if the *important* part of the PR is `SetTransaction`, I would recommend isolating that aspect.

dmytro-gokun:
@mgravell Gotcha. Well, basically idea was to make transaction both gettable & settable, so it's easy to have one transaction spanning over a Dapper connection and a EF connection (or anything else). I will work on that PR.

NickCraver:
> I do not get your point about nested transactions though. Could you elaborate?

Yep, some providers support nested transactions, due to the check here on begin, this would now throw and be unsupported :) Previously, it would let the provider handle such concerns (which is the correct path...we can't know what is or isn't supported at this level).

NickCraver:
This still needs love w.r.t. nested transactions in a database, but would be fine for the next major release after that. To eliminate some of the coupling here, we are moving things like Dapper.Rainbow out of the main repository into their own repos...so that they can also have their own major/breaking releases. Please see #1658 for details there.

PR #1257, opened 2019-05-06
Expose the underlying connection from Rainbow Database so it's easier to use Dapper alongside other DB frameworks

dmytro-gokun:
https://github.com/StackExchange/Dapper/issues/1253


mgravell:
seems fine to me; merging

dmytro-gokun:
Thanks!

PR #1280, opened 2019-06-13
Stop setting current index local which is nonexistent in ValueTuple deserializer

### LINKED ISSUE #1279: InvalidProgramException on prerelease when first field is a non-numeric type
@mgravell It seems very likely that this one is on me via https://github.com/StackExchange/Dapper/pull/1242. I'll get a full repro and fix it ASAP.

https://www.myget.org/feed/stackoverflow/package/nuget/Dapper/1.60.14 works fine with `(int, string, string, DateTime?, DateTime?, DateTime?, decimal?, DateTime?, string, int?, DateTime?, DateTime?, int?)` rows.

With `(string, byte[], byte[])`, it breaks:

> System.InvalidProgramException:
> Common Language Runtime detected an invalid program.
>    at Deserialize669fc2c6-0ab9-4b1c-ae66-ee2bfbcc0010(IDataReader )
>    at Dapper.SqlMapper.<QueryRowAsync>d__34`1.MoveNext() in C:\projects\dapper\Dapper\SqlMapper.Async.cs:line 500

jnm2:
Fixes #1279, a bug introduced in https://github.com/StackExchange/Dapper/pull/1242.

@mgravell I'm sorry I broke master. Hopefully the fix is super easy to review.

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -3178,7 +3178,7 @@ namespace Dapper
 
         private static void GenerateDeserializerFromMap(Type type, IDataReader reader, int startBound, int length, bool returnNullIfFirstMissing, ILGenerator il)
         {
-            il.DeclareLocal(typeof(int));
+            var currentIndexDiagnosticLocal = il.DeclareLocal(typeof(int));
             il.DeclareLocal(type);
             il.Emit(OpCodes.Ldc_I4_0);
             il.Emit(OpCodes.Stloc_0);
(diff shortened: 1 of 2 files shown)


Dapper/SqlMapper.cs
```diff
@@ -3379,7 +3383,7 @@ private static void GenerateDeserializerFromMap(Type type, IDataReader reader, i
             }
             il.MarkLabel(allDone);
             il.BeginCatchBlock(typeof(Exception)); // stack is Exception
-            il.Emit(OpCodes.Ldloc_0); // stack is Exception, index
+            il.Emit(OpCodes.Ldloc, currentIndexDiagnosticLocal); // stack is Exception, index
```
jnm2: I'm going to submit a PR after this one which requires a LocalBuilder instance to be passed to each stloc and ldloc to make it easier to follow what's happening and to add guardrails for future refactoring.
It would have been hard to ignore the mistake in my first PR if LocalBuilders were passed around instead of using magic indexes without passing them around.

jnm2:
@mgravell The check that failed (WIP) is saying that it needs someone with StackExchange org admin permissions to accept new permissions.

mgravell:
Thanks for resolving so promptly - merged.

jnm2:
@mgravell Using the new prerelease, everything seems to be working as anticipated.

Dapper/SqlMapper.cs
jnm2: https://github.com/StackExchange/Dapper/pull/1282

PR #1282, opened 2019-06-15
Use LocalBuilders rather than hardcoding or passing local indexes

jnm2:
As promised at https://github.com/StackExchange/Dapper/pull/1280#discussion_r293167146.

This should help with refactoring bugs like https://github.com/StackExchange/Dapper/issues/1279. If `OpCodes.Stloc, variableName` had been used rather than `OpCodes.Stloc_0`, the mistake would have been impossible to overlook.

Given the complexity of the IL-generating methods, it would probably be an okay idea to ban the use of any `*loc*` opcodes unless a LocalBuilder is passed.


Dapper/SqlMapper.cs
```diff
@@ -3608,64 +3611,6 @@ private static MethodInfo ResolveOperator(MethodInfo[] methods, Type from, Type
             return null;
         }
 
-        private static void LoadLocal(ILGenerator il, int index)
```
jnm2: These methods duplicate code that has been part of the BCL:
https://referencesource.microsoft.com/#mscorlib/system/reflection/emit/ilgenerator.cs,d4860f97f6e651d9
https://source.dot.net/#System.Private.CoreLib/src/System/Reflection/Emit/ILGenerator.cs,d4860f97f6e651d9

Dapper/SqlMapper.cs
```diff
@@ -2391,19 +2391,21 @@ internal static IList<LiteralToken> GetLiteralTokens(string sql)
             var il = dm.GetILGenerator();
 
             bool isStruct = type.IsValueType();
-            bool haveInt32Arg1 = false;
```
jnm2: Name should have been `haveInt32Loc1`

jnm2:
@mgravell Does this look good to you?

mgravell:
Looks good; not sure whether this internally optimizes for the 0/1/2/3/S cases, but... probably fine!

mgravell:
ah, I just saw your comment addressing exactly that, thanks; looks good; merging, thanks!

Issue #1234, opened 2019-04-08
TypeHandler not used for null values


epolikarpov:
Custom type handler is not called by Query* or Execute* methods for null values. In the following sample for Microsoft SQL I expect to insert DbNull value into the table if Foo.Data is null. Instead, an exception is thrown due to an invalid query parameter type (nvarchar instead of varbinary).

```csharp
using Dapper;
using System;
using System.Data;
using System.Data.SqlClient;

namespace DapperBug
{
    class Program
    {
        static void Main(string[] args)
        {
            SqlMapper.AddTypeHandler(new ValuesHandler());

            var connectionString = "Data Source =.; Initial Catalog = tempdb; Integrated Security = True;";
            var sql = @"CREATE TABLE #Test (Id int IDENTITY(1,1) PRIMARY KEY, Data binary(2));
                        INSERT INTO #Test (Data) VALUES (@Data); SELECT SCOPE_IDENTITY() AS Id";
            using (var connection = new SqlConnection(connectionString))
            {
                var foo = new Foo();
                int id = connection.ExecuteScalar<int>(sql, foo);
                Console.WriteLine(id);
            }
        }
    }

    class Values
    {
        public Values() { }
        public Values(byte first, byte second)
        {
            First = first;
            Second = second;
        }

        public byte First { get; set; }
        public byte Second { get; set; }
    }

    class Foo
    {
        public int Id { get; set; }
        public Values Data { get; set; }
    }

    class ValuesHandler : SqlMapper.TypeHandler<Values>
    {
        public override Values Parse(object value)
        {
            if (value is byte[] bytes)
            {
                return new Values(bytes[0], bytes[1]);
            }

            return null;
        }

        public override void SetValue(IDbDataParameter parameter, Values value)
        {
            parameter.DbType = DbType.Binary;
            if (value == null)
            {
                parameter.Value = DBNull.Value;
            }
            else
            {
                parameter.Value = new byte[2] { value.First, value.Second };
            }
        }
    }
}
```

The resulting request on the server:

```sql
exec sp_executesql N'CREATE TABLE #Test (Id int IDENTITY(1,1) PRIMARY KEY, Data binary(2));
INSERT INTO #Test (Data) VALUES (@Data); SELECT SCOPE_IDENTITY() AS Id',N'@Data nvarchar(4000)',@Data=NULL
```
I'm trying to use a custom type handler for `Systems.Drawing.Image`

Quppa:
[...]

PR #1284, opened 2019-06-17
Issue #1234 - TypeHandler: Allow modification of parameter when value is null

Quppa:
Provides a workaround for #1234.

In short, implementers of `SqlMapper.TypeHandler<T>` don't have the chance to set `DbType` for a null database parameter of type `T`. `DbType` seems to default to `DbType.String`, and this will cause (for example) an `INSERT` into a `VARBINARY` column to fail with the error `Implicit conversion from data type nvarchar to varbinary is not allowed. Use the CONVERT function to run this query.`.

Modifying `ITypeHandler.SetValue` to call `SetValue` even when the value is `DbNull` would be a breaking change for existing code that expects only non-null values, so my fix is to add a `SetNullValue` method which can optionally be overridden so that `DbType` can be modified.

### DIFF
--- a/Dapper.Tests/TypeHandlerTests.cs
+++ b/Dapper.Tests/TypeHandlerTests.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Data;
+using System.Data.SqlClient;
 using System.Linq;
 using System.Reflection;
 using Xunit;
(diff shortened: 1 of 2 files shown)


jseekamp:
Hit this issue today as well. Any chance this pull request could be pulled in?

axm:
Hit this issue too. What are the plans for this feature?

nschlepp:
We are also running into this issue, are there plans to merge this up?

lukblazewicz:
Same here. Is there any workaround for forcing null values to be passed through a handler?

PR #1295, opened 2019-07-12
Add ExecuteReaderAsync overloads that return Task<DbDataReader>

### LINKED ISSUE #1140: connection.ExecuteReaderAsync returns IDataReader which has no ReadAsync
It's currently necessary to cast the returned IDataReader to DbDataReader in order to do anything asynchronous with it:

```c#
using (var reader = (DbDataReader)await connection.ExecuteReaderAsync(command).ConfigureAwait(false))
{
    while (await reader.ReadAsync(cancellationToken).ConfigureAwait(false))
    {
        // ..
    }
}
```

The extension method takes an IDbConnection, so this makes sense:

https://github.com/StackExchange/Dapper/blob/48f6754ad69c5bc196612ec23f4887062fa87e17/Dapper/SqlMapper.Async.cs#L1116

Would you please define this extension method next to it?

```c#
public static Task<DbDataReader> ExecuteReaderAsync(this DbConnection cnn, CommandDefinition command)
```

That way this code sample just works:

```c#
using (var reader = await connection.ExecuteReaderAsync(command).ConfigureAwait(false))
{
    while (await reader.ReadAsync(cancellationToken).ConfigureAwait(false))
    {
        // ..
    }
}
```

jnm2:
Closes #1140 

In order to avoid a binary breaking change, new overloads must be added rather than changing the return type of the existing methods.

This causes a potential source-breaking change because the new DbConnection overloads will take precedence over the existing IDbConnection overloads when users recompile their projects:

```cs
// CS0029: Cannot implicitly convert type
// 'System.Threading.Tasks.Task<System.Data.Common.DbDataReader>'
// to 'System.Threading.Tasks.Task<System.Data.IDataReader>'

Task<IDbDataReader> readerTask = dbConnection.ExecuteReaderAsync(...);

// Or

Foo(dbConnection.ExecuteReaderAsync(...));

void Foo(Task<IDbDataReader> readerTask)
{
    // ...
}
```

/cc @mgravell

### DIFF
--- a/Dapper/Extensions.cs
+++ b/Dapper/Extensions.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Dapper
+{
+    internal static class Extensions
+    {
+        /// <summary>
+        /// Creates a <see cref="Task{TResult}"/> with a less specific generic parameter that perfectly mirrors the
+        /// state of the specified <paramref name="task"/>.
+        /// </summary>
+        internal static Task<TTo> CastResult<TFrom, TTo>(this Task<TFrom> task)
+            where TFrom : TTo
+        {
+            if (task is null) throw new ArgumentNullException(nameof(task));
+
+            if (task.Status == TaskStatus.RanToCompletion)
+                return Task.FromResult((TTo)task.Result);
+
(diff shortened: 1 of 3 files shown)


mgravell:
looks good, thanks; merging

PR #1307, opened 2019-08-12
Support of PostgreSQL arrays: using GetStructDeserializer for arrays

alexeyfadeev:
There are [array types in PostgreSQL](https://www.postgresql.org/docs/11/arrays.html)

Array are supported by Npgsql. e.g. we have a model class:
`public class Model {`
`  public int Id { get; set; }`
`  public int[] Data { get; set; }`
`}`

Now it's possible to do with Dapper:
`string sql = "select n0.* from public.number_array n0 where 11 = any(n0.data)";`
`var result = conn.Query<Model>(sql).ToList();`
Array member is mapped right way.

But if we try to specify this member in SQL and set array as result type:
`string sql = "select n0.data from public.number_array n0 where 11 = any(n0.data)";`
`var result = conn.Query<int[]>(sql).ToList();`
then we get an ArgumentException.

The reason is GetDeserializer choses GetTypeDeserializer instead of GetStructDeserializer for array type. I tried to include arrays in 'known types' on checking and it runs on.


alexeyfadeev:
@mgravell @NickCraver can you take look at this?

alexeyfadeev:
up
@mgravell @NickCraver

NickCraver:
I think this is handled as of #1598 (and will be in the next release). It's enabled by default though, without handlers in play. Can you please confirm if that'll handle your use case too?

PR #948, opened 2018-02-17
fix Insert<T> when T is declared as IEnumerable<T>

bdrupieski:
When inserting, when checking if the type implements IEnumerable<>, we weren't checking that the passed type is itself declared as IEnumerable<>. IEnumerable<> doesn't implement IEnumerable<>; it _is_ IEnumerable<>.

This PR changes adds a check for if the type is itself IEnumerable<> in addition to if it implements IEnumerable<>.

This fixes #930.

PR #1319, opened 2019-08-29
actualizacion desde el original (#1)

### LINKED ISSUE #427: System.Xml.Linq.XElement cannot be used as a parameter value at Dapper
I am using dapper dot net for making Db calls.

now one of my sp having input parameter as Db type XML

Now i am executing sp like

```C#
var param = new DynamicParameters();
param.Add("@TagXML", string.IsNullOrEmpty(student.TagXML)?null:XElement.Parse(student.TagXML));
param.Add("@CustomTagXML", string.IsNullOrEmpty(student.CustomTagXML) ? null : XElement.Parse(student.CustomTagXML));
param.Add("@Return", strRet, direction: ParameterDirection.Output);
param.Add("@RetunErrMsg", strRetMsg, direction: ParameterDirection.Output);
using (var connection = new SqlConnection(ConnectionString))
{
    if (connection.State != ConnectionState.Open)
    {
        connection.Open();
    }
    connection.Execute("My Sp here", param, commandType: CommandType.StoredProcedure);
}
```

This code giving me error:

```
System.NotSupportedException: The member TagXML of type System.Xml.Linq.XElement cannot be used as a parameter value
   at Dapper.SqlMapper.LookupDbType(Type type, String name, Boolean demand, ITypeHandler& handler)
```
Why this error coming? How I can use XML type of inputs in StoredProcedure.?

### LINKED ISSUE #436: SqlBuilder regex expression removes chars outside /** **/ clauses

### LINKED ISSUE #461: TypeHandler is not called for parameterised constructor

### LINKED ISSUE #466: Microsoft.Data.Sqlite and CommandBehavior.SingleResult throws exception
See https://github.com/StackExchange/dapper-dot-net/blob/master/Dapper/SqlMapper.Async.cs#L210

NuGets:
Microsoft.Data.Sqlite 1.0.0-rc1-final
Microsoft.Extensions.PlatformAbstractions 1.0.0-rc1-final

dapper-dot-net master branch

I am trying to retrieve a list of entries from a table.
``` C#
columnMaps = (await connection.QueryAsync<ColumnMap>("SELECT column_id, column_name, ordinal, data_type, variable_type FROM column_map ORDER BY ordinal;").ConfigureAwait(false)).ToList();
```

This exception is thrown.
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at Microsoft.Data.Sqlite.SqliteCommand.<ExecuteDbDataReaderAsync>d__52.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Dapper.SqlMapper.<QueryAsync>d__22`1.MoveNext() in D:\GitHub\dapper-dot-net\Dapper\SqlMapper.Async.cs:line 210

Removing `CommandBehavior.SingleResult` worked, though that is obviously not the universal answer.

### LINKED ISSUE #468: Pass enums as (boxed) basic primitives, not (boxed) as enums
See #467, dotnet/corefx#1613, and the existing `AdoNetEnumValue` test. This is a recurring problem that needs to be fixed once and for all, especially if the "fix" is provider specific (in which case: fixing it in `SqlClient` won't fix it in Sqlite, npgsql, etc)

### LINKED ISSUE #524: why dapper return Int64 ?
```csharp
var sql = @"select count(1) from Products;
                            select * from Products order by ProductId asc limit @ItemsPerPage Offset @Offset";

                using (var multi = _db.Connection.QueryMultiple(sql, new { ItemsPerPage = page.ItemsPerPage, Offset = (page.CurrentPage - 1) * page.ItemsPerPage }))
                {
                    var totalitems = multi.Read<int>().Single();
                    var data = multi.Read<Product>().ToList();
                }
```
Above is wrong. and push a  `System.InvalidCastException'  

  If I change the following way ,it's correct ?
```csharp
var totalitems = multi.Read<Int64>().Single();
```
Is a Bug ?

However, the query does not have this problem.

### LINKED ISSUE #549: DataTable / DapperTable
Why does the file is named `DataTable` but the class is named `DapperTable`?

https://github.com/StackExchange/dapper-dot-net/blob/master/Dapper/SqlMapper.DataTable.cs

### LINKED ISSUE #601: parameters with special characters (æøå)
the "smellsLikeOleDb" regex in version 1.42 (i know it is an old version - but cannot upgrade right now) does not take into account any special characters (æøå) in parameter names.

Well atleast as far as i can see that is the problem. I may have been fixed in later versions but as of now, we cannot upgrade right now - we have tried, and we need to make a focused effort to do it.

Am i wrong in my findings?

regards
Dennis

### LINKED ISSUE #634: Dapper actually does some connection management
Under the **Limitations and Caveats** section in the README.md it says:

> Dapper does not manage your connection's lifecycle, it assumes the connection it gets is open AND has no existing datareaders enumerating (unless MARS is enabled)

Which isn't completely true anymore. It's a small thing but it misled me as a new user of the library and can introduce some annoying and unnecessary code to open and close the connection all the time.

### LINKED ISSUE #992: Reference to defunct Stack Overflow docs on https://stackexchange.github.io/Dapper/
I was looking for examples of Dapper usage and stumbled upon https://stackexchange.github.io/Dapper/. That page has the sentence:

> More examples coming soon on Stack Overflow docs.

Suffice it to say, that's an unusual definition of "soon" since [Docs was discontinued 2017 August 8](https://stackoverflow.com/documentation) :).

joangilabert:
* Tests: add 4 attempts to repro #457

* Change CE package to one that might work

* Fix *all* the CE references

* Don't attempt to load native assemblies on DNX

* Detect app-veyor for MySql/Postgresql

* Standardize on test db for appveyor

* actual/expected in asserts are the wrong way round

All the tests are wrote in the form actual.IsEqualTo(expected). This
made for some odd error messages when a test failed.

* Initial test setup for #461; note that this does not repro, due to running on SQL-Server - needs to be changed to SQLite to repro

* Fix #461 - change FindConstructor to allow parameters that have a type-handler defined

* New feature: PadListExpansions - reduces query plan saturation by padding "in" lists and populating with nulls; opt-in (see remarks on setting)

* 1.50-beta8 deploy

* Fix typos in the comments for PadListExpansions

* Test for SO35470588

* Readme.md: update performance tests link

* Example for SO35554284

* Add SQLITE to test builds; add tests for issues #466 and #467; fix #466 by adding retry-with-fallback flags strategy

* Update to sqlite rc2, because without that nothing works; add extra tests around sqlite oddities; update tests to reflect sqlite rc2 fixes

* Fix #468 - CreateParamInfoGenerator should treat enums as primitives, not box them as enums

* Don't check for null in the basic enum case

* Support get-only C# 6 properties, i.e. `public int Id {get;}`

* prefer an exact-case backing-field match over a wrong-case regular field (essentially: this promotes get-only properties over regular fields)

* Support xml types by default (XmlDocument, XDocument, XElement) - fix #427

* Set the DbType for XmlDocument / XDocument / XElement

* When doing list expansions (6=>10 etc), use the last non-trivial value - do *NOT* use null, as it causes problems with "not in" clauses; add test for the same

* Recreate all the xproj to reflect the current tooling - old versions break VS15; enable XDocument etc in corefx (#if block was too large)

* Heh, AspNet and DNX (only) work in VS2015; DNX and DotNet (only) work in VS15; I guess we need to use DNX for now, then, as the only one that works in both.

* 1.50-beta9

* Unify test credentials for MySQL; add passing test for SO36303462

* Bumping dependencies version numbers

* Revert "Bumping dependencies version numbers"

This reverts commit 70ee9a721ee7466f84bedc63f190f9620e18b24f.

* Readme: Better explanation for multi mapping

* fix outpt paths

* Get building

* PowerShell build scripts

* PowerShell build updates: include .Contrib

* PowerShell build script love

* Add bash build scripts

Note: these are hosed on OS X due to file handle limits, see
https://github.com/dotnet/cli/issues/809 for details.

* Bash build: fix test runner on OS X

Note: no tests pass because there’s no integrated security option (or
widely available SQL server) on OS X at the moment. But the build and
test runners are up and going.

* Update .gitgnore for build pathing

* .Net Core: remove floating versions (use stable ones on Nuget.org) and get this puppy back in action.

Also: it's 2016.

* Add newer feed back **only for the tests projects**, Sqlite for example has bugs in rc1-final fixed in rc2-15948

* Nuget.org feed only, skip the 2 Sqlite tests until RC2+ packages with the fix are out of pre-release.

* Fix assembly name for Contrib tests

* Runtimes fix, per https://github.com/dotnet/cli/issues/2578

* Fix EntityFramework.StrongName build for RC2

* Update to RC2 final build (24027) libraries. Also bump to beta10

* RC2: Remove CoreFX #1612 blocker - fix is released, yay!

https://github.com/dotnet/corefx/issues/1612 is now closed, tests
working.

* Upgrade project.json files to RC2 format.

Note: on a nuspec comparison, the following dependencies aren't there
anymore: Microsoft.CSharp, mscorlib, System, and System.Core aren't
explicitly specified in frameworkAssemblies anymore. The only one of
these I can imagine mattering is Microsoft.CSharp.

* It's time for appveyor.yml

Tooling changes need to live in the repo now so the build doesn't FUBAR
every time we change the script.

* It took several people a while to find this one.

* RC2 Version bump, since we locked ourselves in with semver on beta9

* Update build scripts to pack for us

* JSON RC2 fixes

* Interfaces are back, simplifying code.

Related issue: #520

* net451 test prep (not quite ready - SQL Geo and interop failing)

* XMLDoc fixes for latest methods and missing DataSet in .NET Core

* .xproj: RC2 upgrade

This fixes null refs with the DNX MSBuild targets when building from
Visual Studio.

* Switch version suffix to be a build parameter

Note: build.sh needs love from OS X/*nix later to match

* This .0 addition happens automatically...make it explicit instead.

* git blame game

* Test for #524 (believed to be a failing test; cannot validate or fix until I get the latest tooling installed)

* Fix #524 - `GridReader` should use the same conversion code as the regular reader

* Add InListStringSplitCount; when set, causes `in @foo` parameters to be implemented using string_split

* fix case of generated SQL in string-split

* Dapper.EF.StrongName: should snk for *all* builds!

* Fixed some syntax errors in project.json files.

* .Net Core 1.0 RTM update

* Remove a bunch of unnecessary (and *theoretically* dangerous, in a very unlikely scenario involving custom cultures installed on a box) C# 7 string $"...{...}..." usages

* Simplify the string_split implementation

* 1.50.1 deploy

* Rename DataTable=>DapperTable; fix #549

* Failing test for #552

* (just to prove that #552 still happens even if the only values seen are 0 and 1)

* Fix for #569

* 1.50.2

* Repackage Dapper.EF to fix System.Runtime; deploy as .52; fix xunit refs; add test for #570

* Added tests to complement documentation of pseudo-positional parameters

* Fix SqlBuilder.OrWhere method which was actually yielding an 'AND WHERE' statement.

* Operate on db records instead of files.

* Add CommannDefinition versions of Query(First,FirstOrDefault,Single,SingleOrDefault)Async and a few unit tests for them

* Improve tests

* failing test for #563

* fix broken test to actually include the error!

* Make CommandBehavior more configurable; default to SingleRow=allowed,SingleResult=disallowed - but expose options; this fixes issues like #563 automatically, and allows the performance aspect  of #554 to be fixed via a global setting

* support for international parameter names (regex categories); fix #601

* Add some <remarks> to the new options

* Capitalization

* Update readme.md

Missed second bracket in ID example

* Set true to GridReader.IsConsumed when MultiReadInternal is called

* Non-greedy match for comments regex

Fixes #436, #280
My editor has also normalized a few line endings to LF like 99% of the
file was.

* Fix #634 - dapper now handles connection lifetime and has for ages

* 1.50.3-beta1: looks at #637  re `Guid`

* Toolng cheese move - lock in preview2 for now.

* Fix Contrib test resolution

Note: there's still a 1.0.0 vs 1.1.0 quirk here - digging.

* Add SqlMapper.RemoveTypeMap() method

* Update Readme.md

* Add test DiscriminatedUnionWithMultiMapping

DiscriminatedUnionWithMultiMapping - test for check correct MultiMapping (splitOn) functionality with discriminated row parser.

* added firebird support

added firebird sql adapter to allow record insert and key retrieving

* Revert "1.50.3-beta1: looks at #637  re `Guid`"

This reverts commit 91725fb309a9e5948f760fac420dd805659f0749.

* Update README with documentation for [ExplicitKey] attribute

* Add GetRowParser documentation/example

* Fix missing Firebase Async, issue #695

TODO: All of these need race love in V2, complete re-vamp.

* Rename package to "Dapper"

Related: #683

* Remove note about requiring open connections

* Added Belgrade Sql Client

In the performance benchmark is added Belgrade Sql client data access
library (https://github.com/JocaPC/CLR-Belgrade-SqlClient). This is a
wrapper around ADO.NET classes that executes queries in async manner
using callbacks. "Belgrade.Sql.Client": "0.6.2" is aded in project.json
file, and new flag "BELGRADE" controls should the test be executed.

* migrate release notes to [master]/docs

* Set theme jekyll-theme-cayman

* Add a full stop just so I can tell if the docs migration is working

* Fix Dapper release notes link

* Security!

* Fixed test issue in Simple.Data performance test (#723)

* Added fix for Simple.Data tast

Due to the missing .FirstOrDefault() in Simple.Data test, performance
numbers for Simple.Data were too high (better than hand coded data
reader). Added FirstOrDefault() to actually fetch the data.

* VS 2017 .csproj Migration
Due to the way VS test works (by injecting an executable entry point), the performance tests needed to be factored out. I also organized all of our existing tests better along the way.
Actual code changes to Dapper itself are very minor, only formatting and documentation fixes (which we need many more of).

The build.ps1 script is upated to work, but note that <frameworkAssemblies> is not working in .nuspec inside the packages in the new .csproj system. I consider this to be a blocker. Issue is here: https://github.com/NuGet/Home/issues/4853

* Correct minor typographical error in Readme.md (#732)

Under the `List Support` section, the sentence opens with `Dapper allow you`, which should instead be `Dapper allows you`. This PR will correct this mistake, restoring balance to the universe.

After originally spotting this, the most minor of mistakes, I decided to leave it and move on with my life, since it likely effected nobody. Four hours passed and the thought that it existed unchecked had fully manifested itself within my subconscious and I was no longer able to continue with my existence knowing that I had seen this mistake and not exercised my power to correct it.

Thank you, and sorry.

* Define expected behaviour for value-tuple parameters and returns, and implement

* remove redundant test code (oops)

* Better matching of value-tuple names when wrong number of columns - should return nulls for unmapped

* Add tests for the too few/too many columns scenarios for value-tuples; happy to discuss alternative expectations

* Fix Readme typos (#739)

* Readme.md: Fixed link to test suite (#764)

* VS 2017 Build - Making things work (#766)

* Progress on VS 2017 build tooling
* Finally, a working VS 2017 tooling build

* Update appveyor.yml for 2017 build

* Appveyor: update services

* Let's try this again

* Again!

* More path changes

* Appveyor: SQL 2016

* Appveyor: create databases

* Let's just keep adding YAML until this works.

* Appveyor: Formatting & combine init

* Appveyor: Update Postgres path

* Lots of documentation fixes/additions.

* Fix Microsoft.SqlServer.Types for new .csproj

* Fix type loads for the builds without SQL in the GAC

* Fix SQL spatial type loading

* Skip the MySQL FUBAR test

It's indeed broken on their end. We should consider switching providers,
e.g. https://github.com/mysql-net/MySqlConnector/

* Add build badge to README

* Appveyor: enable MyGet publishing

* Lots of code cleanup, no functional changes

* C# 7 cleanup: pattern matching for TryStringSplit

* Documentation: GridReader

* Documentation: contrib

* Documentation: EntityFramework

* Documentation: Rainbow

* Tests: make analyzers happy because why not

* Documentation and formatting: secondary types

* Documentation and formatting: SqlMapper

* Benchmarks v2: BenchmarkDotNet and friends

This is based on BenchmarkDotNet (while preserving the legacy format
with minor improvements as well - legacy runs much faster). See #666 for
details. Not an ominus number at all.

Note: this code will get a bit simpler with BenchmarkDotNet categories,
see https://github.com/dotnet/BenchmarkDotNet/issues/248 for details.

* Dependency: Dapper doesn't need SqlClient, only Data.Common

* Benchmarks: customization and formatting

More cleanup to do (removing Type column next), but getting to some very
usable output now.

* netstandard2.0 functionality (w/ runtime breaks)

This adds a netstandard2.0 build to Dapper (and StrongName until v2),
adds a netstandard2.0 test lineup, restores most functionality (except
UDTs for SQL, e.g. .UdtTypeName isn't in netstandard2), and also breaks
things not actually implemented or implemented correctly in
netstandard2.0. Pushing this up so we can work through these with the
.NET teams.

Several things are still broken here (and fail tests):
- Parameter decimal values (not filed yet)
- DataTables as parameters
(https://github.com/dotnet/corefx/issues/19708)
- .GetSchemaTable() (not yet filed, implementation is completely missing
from SqlDataReader in CoreFX)

While the compile is clean, the above are runtime issues.
`SqlParameter.UdtTypeName` is a separate issue, that one just wasn't in
`netstandard2.0` at all...so we can't support UDTs there until it is.
See https://github.com/dotnet/corefx/issues/17126 for details.

* Fix MySQL tests and disposal (#774)

* Don't create the connection in Dispose: this avoids creating a SQLConnection simply to Dispose it.
* Run MySQL tests against the open MySqlConnection: 'connection' is a protected variable in the TestBase class that is a SQL Server connection, not a MySQL connection.
* Skip broken MySQL tests: as per issue #295, setting "AllowZeroDateTime=True" in the connection string causes an InvalidCastException in SqlMapper.

* Benchmarks: performance and formatting

* Minor README tweaks. (#776)

Use an HTTPS link where possible.
Fix 404 for performance tests.

* Fix for InsertAsync in PostgresAdapter (#689)

* Fix for PostgresAdapter in SqlMapperExtensions.Async. Copied same line of code from SqlMapperExtensions to fix RuntimeBinderException on InsertAsync
* Cleanup

* Try and use BenchmarkDotNet correctly

OperationsPerInvoke is an informational, not instructional property.
What we actually want is the UnrollFactor. Moved into the job definition
to simplify things.

* fix issue #731: AddDynamicParameters repeated if paramtype is DbStiring should work.

* Added EntityFrameworkCore performance tests (#778)

* Perf: add EntityFramework Core to legacy tests

* Markdown tables (#800)

Markdown tables instead of html
“Can be faster” in correct row

* Enable SqlDataRecord TVPs on corefx. (#801)

* Revert "Enable SqlDataRecord TVPs on corefx. (#801)" (#803)

This reverts commit e7ad0874de8b3680f0453789b10f6e78d436dc40.

* Appveyor: use the VS 2017 preview image

* Avoid allocation for linq in CreateParamInfoGenerator and added some other micro optimizations

* Enable SqlDataRecord TVPs on netstandard1.3. (#802)

* Enable SqlDataRecord TVPs on netstandard1.3.
* Update netstandard2.0 SqlClient references to Preview 2.
* Skip TransactionScope tests for netcoreapp2.0.
- According to https://github.com/dotnet/corefx/issues/12534, ambient transaction enlistment is not included with .net core 2.0.
* Update references in Dapper.StrongName.
* Disable parallel test run to stabilize flaky type handler tests.
* Disable parallel test run by assembly attribute.
* Move Issue461 tests to TypeHandlerTests.
* Group type handler tests in a collection, re-enable parallelization.
- Being part of the same collection, type handler tests will run sequentially. All other tests can run in parallel.
* Move tests relying on query cache & type maps to a separate collection.

* Upgade libs, fix tests

This uses the new non-parallel-at-the-end I added to xUnit in https://github.com/xunit/xunit/pull/1411 This way we can run any global test that are other-test-unfriendly much more easily.

* More test fixes

* Appveypr: update to base 2017 image again

.NET Core 2 SDK is on the base Appveyor image now

* Update semver, add to solution

* Fix for #808

Add dynamic overloads for:
- QueryFirstAsync
- QueryFirstOrDefaultAsync
- QuerySingleAsync
- QuerySingleOrDefaultAsync

This may change a lot in 2.x, but many people asking for them now so here they are. Unit tests also fixed for QuerySingleOrDefaultAsync.

Note: this builds on the netstandard2 branch because of the xUnit changes to go with VS 2017 15.3.

* Fix for #815

New overload so this isn't a breaking change. We can collapse down in v2.0 (including the Impl method).

* Update README to remove advetisements

* Tests: xUnit cleanup

This change:
- Removes the Assert shim
- Uses recommended xUnit methods (improving fail errors)
- Cleans up exception tests
- General test formatting fixes
- Removes the netcoerapp1.0 only testing (oops commit)

* Cleanup

* Fix unit tests

* Update README for new assertions

* net40 is dead - remove #if ASYNC bits

* .editorconfig and formatting

* Misc build analyzer fixups

* Lib updates and prep for netstandard2.0 alpha

* Dapper.Contrib: netstandard2.0

* Fix build script

Removes NuGet workaround for pre-2.0 and lines up with other script
enhancements since.

* MyGet feed!

Adding a dapper specific feed, for much easier pushing to NuGet. When
they implement Ctrl+Click in the UI this won't be necessary...or we
break down and automate it finally. Depends on signing plans later.

* Add literal replacement example to readme

* Additional notes on literal replacements

* words are hard

* Tests for DateTime2 compatibility

This shows the importance of using DbType.DateTime2 in Dynamic Parameters to avoid precision truncation

* Shortened the changes

Combined 2 tests into one

* when inserting, check if type implements IEnumerable<T> to determine if it's a collection (#880)

* when inserting check if type implements IEnumerable<T> to determine if it's a collection
* when inserting, check if type is a collection by comparing its implemented interfaces to unbound generic type IEnumerable<> instead of the name "IEnumerable`1"

* bump semver for next release

* Ignore .vscode/

* 1.50.4 release

* Bump semver for next release to 1.50.5-alpha*

* Fix benchmark description typos (#932)

* fix equal tests (#944)

* fix equal tests

* fix test

* Test and fix for #591

This brings behavior QueryAsync behavior on par with Query (sync) and properly returns nothing when an empty result set is encountered.

* Code formatting fix (tabs to spaces) (#896)

* Make DapperRow implement IReadOnlyDictionary<string, object> (#913)

* Change to GetAll to allow the usage of a nullable type in T when T is an interface… #933 (#936)

* Change to GetAll to allow the usage of a nullable type in T when T is an interface.

* Change to GetAll to allow the usage of a nullable type in T when T is an interface: further change to check if val == null and if so then don't try to assign it as a property value.

* Change to Get to allow the usage of a nullable type in T when T is an interface. (effectively the same change as the previous change to GetAll.)

* Change to GetAsync and GetAllAsync to allow the usage of a nullable type in T when T is  an interface.

* Added in UserWithNullableDob and IUserWithNullableDob.

Added in tests for GetAndGetAllWithNullableValues and  GetAsyncAndGetAllAsyncWithNullableValues.

* Added in .ConfigureAwait(false) in the GetAsync Test.
Added comment to identify which issue the test relates to.

* Added NullableDates tables to the test databases.
Adjusted variable names in GetAndGetAllWithNullableValues and GetAsyncAndGetAllAsyncWithNullableValues.

* Changed IUserWithNullableDob to INullableDate.

* fix Insert<T> when T is declared as IEnumerable<T> (#948)

* Removing unjustified use of AppendFormat in StringBuilder class (#989)

* Remove reference to Stack Overflow docs 

Should resolve #992

* Better error messages for .*Async methods when not using  DbConnection (#986)

* Better error messages for .*Async methods when not using  DbConnection

Perhaps in v2 we move these to DbConnection itself ratherr than IDbConnection for the extensions to eliminate the case, but that's breaking...for now we can at least throw a better error. See #757 for details.

* Tweak errors to allow open IDbConnections and those generating DbCommands

Same checks but a bit more lenient with more specific messaging.

* Tweaking error messages and supporting IDbConnection implementations … (#987)

* Fixed dead link in Readme.md

Replaced a dead link in Readme.md by a link to the WaybackMachine

* Readme.md misc tidying


joangilabert:
Actualizado 29 agosto 2019

PR #1326, opened 2019-09-10
Handle PostgreSQL chars correctly

jjonescz:
PostgreSQL columns with type `"char"` are read by `NpgsqlDataReader` as .NET `char`s. However, current implementation of `SqlMapper.ReadChar` expects `DbDataReader`s to read characters as .NET `string`s.


mgravell:
In principle fine, but can the code be tweaked so that it only does the
extra type test if the test for string fails? The string case seems more
likely in the general case.

mgravell:
awesome, thanks; merging

PR #1330, opened 2019-09-19
Add column attribute for properties.

Sjoo90:
This pull request adds a new attribute for properties to name-mapping for Dapper.Contrib.

### DIFF
--- a/Dapper.Contrib/Readme.md
+++ b/Dapper.Contrib/Readme.md
@@ -128,6 +128,17 @@ Dapper.Contrib makes use of some optional attributes:
         public string Name { get; set; }
     }
     ```
+* `[Column("Columnname")]` - use another column name instead of the name of the property
+
+    ```csharp    
+    public class Employee
+    {
+        public int Id { get; set; }
+        
+        [Column ("FULLNAME")]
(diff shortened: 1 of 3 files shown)


Sjoo90:
@mgravell @NickCraver can you take a look at this? This solution I use in my project to help my columnmapping for Dapper.Contrib.

NickCraver:
Hey Patric! First, thanks for the ideas and contribution here.

Unfortunately, this problem is much larger to solve holistically. We're trying to map that out in #722 on what to do in v3 - it's something we need to step back and design into the core of Dapper to make it work properly.

On this PR specifically, the changes add to an interface which is a breaking change for all downstream consumers (they'll get a `MissingMethodException`). That's not something we can merge in without breaking people...and in the next break we're going to add much more integrated support anyhow. Does that make sense? I'm happy to discuss further, but we'd ideally have time ASAP but the reality is it's going to be a while and with breaks it has to be coordinated...so we want to do such breaks as rarely as possible.

TrevorDArcyEvans:
Similar changes are also needed for _Insert_

NickCraver:

PR #1334, opened 2019-09-21
Use loop variable instead of 0

DamirAinullin:
i instead of 0.


NickCraver:
Oh interesting - do you have an example test where this was failing before that we can add to prevent regression?

DamirAinullin:
Changed exisiting unit test if you don't mind. Added one more level. Without the fix this test will be failed.

DamirAinullin:
Is it OK?

mgravell:
merging, thanks

PR #1336, opened 2019-09-22
Fix arguments for String.Format

DamirAinullin:
Removed pageSize because it is not used in string.format


NickCraver:
I do appreciate the fix, but on this specifically: Massive is a third-party include and should remain as-is for performance testing :)

DamirAinullin:
Ok, I see

PR #1339, opened 2019-09-25
Replaced ContainsKey and [] with TryGetValue

DamirAinullin:
in order to calculate hashCode only one time.


mgravell:
thanks; that's great

PR #1342, opened 2019-09-30
Add IDictionaryEnumerator implementation to DapperRow 

ydemetriades:
Implement IDictionaryEnumerator on DapperRow
Issue fix for [CoreFX #41329](https://github.com/dotnet/corefx/issues/41329)

### DIFF
--- a/Dapper/SqlMapper.DapperRow.cs
+++ b/Dapper/SqlMapper.DapperRow.cs
@@ -10,9 +10,13 @@ namespace Dapper
         private sealed partial class DapperRow
             : IDictionary<string, object>
             , IReadOnlyDictionary<string, object>
+            , IDictionaryEnumerator
         {
             private readonly DapperTable table;
             private object[] values;
+            private int dicEnumerIndex;
+            private bool dicEnumerHasCurrent;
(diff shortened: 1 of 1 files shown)


NickCraver:
First, thanks for the contribution! We were ultimately waiting to see what .NET Core did here (in the hope they fixed the known compatibility issue) - and it looks like they've resolved the issue. I just gave a try to a sample I had open:
```cs
using (var conn = new SqlConnection("Data Source=.;Integrated Security=True"))
{
    var obj = conn.Query<object>("Select 10 Column1, 20 as Column2, 30 as Column3").ToList();
    var inst = JsonSerializer.Serialize(obj);
    inst.Dump();
}
```
Result:
```
[{"Column1":10,"Column2":20,"Column3":30}]
```

Since the root issue is solved, I think we should close this out. The implementation here unfortunately adds overhead (in the form of allocations) to _every single DapperRow instance_ for everyone to work around the original issue. Since that's the case and the original problem is solved another way, I hope we can agree there :)

ydemetriades:
@NickCraver thanks for the review! Since the root issue has been solved, we agree that we can reject thr PR! :)

Issue #1344, opened 2019-10-02
MultiExecute don't need repeat set CommandText


shps951023:
https://github.com/StackExchange/Dapper/blob/247e498f0e52c88d0c58c6f1fa50457bb6ab5888/Dapper/SqlMapper.cs#L546

Because it use same DbCommand.     
It can be removed, there is some efficiency in a large number of executes.  
I create a pull reqeust #1345   

![image](https://user-images.githubusercontent.com/12729184/66030515-1b43d080-e534-11e9-89df-3202604e078d.png)

PR #1345, opened 2019-10-02
MultiExecute don't need repeat set CommandText

shps951023:
Issue : MultiExecute don't need repeat set CommandText #1344

https://github.com/StackExchange/Dapper/blob/247e498f0e52c88d0c58c6f1fa50457bb6ab5888/Dapper/SqlMapper.cs#L546

Because it use same DbCommand.   
Is there a chance to delete it? there is some efficiency in a large number of executes

image


shps951023:
Yes,I found it then  i close it.

PR #1347, opened 2019-10-03
Add installation instructions when using .NET CLI instead of nuget.exe

jawn:
Note I haven't found a way to port -Pre(release) to dotnet add, so I didn't add that instruction.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,6 +10,8 @@ Questions on Stack Overflow should be tagged [`dapper`](https://stackoverflow.co
 
 From NuGet:
 
+### Using Package Manager
+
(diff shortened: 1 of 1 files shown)


mgravell:
Now I'm wondering whether the best thing is just to defer all of them to nuget's site, which shows multiple options...

Still thinking...

Appreciate he input; just trying to think whether we can fix it for more things by having less instead.

jawn:
> Now I'm wondering whether the best thing is just to defer all of them to nuget's site, which shows multiple options...

Yeah, I was thinking the same...

On the other hand, I guesstimate that nuget and dotnet add will be the most frequently used.

Issue #1348, opened 2019-10-07
Dapper.Contrib - Custom map of types without using attributes


jfontsaballs:
Allow keys (implicit and explicit) and calculated properties to be defined externally, without the use of attributes. Also, allow the mapping of properties to columns with different names.

Issue #688, opened 2017-01-28
Dapper v2 planning and discussion


NickCraver:
It's time for a major release and we've just been busy with other libraries. In context though, Dapper is far less work to get there. `netstandard` is already supported and we don't have a long list of desired (breaking) changes to the APIs. Starting a list here to track:

### API changes **(breaking)**:
- [ ] Optional `CancellationToken` params on all Async methods (related: #484)
- [ ] Dapper.Contrib `Insert<T>` and `Insert<T, TKey>` changes for `ISqlAdapter` (discussion in #640, related: #559, #587)
- [ ] Dapper.Contrib - Allow custom `ISqlAdapter` (#497) (note: this locks us down for any future interface additions)
- [ ] `SqlBuilder` needs a *lot* of tests and love (#573, #647)
- [ ] Overload reduction via `CommandFlags` (related: #474)

### Package Changes **(breaking)**:
- [ ] Re-title to "Dapper" & rename Repo (#683)
- [ ] Strong-Name for `Dapper` (deprecating [`Dapper.StrongName`](https://www.nuget.org/packages/Dapper.StrongName/))
- [ ] Strong-Name for `Dapper.Contrib` (#430, #685)

### Missing APIs (non-breaking):
- [ ] `QueryAsync<T>` and related have a few missing from the single/first additions
- [ ] Dapper.Rainbox `netstandard` support (#676)

### PRs to examine:
- [ ] #471 - Allow custom TypeHandlers for standard types (and `.SetValue()`) (related: #206, #259, #303, #319, #433, #571, #607)
- [ ] #458 - Custom type handlers for enums are ignored
- [ ] #512 - Fixed #497 by exposing SqlMapperExtensions.AdapterDictionary as public
- [ ] #558 - Temporary fix (really want to fix the actual issue: #555)
- [ ] #597 - Allow multi-map async queries to be unbuffered (related: #596)
- [ ] #636 - Dapper.Contrib - Support for composite keys in `Get` method

### Cleanup
- [ ] Moving to VS 2017 RC and new `.csproj` system (off Preview 2 `project.json` tooling)
- [ ] Organize folders (after PRs!): `src`, `tests`, `docs`, etc. (and the same in the solution)
- [ ] Release notes/documentation: moving into `/docs` ([GitHub feature post here](https://github.com/blog/2228-simpler-github-pages-publishing), note: this allows PRs to update docs easily too)
- [ ] Global: Interpolated String usage (#651)
[...]

PR #1349, opened 2019-10-07
Add custom mapping of properties to columns

jfontsaballs:
Issue #1348


TroySchmidt:
Isn't this what Dapper.Fluentmap is for?

NickCraver:
Indeed we may support some form of configuration in v3 (I'll try and get #688 updated, but see #722 for current discussion) - a map like this wouldn't be the path though, since any bits will necessitate being in the core `Dapper` library, where Contrib would only be consuming.

NickCraver:
See comments above on #722 please - going to close this one out to cleanup :)

PR #1354, opened 2019-10-17
[add] github issue templates in order to navigate users to forum

mrgchr:
Hi,

This PR contains 3 types of `.gihub/ISSUE_TEMPLATE` for `Bug report`, `Feature request` and `Questions`.
Inspired by https://github.com/serverless/serverless and https://github.com/microsoft/vscode .
(like [this](https://github.com/microsoft/vscode/issues/new/choose) )

I believe github `Issue` should be used for bug tracking and feature requests.
However, some users are opening general question like asking _how to use Dapper_.

I hope that these templates reducing those questions and would provide beneficial results for both of us(developer and users).

My native language is not English, my PR would contain imperfect English.


mgravell:
in principle I'm fine with this; suggestion, though:

```
- Dapper ver. X.Y.Z
- Dapper.Contrib ver. X.Y.Z
- SQL Server 20XX
```

this could be clearer, i.e.

```
- Dapper version X.Y.Z
- .NET runtime; for example: ".NET Framework 4.7.2" or ".NET Core 3"
- addition libraries, if any; for example: "Dapper.Contrib version X.Y.Z"
- target RDBMS, when relevant; for example: "Microsoft SQL Server 2017"
```

mrgchr:
Thank you for your reply.
Updated as 790b95a.

PR #1363, opened 2019-10-24
Use InvariantCulture in FlexibleConvert

sebastienros:
In Sqlite, `DateTime` is stored as `String`, and custom serializers ultimately use `Convert.ChangeType` to do the conversion. The generated code however doesn't set the Invariant culture, like it's done everywhere else, and the conversion can fail if the current culture doesn't match the standard DateTime serialization format from Sqlite.


mgravell:
Looks good, thanks; sorry for the omission

PR #1367, opened 2019-11-03
Added support to SQLAnywhere Database / ODBC

kidmar:
I've added support to SAP Sybase SQL Anywhere Database.
This database uses ":" insted of "@" to identify query parameters.
It also requires an ODBC connection and does not support named parameters.
Because of this I've added two functions to the ISqlAdapter interface:
 - GetParametersPrefix to get the correct prefix for the database
 - GetForcePositionalParameters to force the parameters to be positional in Insert/update statements

I've then added a method to set the default Database provider (to be used insted of hard coded SQL Server) because the ODBC connection does not report provider's name:
 - SetDefaultAdapter

I've tried to make it as simple as possible and it should have zero impact on existing code.

### DIFF
--- a/Dapper.Contrib/Dapper.Contrib.csproj
+++ b/Dapper.Contrib/Dapper.Contrib.csproj
@@ -7,6 +7,8 @@
     <Authors>Sam Saffron;Johan Danforth</Authors>
     <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <PackageId>KubePackages.Dapper.Contrib</PackageId>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


NickCraver:
Hey, thanks for the contribution! Unfortunately, this is very similar to the set of changes in #1053 which is pending a major release for consideration. Changing the interface is breaking and we need to re-engineer how this works in a major release anyhow...it's unlikely this adapter will look like it does today, since the extensibility is a major pain point and has all the flaws and issues these PRs bring to the surface :(

kidmar:
I didn't think that the Interface could break others libraries.
If this is the only problem I can workaround that...
Option 1: Use C# 8.0 feature "Default interface implementation" that was introduced exactly for that reason (I suppose). That way the Interface will provide default implementation that will return "@" and dependant libraries won't get broken

Option 2: I can try to implement the same feature without touching the interface (by adding static functions)

I think option 1 is the best, but please let me know if something like this has a chance to get pulled in

NickCraver:

PR #1392, opened 2020-01-05
Annotate nullability on public APIs

jnm2:
I didn't enable NRTs everywhere, just enough to get the whole public surface annotated.

More work can be incrementally done by adding `#nullable enable` to remaining files and fixing warnings. Once all files in a project have `#nullable enable`, the preprocessor can be replaced with `<Nullable>enable</Nullable>` in the csproj.

/cc @mgravell

### DIFF
--- a/Dapper/CommandDefinition.cs
+++ b/Dapper/CommandDefinition.cs
@@ -4,6 +4,8 @@ using System.Reflection;
 using System.Reflection.Emit;
 using System.Threading;
 
+#nullable enable
+
 namespace Dapper
 {
     /// <summary>
@@ -11,7 +13,7 @@ namespace Dapper
     /// </summary>
     public struct CommandDefinition
     {
-        internal static CommandDefinition ForCallback(object parameters)
+        internal static CommandDefinition ForCallback(object? parameters)
         {
             if (parameters is DynamicParameters)
             {
@@ -36,12 +38,12 @@ namespace Dapper
         /// <summary>
         /// The parameters associated with the command
         /// </summary>
-        public object Parameters { get; }
+        public object? Parameters { get; }
 
         /// <summary>
         /// The active transaction for the command
         /// </summary>
-        public IDbTransaction Transaction { get; }
+        public IDbTransaction? Transaction { get; }
 
         /// <summary>
         /// The effective timeout for the command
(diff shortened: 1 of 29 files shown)


NickCraver:
As noted on Twitter and other places many times: I'm personally against doing this. It's not supported, the tooling doesn't completely work, it makes assumptions about the library unclear, and to boot we're also hosing a lot of other PRs in the process. IMO, whether we do this (and then how) should be discussed in an issue before we get into PRs here.

On the "unclear" front: it makes it easy to make a public API, say it's not nullable, and then the user not check if the thing passed in was null/handle it/etc. That's not right because across library boundaries it's not actually supported for anyone not _also_ using C# 8+. Anyway, let's discuss in an issue please - this PR is a bit of a surprise after other recent discussions on the matter.

jnm2:
@NickCraver Fair points. Hosing PRs is a huge cost. I think the tooling works 100% due to ReferenceAssemblyAnnotator. The specific concern about the "unclear front" doesn't sound right to me; nulls will still be checked even where the public API is non-nullable, and pre-C# 8 consumers will see no difference either way. I'm saying this to try to be helpful, but I'm not invested at all at getting this through.

Since you're personally opposed to it, I don't have a reason to pursue this further at the moment. I'll open an issue if a real-world situation brings me back and no one else has opened one at that point.

NickCraver:
On this piece:
> nulls will still be checked

We need to remain vigilant that they are - it's still on us and that's what is easily deceiving...and can't rely on C# 8 bits on public surfaces is what I mean, or anything they pass down.

I get the wins, I just have a _lot_ or reservations here and don't want to hack around a broken system to make it kinda work (esp in .NET Framework). IMO, let's talk it out in an issue. We're going to need to do this or not across many libraries and the issues are largely shared between them (`DBNull` notwithstanding).

*If* we do this, it should coincide with a PR burn down (which we need a lot of time for) and the v3 large effort with a lot of changes in also at a PR-breaking low point if that makes sense.

jnm2:
I'm not asking for the feature but it sounds like you're asking me to open a discussion issue anyway, so I will. Sorry for surprising you with the PR! It was low-effort on my part and I thought there was a 30% chance it would be helpful in some way. At least you'll have some reference code to grimace at. :D

NickCraver:
I definitely don't want to throw away work and always hate doing so - cool with me grabbing your branch here and pushing it up as a branch in the repo here in case we do this to start with/merge in later?

jnm2:
For sure! And please don't feel like you have to do something with it. I didn't think about the pressure it would create when I sent the code your way or I would have touched base first.

Issue #818, opened 2017-07-21
DefaultTypeMap.MatchNamesWithUnderscores is ignored for ctor parameters


julealgon:
While testing the `MatchNamesWithUnderscores` property with one of my stored procedures, I noticed that if I set it to true to ignore the underscores in my proc columns, it will map them properly into class properties, but still fail for constructor arguments using the same rule.

I'd expect the rule to be consistent across all name-mathing in the type mapper, as the toggle seems to imply it is for "all names" due to it's name, `Match ... NAMES ... WithUnderscores".

I want my model to be immutable, and now I cannot have properly named arguments in my code.

Would it be possible to fix this inconcistency by applying the same logic when trying to match constructor argument names as well?

agjini:
maybe we could submit a PR for this change

andreas-antoniou-cko:
Created a PR: https://github.com/StackExchange/Dapper/pull/1394

andreas-antoniou-cko:
Is there anything more I need to do to get this PR merged? I am unsure if I missed out a step or if this is just due to everyone's time constraints.

chrsin:
@andreas-antoniou-cko I think @NickCraver is saying it can't be added since it would break stuff.
@NickCraver would you accept a new boolean on DefaultTypeMapper that signifies if the stuff should work on constructors aswell?

This issue makes working with postgres, records and dapper a real pain. :'(

https://stackoverflow.com/questions/66624066/dapper-postgres-and-camelcase

senritsu:
fwiw, this is still an issue and is still annoying. +1 for even an opt-in solution via some configuration option, anything would be better than having to alias all columns in result sets.

macco3k:
Is it really so big of a deal to include the linked PR? From what I can see it looks backward compatible with the current behaviour, so this should not be breaking existing code.

mgravell:
Completed via #1947

mgravell:
https://github.com/DapperLib/Dapper/releases/tag/2.0.151

PR #1394, opened 2020-01-06
Added support for MatchNamesWithUnderscores and ctor parameters

andreas-antoniou-cko:
Fix for #818

### DIFF
--- a/Dapper.Tests/MiscTests.cs
+++ b/Dapper.Tests/MiscTests.cs
@@ -1130,5 +1130,26 @@ insert TPTable (Value) values (2), (568)");
             public int Id { get; }
             public string Name { get; } = "abc";
         }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


agjini:
Thanks for your fix. Looking the test it seems to work as expected. I allowed myself to make some comments on your code.

NickCraver:
Looking through this, I'm pretty sure it's a breaking change. It's not a new option, so people will have it on - the case of say `ColumnName` and `Column_Name` existing creates a conflict - anything _differentiated_ by an underscore would be impacted by this. I can think of cases within the Stack Overflow API this would happen, as we use underscores or multiple underscores to differentiate some parameters there. In short: I don't think we can do this.

Perhaps how we map to constructors needs to be pluggable, like thinking in #722 for columns?

Issue #1395, opened 2020-01-07
Dapper.Contrib inconsistent id field quoting with PostgreSQL


ryan-todd:
With PostgreSQL, the manner in which a key property name is quoted is inconsistent between the `Update`/`Delete` and `Get`/`Insert` methods, causing either `Update` and `Delete`, or `Get` and `Insert` to potentially fail depending on database configuration.

`Update` and `Delete` use `PostgresAdapter.AppendColumnNameEqualsValue` when appending the key field to sql, which puts `"` around the field as expected:
https://github.com/StackExchange/Dapper/blob/981cfe4ddfde421b9d572d8f5c2cda259be20a65/Dapper.Contrib/SqlMapperExtensions.cs#L1039

However, `Get` and `Insert` fail to put quotes around the key field:
- `Get` inserts the key field into the sql string directly ([link](https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.cs#L179))
- `Insert` calls `PostgresAdapter.Insert` which fails to include quotes as part of the `RETURNING` sql ([link](https://github.com/StackExchange/Dapper/blob/master/Dapper.Contrib/SqlMapperExtensions.cs#L1004))

The result of this is if the id property is not all lower-case (both in the .NET class and in Postgres), then either `Get` and `Insert`, or `Update` and `Delete` will fail, because Postgres is case sensitive and will automatically transform any column identifier to lower-case unless it is wrapped in double quotes. E.g.:
- if the key field is `ID` in .NET and `ID` in Postgres, `Get` and `Insert` will fail because Postgres will automatically transform `ID` to `id`, as it is not enclosed in double quotes
- if the key field is `ID` in .NET and `id` in Postgres, `Update` and `Delete` will fail because Postgres won't automatically transform `"ID"` to lower-case because it is enclosed in double quotes

A solution for this would be:
- `Get` should append the ID field with an adapter-specific call much like `Update` and `Delete` call `ISqlAdapter.AppendColumnNameEqualsValue`
- `PostgresAdapter.Insert` should use double quotes when appending the key field to the sql StringBuilder

icambron:
I got hit by this too. I think the right fix for `Insert` is to use `AppendColumnName()`, though

https://github.com/StackExchange/Dapper/blob/981cfe4ddfde421b9d572d8f5c2cda259be20a65/Dapper.Contrib/SqlMapperExtensions.cs#L1027

[...]

PR #1396, opened 2020-01-07
Issue #1395 - Dapper.Contrib inconsistent id field quoting with PostgreSQL

ryan-todd:


Dapper.Contrib/SqlMapperExtensions.cs
```diff
@@ -1001,7 +1007,7 @@ public int Insert(IDbConnection connection, IDbTransaction transaction, int? com
                 if (!first)
                     sb.Append(", ");
                 first = false;
-                sb.Append(property.Name);
+                sb.Append($"\"{property.Name}\"");
```
icambron: isn't this the same as `adapter.AppendColumnName(sb, property.Name)`?

ryan-todd: Yep, not sure how I missed this one!

rclarke2050:
can someone tell me when this is going in please?
I am reaching this issue now with lambda .NetCore on AWS Aurora.  Aurora names the fields in lowercase by default, even if i create the tables with Casing.

In the .net classes i use, they're in PascalCase, and for .Net SQL connections it doesnt matter. but for NpgsqlConnection, it totally doesnt like it.

Please can we get this PR finished?

ryan-todd:
Not sure on the process as this is my first real contribution to a public repo - what's the next step to get this going?

mgravell:
Hi; sorry, we (the green-lighters) are swamped; sorry if we didn't get to this yet. But; let's see what we can do here!

IMO the thing that is missing here is a test; something that demonstrates the problem being tackled and fixed, so we can have confidence in what it is doing. Is it possible to add a test for this? there is a `PostgresqlTests` class (or you can create a new class that extends `: TestBase<PostgresProvider>`)

ryan-todd:
Hi mgravell, that's ok! I'll have a look at tests when I next get a chance, or if anyone else thinks they can add this, feel free.

NickCraver:

Issue #1399, opened 2020-01-14
Selecting a nullable tuple always returns null


JulianRooze:
Currently, this code always returns null:

```csharp
connection.QuerySingleOrDefault<(int id, string name)?>("select 1, 'Bob'");
```

That is because `IsValueTuple` does not handle the case where it's a `Nullable<ValueTuple<>>` and because of that the special logic for value tuple deserializer generation (`GenerateValueTupleDeserializer`) is not called. Then presumably it follows the normal deserializer logic for structs and it would expect the above query to return columns called Item1 and Item2.

JulianRooze:
I've submitted a PR with fixes and tests for this issue: https://github.com/StackExchange/Dapper/pull/1400

NickCraver:
Fix merged in, thank you!

PR #1400, opened 2020-01-14
Fix handling of selecting nullable value tuples (fixes issue #1399) 

JulianRooze:
`SingleOrDefault<(int, int)?>(query)` always returned null because the IsValueTuple did not handle nullable types and thus the `GenerateValueTupleDeserializer` method was never called.

I've added test for the case when the query does not return something (the tuple should be null) and when it does return something. Additionally added tests for the edge-case where you select 8 or 15 element tuples to make sure nothing broke there.

### DIFF
--- a/Dapper.Tests/TupleTests.cs
+++ b/Dapper.Tests/TupleTests.cs
@@ -41,6 +41,22 @@ namespace Dapper.Tests
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


benbryant0:
Should also fix https://github.com/StackExchange/Dapper/issues/1108

Issue #1372, opened 2019-11-11
Problem with TypeDescriptor with NULL columns generated from Query<dynamic>


softwarekamal:
Hello **Mark**, 
Wish you are good, sorry for disturbing you atm.

But  I found something I think its not normal behavior of Query<dynamic>() which you was implement   SqlMapper.DapperRow.Descriptor for it before. What a great work you done.

Sometimes in database table the full column of data can be NULL, for Example "BranchID" Columns of type int.

When attach Grid DataSource with Query<T>("SELECT * FROM Table") strongly typed class. If all columns are NULL there's no problem. And Grid Editors (Cells) will behavior as integral numbers so I can't post a string inside cell of BranchID Column.

Now back to dynamic version wish is good for getting data at run-time.  
1.	If the Column BranchID contains at least any integral number. The Grid Cell Editor will behavior as int ,  object{int} from Descriptor I think.
2.	If the Column BranchID doesn't contains any integral number and no default values for that column and Column only have NULL for whole rows.  It will be object of null
And here's a problem came with Grid Cell Editor which it will behave as object not int, and now it accepts any random words "9921RPG"

I suspicious something maybe modified at following lines of code…

https://github.com/StackExchange/Dapper/blob/master/Dapper/SqlMapper.DapperRow.Descriptor.cs
![a](https://user-images.githubusercontent.com/33269552/68604499-4d413e80-04b3-11ea-8bfc-cccdaccb9755.jpg)

softwarekamal:
**Attachment**
[TypeDescriptor to fix types of null columns.zip](https://github.com/StackExchange/Dapper/files/3832334/TypeDescriptor.to.fix.types.of.null.columns.zip)

mgravell:
Is the zip your proposed you changes? If so, a PR would be preferable.

softwarekamal:
Hello Mark, thanks for your reply
I create it here 
https://github.com/StackExchange/Dapper/pull/1373

rhubley:
I am also in need of the proper types from fields in a Dapper Row

I created #1409 to fix the git issue that was in 1373.

PR #1373, opened 2019-11-11
Can you Support Types on Descriptor API for preventing null objects in Grid Cells.

softwarekamal:
From Issue Here : https://github.com/StackExchange/Dapper/issues/1372

Problem that Descriptor API which works with `Query<dynamic>` and `DapperRow`, with null columns, it make Grid Cell Editors accept random keywords rather than only input from the Column `Type`, So can you please look at following fix:

**Modifications** 

**SqlMapper.DapperRow.Descriptor.cs**

```
                internal static PropertyDescriptorCollection GetProperties(DapperTable table, IDictionary<string,object> row = null)
                {
                    
                    string[] names = table?.FieldNames;
                    Type[] types = table?.FieldTypes;
                    if (names == null || names.Length == 0) return PropertyDescriptorCollection.Empty;
                    var arr = new PropertyDescriptor[names.Length];
                    for (int i = 0; i < arr.Length; i++)
                    {
                        var type = row != null && row.TryGetValue(names[i], out var value) && value != null
                            ? value.GetType() : types[i];
                        arr[i] = new RowBoundPropertyDescriptor(type, names[i], i);
                    }
                    return new PropertyDescriptorCollection(arr, true);
                }
```

**SqlMapper.cs**

```
  internal static Func<IDataReader, object> GetDapperRowDeserializer(IDataRecord reader, int startBound, int length, bool returnNullIfFirstMissing)
        {
            var fieldCount = reader.FieldCount;
            if (length == -1)
            {
                length = fieldCount - startBound;
            }

            if (fieldCount <= startBound)
            {
                throw MultiMapException(reader);
            }

            var effectiveFieldCount = Math.Min(fieldCount - startBound, length);

            DapperTable table = null;

            return
                r =>
                {
                    if (table == null)
                    {
                        string[] names = new string[effectiveFieldCount];
                        Type[] types = new Type[effectiveFieldCount];

                        for (int i = 0; i < effectiveFieldCount; i++)
                        {
                            names[i] = r.GetName(i + startBound);
                            types[i] = r.GetFieldType(i + startBound);
                        }

                        table = new DapperTable(names,types);
                    }

                    var values = new object[effectiveFieldCount];

                    if (returnNullIfFirstMissing)
                    {
                        values[0] = r.GetValue(startBound);
                        if (values[0] is DBNull)
                        {
                            return null;
                        }
                    }

                    if (startBound == 0)
                    {
                        for (int i = 0; i < values.Length; i++)
                        {
[...]

PR #1409, opened 2020-01-29
Support Types on Descriptor API

rhubley:
this is an implementation of
Can you Support Types on Descriptor API for preventing null objects in Grid Cells. #1373

there seems to be a git issue with that pull request. I hope you don't mind @softwarekamal

### DIFF
--- a/Dapper/SqlMapper.DapperRow.Descriptor.cs
+++ b/Dapper/SqlMapper.DapperRow.Descriptor.cs
@@ -63,12 +63,13 @@ namespace Dapper
                 internal static PropertyDescriptorCollection GetProperties(DapperTable table, IDictionary<string,object> row = null)
                 {
                     string[] names = table?.FieldNames;
+                    Type[] types = table?.FieldTypes;
                     if (names == null || names.Length == 0) return PropertyDescriptorCollection.Empty;
                     var arr = new PropertyDescriptor[names.Length];
                     for (int i = 0; i < arr.Length; i++)
                     {
                         var type = row != null && row.TryGetValue(names[i], out var value) && value != null
-                            ? value.GetType() : typeof(object);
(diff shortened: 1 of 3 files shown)


softwarekamal:
Exactly !, Am poor at GitHub so Thanks for uploading it. I wish they move it to a Branch.
It's very important for Dynamic and GridViews NULL values to accept only the specified Type. Not random value.

softwarekamal:
Hello **Marc**, Does this issue has been fixed :)

NickCraver:
Note: this would fix #1372

softwarekamal:
Hello Nick, nice to meet you :)
Yes I uploaded 3 files in (Files Changed) tab above. Its really awesome to commit it during its fix the type problem when working with Query<dynamic> with a GridViews, Bindable Controls. Which it was gave a bad type. 
Thank you again **nick** and **mark** thank you both for your great project and efforts :)

NickCraver:
Trying to catch up on PRs here - can we please add a test illustrating the usage here? This ensures future changes don't regress the fix :)

softwarekamal:
Hello Nick, thank you for reply.
You can test as following in WinForms app
```
BindingList<dynamic> dataSource = new BindingList(Query<dynamic>("SELECT * FROM Item"));
dataGridView1.DataSource = dataSource;
```

1. Try with empty SQL table no rows. (You will find GridView Cells Integer, DateTime one accept string, any data-types)
2. Now add single Row from SQL itself. then Test GridView again. (You will find it accepts a correct values)
3. After you commit changes all it do is just pass a Field Type while pass to DapperRow, DapperTable things. see commit please.

_As for test I think its 100%_ not effects any related Dapper Core. Its all about just specify the type of real row cell of DapperRow. in TypeDescriptor api that Mark Gravel do it before.

**Thank you much again.**

Issue #1366, opened 2019-10-31
Support for EF Core spatial data


lettucebo:
Now EF Core support for spatial data begin from EF Core 2.2 and using NetTopologySuite as spatial data type base

Is there any plan for support NetTopologySuite Geometry data type?

mgravell:
Great question. Now that this exists, we should investigate.

lettucebo:
I have implement one version for Entity Framework Core
https://github.com/lettucebo/Dapper/tree/master/Dapper.EntityFrameworkCore

While I'm going to try is this working, I stuck at a strange point saying it can't find `Microsoft.SqlServer.Types` but I quite sure I never touch Microsoft.SqlServer.Types no matter from Dapper or my test project

And the exception message seems saying it come From `System.Data.SqlClient` : 
```
FileNotFoundException: Could not load file or assembly 'Microsoft.SqlServer.Types, Version=10.0.0.0, Culture=neutral, PublicKeyToken=89845dcd8080cc91'. The system cannot find the file specified.
```
![Annotation 2019-11-01 002339](https://user-images.githubusercontent.com/891383/67966608-28410600-fc3f-11e9-9588-3cf707c90f06.png)

Need some help insight
https://github.com/dotnet/SqlClient/issues/30#issuecomment-459540715

michaelgregson:
Thank you @lettucebo; great stuff.

Same issue here, setting the column type to Geography in MSSQL seems to instruct the System.Data.SqlClient to use SqlGeography type from Microsoft.SqlServer.Types only when reading.

Adding a reference to Microsoft.SqlServer.Types solves the error.

lettucebo:
@michaelgregson 

Yes, but I want to run my ASP.NET Core web application at Linux

Since Microsoft.SqlServer.Types is base on .NET Framework, so it might not work at Linux platform

steveoh:
I just bumped into this issue myself...
```
The member shape of type NetTopologySuite.Geometries.Geometry cannot be used as a parameter value
```
This would be wonderful if it worked...

carlin-q-scott:
I found this very simple mapper works with Npgsql:
```C#
            SqlMapper.AddTypeHandler(typeof(Point), new SqlMapper.UdtTypeHandler("geography"));
```

PR #1412, opened 2020-02-13
add Dapper.EntityFrameworkCore for NetTopologySuite handler

lettucebo:
EF Core support for spatial data begin from EF Core 2.2 and using NetTopologySuite as spatial data type base

Adding TypeHandler to resolve NetTopologySuite

Base on https://gist.github.com/bricelam/7eca234674c3ca4150872f899af37611

Sample Code: https://gist.github.com/lettucebo/12076a87afdc0c744c25fa9da4e7c86a

> There is a small catch, when selecting spatial data type from SQL Server, have to Serialize the data in the query
> Example: `Feature.Serialize() AS Feature`

Reference: #1366

### DIFF
--- a/Dapper.EntityFrameworkCore/NetTopologySuiteGeometryHandler.cs
+++ b/Dapper.EntityFrameworkCore/NetTopologySuiteGeometryHandler.cs
@@ -1,0 +1,52 @@
+﻿using System.Data;
+using System.Data.SqlClient;
+using NetTopologySuite.Geometries;
+using NetTopologySuite.IO;
+
+namespace Dapper.EntityFrameworkCore
+{
+    /// <summary>
+    /// Type-handler for the Geometry spatial type.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class NetTopologySuiteGeometryHandler<T> : SqlMapper.TypeHandler<T>
+        where T : Geometry
+    {
+        readonly bool _geography;
+        readonly SqlServerBytesWriter _writer;
+        readonly SqlServerBytesReader _reader;
+
+        /// <summary>
+        /// Create a new handler instance.
+        /// </summary>
+        /// <param name="geography"></param>
+        public NetTopologySuiteGeometryHandler(bool geography = false)
+        {
+            _geography = geography;
+            _writer = new SqlServerBytesWriter { IsGeography = geography };
+            _reader = new SqlServerBytesReader { IsGeography = geography };
+        }
+
+        /// <summary>
+        /// Parse a database value back to a typed value.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public override T Parse(object value)
+            => (T)_reader.Read((byte[])value);
+
+        /// <summary>
+        /// Assign the value of a parameter before a command executes.
+        /// </summary>
+        /// <param name="parameter"></param>
+        /// <param name="value"></param>
+        public override void SetValue(IDbDataParameter parameter, T value)
+        {
+            parameter.Value = _writer.Write(value);
+
+            ((SqlParameter)parameter).SqlDbType = SqlDbType.Udt;
+            ((SqlParameter)parameter).UdtTypeName = _geography ? "geography" : "geometry";
+        }
+    }
+
+}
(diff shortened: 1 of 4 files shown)


Dapper.EntityFrameworkCore/NetTopologySuiteGeometryHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        readonly bool _geography;
+        readonly SqlServerBytesWriter _writer;
+        readonly SqlServerBytesReader _reader;
+
+        /// <summary>
+        /// Create a new handler instance.
+        /// </summary>
+        /// <param name="geography"></param>
+        public NetTopologySuiteGeometryHandler(bool geography = false)
```
mgravell: something about this bool parameter is making me itch a little; I wonder if there's an alternative - for example, taking an enum or even the `string` (udt-type-name)? or maybe a common abstract base type `NetTopologySuiteHandler<T>` that has an `abstract` type name property, and two concrete sub-types (that could presumably be `internal`): `NetTopologySuiteGeometryHandler<T>` and `NetTopologySuiteGeographyHandler<T>`?

mgravell:
This is looking really interesting! However, before we could merge it, I'd be very keen to see tests (presumably based on the example usage) that show it actually working.

Also: can I confirm for the record that you're free, able and willing to contribute the code to the project? (we *really* need to get a CLA bot automated here!)

Dapper.EntityFrameworkCore/NetTopologySuiteGeometryHandler.cs
lettucebo: I will change to `NetTopologySuiteGeometryHandler<T>` and `NetTopologySuiteGeographyHandler<T>`

lettucebo:
I'm happy to contribute the code to the project.

But I still need to mention that, part of code is coming from a public Gist: https://gist.github.com/bricelam/7eca234674c3ca4150872f899af37611

And I will write Tests

mgravell:
pining @bricelam - can I confirm that there's no licensing / copyright issues here?

bricelam:
I, Brice Lambson, hereby release the original wonky sample code under the terms of [The Unlicense](https://choosealicense.com/licenses/unlicense/). Go nuts.

bricelam:
While I have you (and on a related note), would it be possible to make [these type handlers for Microsoft.Data.Sqlite](https://docs.microsoft.com/dotnet/standard/data/sqlite/dapper-limitations#data-types) unnecessary?

lettucebo:
@mgravell - I have separate the handler and write tests for both handler

NickCraver:
A few top level thoughts here after giving it an overview:
1. This doesn't seem to do anything directly to EF Core, it seems like it's really `Dapper.NetTopologySuite` (could be an external lib to the repo in that regard, but different discussion).
2. The tests are great! But they include 9MB of binaries - we shouldn't do that in the repo. Additionally it looks like the tests have parallelism conflicts (assuming this since it's explicitly disabled). That's build time minutes and should be fixed if it's to live in this repo.
3. Given how we're going - I'm hesitant to add this library as-is because it can't be signed and depend on `Dapper` (it'd have to be `Dapper.StrongName`)...but we're going to sign in v3 so that's another lib break.
4. I _would_ say it could share the EF test project on a conditional compile, but given EF 6.3 supports netstandard that's a conflict...and see 1 about this anyway, I'd propose this isn't really an EF Core lib, especially since it pulls in a dependency that's very much another party in play here.

NickCraver:
@bricelam Would you mind opening an issue on that specifically? The problem we've always run into is we can't easily do that due to how different database store or don't store certain portions, e.g. EF IIRC chooses and Endianness for `Guid` <-> `byte[]` but it's arbitrary and unless everyone agrees we're kinda SOL there (and ADO.NET doesn't have a place for this consistent, today...since the providers may or may not implement the per-type methods on reader and such).

That being said - we're thinking about designs for Dapper v3 and very expansive changes like `IDapper`, the ability to inject, settings on that instance, more configurability _on that instance_ (and a shared global/static default one). In that world, a user could decide (and we could provide handy defaults) - it'd be much more powerful than our attempt at detection based on the `DbConnection` derivative type today (which also just plain breaks when wrapped). I think of these conversions not to ultimately be dissimilar from whether the database supports native arrays and such.

Hope that makes sense - I don't mean "please open an issue" as dismissive, it's finally time to figure these bits out and I think with the v3 designs we've been spitballing thus far there's a real chance of accommodating it, where the current layout doesn't allow that really.

Issue #581, opened 2016-08-05
Insert with Nullable<int> Id gives "Invalid cast" error


justinjstark:
When the model is defined with an Id of nullable int and the SQL Server database is an `INT IDENTITY(1,1)`, the result of a Dapper.Contrib `INSERT` is:

```
Invalid cast from 'System.Int32' to 'System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]'.
```

Reproduction:

```sql
CREATE TABLE Car
(
    Id INT IDENTITY(1,1),
    Model VARCHAR(50)
)
```

```c#
public class Car
{
    int? Id { get; private set; }
    string Model { get; private set; }
}
```

```c#
var id = sqlConnection.Insert(new Car { Model = "Mazda" });
```

This was working as Dapper 1.42 but is broken as of Dapper 1.50.

alejandroMunozMulet:
Same Issue here dapper.Contrip Insert is broken in 1.50, I had to downgrade to 1.43 to make it work.

cmarrades:
+1. 

Had to downgrade version to get it working.

vilix13:
For me downgrade dont help. In mssql table I have `bigint type`, in class `long?`. Try with 1.5 dapper and 1.2.
But if I change `long?` to `int?` I have another strange error:
```
An exception of type 'System.Data.SqlClient.SqlException' occurred in System.Data.dll but was not handled in user code

Additional information: The variable name '@Id' has already been declared. Variable names must be unique within a query batch or stored procedure.

Must declare the scalar variable "@Surname".
```

Code:
```
var empl2 = new Employee
            {
                Active = empl.Active,
                Birthdate = empl.Birthdate,
                CreatedAt = empl.CreatedAt,
                Deleted = empl.Deleted,
                Firstname = empl.Firstname,
                Middlename = empl.Middlename,
                Sex = empl.Sex,
                Surname = "fdfdf",
                UpdatedAt = empl.UpdatedAt,
                ValidUntil = empl.ValidUntil
            };
            var id = db.Insert<Employee>(empl2);
```

```
public class Employee : Entity
    {
        public int? Id { get; set; }

        public string Surname { get; set; }

        public string Firstname { get; set; }

        public string Middlename { get; set; }

        public string Sex { get; set; }
 
        public DateTime Birthdate { get; set; }

        public DateTime ValidUntil { get; set; }
[...]

PR #1438, opened 2020-04-16
Dapper.Contrib Nullable ID (#581)

jurakovic:
Added support for Nullable ID property in Dapper.Contrib
Fix for issue #581


NickCraver:
Can you please provider any tests illustrating the use case this fixes?

jurakovic:
I have added InsertUserNullableId test
For some reason i cannot run tests in VS. No errors are shown so I don't know where is the problem  :/
But I have tested it somehow outside of solution and it passes :)

gilb195:
I've been experiencing this issue as well since I'm using a nullable int for the ID in my models. Thank you @jurakovic for this PR - would love to see this merged and fixed soon.

NickCraver:

PR #1443, opened 2020-04-28
Replaced constant with index in loop

DamirAinullin:
Replaced constant with index in loop


mgravell:
I could have sworn this had been fixed once already, but: looks good, thanks

DamirAinullin:
@mgravell It was done here: 
https://github.com/StackExchange/Dapper/pull/1334
Why it was not merged?

PR #1449, opened 2020-05-01
Set initial capacity for collections

DamirAinullin:
for perfomance reason

### DIFF
--- a/Dapper.Tests.Performance/Massive/Massive.cs
+++ b/Dapper.Tests.Performance/Massive/Massive.cs
@@ -209,7 +209,7 @@ namespace Massive
         /// </summary>
         public virtual List<DbCommand> BuildCommands(params object[] things)
         {
-            var commands = new List<DbCommand>(things.Length);
+            var commands = new List<DbCommand>();
             foreach (var item in things)
             {
                 if (HasPrimaryKey(item))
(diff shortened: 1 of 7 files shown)


NickCraver:
It seems like we're running system-wide analyzers here for cleanup - and while I do appreciate it, it'll be less overhead for us to do that outside the PR queue where we're spending more time reviewing changes here than performing them. Let's take a look at getting this in, then we can take the others :)

Dapper.Tests.Performance/Massive/Massive.cs
```diff
@@ -209,7 +209,7 @@ public virtual DbConnection OpenConnection()
         /// </summary>
         public virtual List<DbCommand> BuildCommands(params object[] things)
         {
-            var commands = new List<DbCommand>();
+            var commands = new List<DbCommand>(things.Length);
```
NickCraver, line 212:
```suggestion
            var commands = new List<DbCommand>();
```
This is modifying Massive - a source package. It shouldn't be touched.

Dapper.Tests.Performance/Massive/Massive.cs
```diff
@@ -323,7 +323,6 @@ public virtual DbCommand CreateUpdateCommand(object o, object key)
             var settings = (IDictionary<string, object>)expando;
             var sbKeys = new StringBuilder();
             const string stub = "UPDATE {0} SET {1} WHERE {2} = @{3}";
```
NickCraver, line 325:
```suggestion
            const string stub = "UPDATE {0} SET {1} WHERE {2} = @{3}";
            var args = new List<object>();
```
This is modifying Massive - a source package. It shouldn't be touched.

PR #1460, opened 2020-05-10
Improve work with StringBuilder to decrease memory traffic and to imp…

DamirAinullin:
…rove perfomance

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -163,7 +163,7 @@ namespace Dapper.Contrib.Extensions
             }
 
             var name = GetTableName(type);
-            var sbColumnList = new StringBuilder(null);
+            var sbColumnList = new StringBuilder();
             var allProperties = TypePropertiesCache(type);
             var keyProperties = KeyPropertiesCache(type).ToList();
             var computedProperties = ComputedPropertiesCache(type);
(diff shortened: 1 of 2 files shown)


NickCraver:
A general issue here is it'll break many existing PRs, but on the other hand: they're all breaking changes to column quoting/casing that can't be merged.

Overall for v3 I'm thinking of drastically changing the interface, maybe making it a base class where `Insert<T>` is what you implement...rather than the adapter. The adapters are just so brittle as so many issues have shown, and what people want to change on casing, etc. If they could be overridden directly, more easily scoped to each provider, PRs to fix one wouldn't break all others, etc. I need to write up ideas for .Contrib v3 (and general), but...yeah, minor brain dump there.

I'm fine with this as-is, given we can't merge the conflicting PRs anyhow. @mgravell thoughts?

NickCraver:
@DamirAinullin dammmit, I intended to merge this _before_ the Dapper.Contrib split to https://github.com/DapperLib/Dapper.Contrib - any chance you're willing to open a PR (with the same branch) there so we can assign credit? I based that repo on the full chain so it should be a clean merge, but I'll take any conflicts if it's not - my fault for not getting this on the list before the split (where we were going to eat all those PR breaks I mentioned above anyway).

NickCraver:
Going to close this one here, but would love to merge it into the destination repo still if willing!

PR #1467, opened 2020-05-20
Add support for Array return value (i.e.: QueryAsync<int[]>))

DarkWanderer:
This change allows to execute queries which return arrays as first-class values, i.e.: `QueryAsync<byte[]>(sql)`.

If the type was not guessed exactly (i.e.: `int[]` vs `byte[]`), a conversion is performed


NickCraver:
This looks good, thanks! Looks like it just needs a master merge up - would you like to handle or me to push to the branch?

DarkWanderer:
Recreated as #1598

Issue #654, opened 2016-12-02
FetchSize optimization possibility for Oracle


salixzs:
There are data retrieval optimization possibility for Oracle database Command and DataReader to specify DataFetch property for any of those (DataReader inherits from Command) to optimize (make faster) amount of data fetched from DataReader in one database trip. 
There are numerous examples on how it can be achieved and set when using pure ADO.NET, but in essence one needs to have to set this parameter for DataReader just before start to read from it.
Some examples are here:
https://kiquenet.wordpress.com/2016/06/18/performance-dataaccess-oracle-and-odp-net/
https://www.codeproject.com/tips/428152/oracledatareader-fetchsize-property
but you can find in Google more examples and explanations with keywords "ODP.NET FetchSize".

```c#
OracleConnection myConnection = new OracleConnection(myConnectionString);
OracleCommand myCommand = new OracleCommand(mySelectQuery, myConnection);
myConnection.Open();
using (OracleDataReader reader = myCommand.ExecuteReader(CommandBehavior.CloseConnection))
{
    // here goes the trick
    // lets get 1000 rows on each round trip
    reader.FetchSize = reader.RowSize * 1000;

    while (reader.Read())
    {
        // reads the records normally
    }
}// close and dispose stuff here
```

It would be nice to somehow set FetchSize to ```reader.RowSize * 1``` for QueryFirst* QuerySingle and ```reader.RowSize * 100``` for Query (returning multiple rows).

Other option would be to give possibility to specify it via some hooking.

Do you see it is possible at all?

mgravell:
It isn't inconceivable; we already do some per-type hacks for some other
things. Can I ask what kind of performance difference were talking about
here in real world scenarios?

salixzs:
Here are some links to blog articles people measuring differences in data retrieval speeds if FetchSize is adjusted:
http://metekarar.blogspot.com/2013/04/performance-improvement-for-odpnet.html
http://grow-n-shine.blogspot.com/2011/05/optimize-net-access-with-oracle.html
http://guyharrison.squarespace.com/blog/2009/10/19/oracle-performance-programming-net.html

Here is ORACLE Video from Oracle Virtual Technology Summit 2015
https://www.youtube.com/watch?v=ozMPGmsKcoA
Start around 10:14, demo on it around 11:08

Here is marginally related article of data retrieval over different networks and how FetchSize affects it:
http://markbairden.blogspot.com/2012/03/database-performance-measuring-effects.html

[...]

PR #1505, opened 2020-07-24
Extension hook for DataReader

epaulsen:
Added an `Action<IDataReader>` property to CommandDefinition, and if set, that action is invoked after ExecuteDataReader is called on the command object.  

This will enable third-party users to access datareader object after creation, and in the Oracle use-case, set FetchSize property, which can have significant impact on performance.  See issue #654 for details.

Intended usage:
```csharp
var result = connection.Query<FooBar>(new CommandDefinition(
                commandText: "SELECT FOO,BAR FROM FOOBAR"
                dataReaderSetup: reader =>
                {
                    if (reader is OracleDataReader oracleReader)
                    {
                        oracleReader.FetchSize = oracleReader.RowSize * 38;
                    }                    
                }));
```

The downside to doing it this way is that the datareader object with all its properties and methods are exposed to api consumer, and any calls to Read() or other methods that move the datareader forward will result in corrupted results returned from Dapper query methods.  However, usage of this should be the callers responsibility.  

By doing it this way, Dapper does not need to know anything about Oracle or other provider specifics.

### DIFF
--- a/Dapper.Tests/MiscTests.cs
+++ b/Dapper.Tests/MiscTests.cs
@@ -167,6 +167,74 @@ namespace Dapper.Tests
             Assert.Equal("Sequence contains more than one element", ex.Message);
         }
 
+        private class DataReaderConfiguratorTest
+        {
+            // Returns true if the DataReaderSetup extension point is invoked after the datareader was created, false if otherwise
+            public bool Test<TReturnType>(IDbConnection conn, Func<CommandDefinition, TReturnType> func)
+            {
+                bool callbackInvoked = false;
+                conn.Execute("create table #dog(Age int, Name nvarchar(max)) insert #dog values(1, 'Alf')");
+                try
+                {
+                    var result = func(new CommandDefinition("select * from #dog", dataReaderSetup: r => { callbackInvoked = true; }));
+                    (result as IDisposable)?.Dispose();
+                }
+                finally
+                {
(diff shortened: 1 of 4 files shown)


epaulsen:
@mgravell : Any issues with this, can this be merged?

NickCraver:
Overall: this is a breaking change (adding optional parameters is a binary breaking change). We definitely get the use case though. Current thinking is we have something like an `IDapperConfig` object that's ambiently available via `AsyncLocal<T>` or something you can pass into all methods. This approach means we can add things like this action on the config, rather than take a breaking change (and new overloads) on every addition.

This is what we're plotting out for vNext overall (you can see some experimental work over in the DapperAOT repo, Marc's working on generator approaches). We can't take this set of changes until a breaking release _anyway_, so the best bet here is to go with the config approach and ensure we add this functionality as one of the config options there.

Does that sound like a workable approach, where you can set once, or per query, options like this?

epaulsen:
Given that the issue is more than five years old, this is obviously not something that is extremely urgent.  
What is the timeline for vNext? 

For this use case, we would need to set the fetchsize per query.  Normal workflow when working with oracledatareader is to retreive the rowsize on the datareader after `cmd.ExecuteReader` and then do some calculations to determine the optimal fetchsize.

vinaymadupathi:
+1 When can this MR be merged and released

Issue #1318, opened 2019-08-28
FeatureSupport broken when using a wrapped IDbConnection


jhaygood86:
Our code uses a "wrapped connection" for APM tracing (inspired by mini profiler), and so this code in FeatureSupport.cs doesn't work:

`            public static FeatureSupport Get(IDbConnection connection)
            {
                string name = connection?.GetType().Name;
                if (string.Equals(name, "npgsqlconnection", StringComparison.OrdinalIgnoreCase)) return Postgres;
                return Default;
            }`

Essentially, connection is a "TracingDbConnection" that wraps and proxies an underlying "NpgsqlConnection" while sending traces to DataDog (that part isn't of concern here). There doesn't appear to be a way for code outside of Dapper to "correct" this.

It looks like using MiniProfiler has the same issue: https://github.com/MiniProfiler/dotnet/issues/319 (which makes sense, since our approach was inspired by MiniProfiler's)

mgravell:
Yes, totally agreed and acknowledged; and yes: I feel this pain too (Stack Overflow uses something *also* related to, but not identical to, mini-profiler as a decorator layer)

*ideally* IMO what we need here is to move away from internal detection based on the type, and towards some kind of external configuration where the caller can influence this; that's part of the "major API overhaul" changes planned.

Note: I *would* have said "v2", but we're in the process of deploying v2 *currently* but with fairly invisible API surface changes, so in reality I think this means "v3" (this change from v2 to v3 has zero effect on timing; the current v2 is purely a semver thing)

kieranbenton:
We have exactly the same issue, wrapping the NpgSqlConnection to add telemetry which prevents Dapper from detecting the connection as being 'array supporting'. 

Whilst I 100% do not condone doing this, if you are looking to work around this issue and have Dapper treat all connections as supporting arrays GLOBALLY - then we do some awful reflection based operations on application start-up as follows:

```
public static void DisableArrayExpansion()
{
	var featureSupportType = Type.GetType("Dapper.FeatureSupport, Dapper", throwOnError: true)!;

	var d = featureSupportType.GetField("Default", BindingFlags.Static | BindingFlags.NonPublic)!.GetValue(null);
	var arraysField = featureSupportType.GetField("<Arrays>k__BackingField", BindingFlags.Instance | BindingFlags.NonPublic);

	bool currentArraySupport = (bool)arraysField!.GetValue(d)!;
	if (currentArraySupport)
[...]

Issue #1153, opened 2018-11-05
Npgsql.PostgresException: 42809: op ANY/ALL (array) requires array on right side


babakarj:
Following this issue https://github.com/MiniProfiler/dotnet/issues/319 . The problem is still there.

When changing
`IDbConnection connection = new NpgsqlConnection(connectionString);`
to
`IDbConnection connection = new StackExchange.Profiling.Data.ProfiledDbConnection(new NpgsqlConnection(connectionString), MiniProfiler.Current);`

Queries containing `Any()` fails with following error:

```
Npgsql.PostgresException
  HResult=0x80004005
  Message=42809: op ANY/ALL (array) requires array on right side
  Source=Npgsql
  StackTrace:
   at Npgsql.NpgsqlConnector.<>c__DisplayClass161_0.<<ReadMessage>g__ReadMessageLong|0>d.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Npgsql.NpgsqlConnector.<>c__DisplayClass161_0.<<ReadMessage>g__ReadMessageLong|0>d.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Npgsql.NpgsqlDataReader.<NextResult>d__46.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Npgsql.NpgsqlDataReader.NextResult()
   at Npgsql.NpgsqlCommand.<ExecuteDbDataReader>d__100.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Npgsql.NpgsqlCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at StackExchange.Profiling.Data.ProfiledDbCommand.ExecuteDbDataReader(CommandBehavior behavior) in C:\projects\dotnet\src\MiniProfiler.Shared\Data\ProfiledDbCommand.cs:line 205
   at System.Data.Common.DbCommand.System.Data.IDbCommand.ExecuteReader(CommandBehavior behavior)
[...]

PR #1518, opened 2020-08-14
support using a wrapped IDbConnection in FeatureSupport

qihangnet:
It will solve the issues about using MiniProfiler. For example #1153 #1318 ...

### DIFF
--- a/Dapper.Tests/Dapper.Tests.csproj
+++ b/Dapper.Tests/Dapper.Tests.csproj
@@ -13,6 +13,7 @@
   </PropertyGroup>
   
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


NickCraver:
Instead of this approach, what we are looking at in vNext is an `IDapperConfig` or some such (present ambient and switch-able) which will allow explicit configuration, rather than relying on detection at all going forward (due to the valid problems you point out here). I'm trying to clean up existing PRs that are either breaking, or conflict with v3 direction to solve the same issues. We do appreciate the contributions and I _hate_ closing PRs, but leaving them forever for lack of time and mismatch on direction isn't fair either, so I'm biting the bullet and trying to sync up current PR state tonight.

PR #1535, opened 2020-09-08
Avoid exception if the key is not Int32

Serg046:
I experience an issue when the key typed as GUID so that I was forced to prepare a custom `ISqlAdapter`, copy paste the code and remove the conversion to int. That's why I suggest checking the type before we cast the result. If you find the change useful we can do the same for other adapters too. Just let me know and I will do it.


Serg046:
Is there any chance to push it? That's a bug. Nothing is working if the key is not a number.

NickCraver:

PR #1540, opened 2020-09-18
New property Settings.RunQueryAsyncMapOnCapturedContext

hermankan:
The property when enabled allows to use captured context by `Func<> map` in `QueryAsync()`, e.g. HttpContext in ASP.NET application.

### DIFF
--- a/Dapper/SqlMapper.Async.cs
+++ b/Dapper/SqlMapper.Async.cs
@@ -915,9 +915,9 @@ namespace Dapper
             bool wasClosed = cnn.State == ConnectionState.Closed;
             try
             {
-                if (wasClosed) await cnn.TryOpenAsync(command.CancellationToken).ConfigureAwait(false);
+                if (wasClosed) await cnn.TryOpenAsync(command.CancellationToken).ConfigureAwait(Settings.RunQueryAsyncMapOnCapturedContext);
                 using (var cmd = command.TrySetupAsyncCommand(cnn, info.ParamReader))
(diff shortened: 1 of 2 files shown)


mgravell:
Question: why? What problem are you seeing that requires this?

hermankan:
There may be ambient data that is used in the `map`, so instead of passing it all the way down the stack to a `QueryAsync()` caller, the data is stored in HttpContext.

mgravell:
I wonder if it would be more pragmatic to include the http context in the capture scope of the lambda used, which would.be simpler and would work *today*; alternatively, execution-context can store ambient data and does not get bypassed by configure-await. Any use?

hermankan:
Indeed, we could use a closure in the lambda, however in our case the lambda is just an entry point to its own call stack to which we would have to pass the closure.

Not sure, how ExecutionContext can help us resolve the issue.

mgravell:
So yes, I'd capture and pass down;

``` c#
var ctx = HttpContext.Current;
Foo(..., () => YourMethod(ctx));
```

Apart from anything else, you're going to have to solve this same problem
when you migrate to .NET Core / .NET 5, since the ambient http context
isn't directly available, which means the scenario you're fixing only helps
with platforms that are likely to drop off the support radar sooner rather
than later (
https://blog.marcgravell.com/2020/01/net-core-net-5-exodus-of-net-framework.html
)

hermankan:
We are supporting a legacy system, so migration to the latest .NET is going to be a huge overhaul per se. Okay, I can live with that, thanks.

NickCraver:
Closing this out to cleanup!

PR #1544, opened 2020-09-22
Update PostgresAdapter to Generate Id Column with AppendColumnName

dogac00:
Currently, PostgresAdapter generates query for columns with quotes like this: `"Column"`. But it does not add quotes in returning clause when we are returning `Id`.

I think this PR should fix "*Column Id does not exist.*" error we encounter in PostgreSQL while using `InsertAsync`.

Related issues:
https://github.com/StackExchange/Dapper/issues/481
https://github.com/StackExchange/Dapper/issues/1213
https://github.com/StackExchange/Dapper/issues/1541

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -493,7 +493,7 @@ public partial class PostgresAdapter
                 if (!first)
                     sb.Append(", ");
                 first = false;
-                sb.Append(property.Name);
+                AppendColumnName(sb, property.Name);
             }
         }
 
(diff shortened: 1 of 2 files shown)


NickCraver:
Unfortunately, this is a breaking change. While it may help you, it will break others - please see #722 on future plans there. In the mean time, you can quote the column name in your attributes if you want it to be case-sensitive.

dogac00:
Which attribute can we use on our property if we want it to be case-sensitive and quoted? As far as I know, Dapper.Contrib currently does not have `ColumnAttribute`.

dogac00:
In addition, don't you think there is an inconsistency in this generated query:

```INSERT INTO Products ("Name", "ItemNumber") VALUES (@Name, @ItemNumber) RETURNING Id```

Columns are quoted but `Id` is not. I think if we quote columns we should quote Id, because it is a column too.

Issue #1536, opened 2020-09-10
Reducing the number of queries created with Decimal data types


bsstahl:
Note: The goal of this issue is to determine if I should submit a pull-request that I think will make the development experience better when facing one particular problem. My understanding of best-practice is to open an issue first to discuss before simply submitting a PR. I have searched StackExchange and this repository for anything that might already have been done relating to this issue and was unable to find anything. If this has already been addressed, or if I have misunderstood something, please let me know.

As I understand it today, the default behavior for decimal (numeric) data types is for Dapper to submit queries using the smallest possible data type that the data will fit in. Thus, if the value is 123.45, it will be submitted in the query as decimal(3,2) assuming the type has not been explicitly specified. This is a very reasonable default, but does result in what is ultimately different queries being submitted to the DB depending on the values of the parameters.

These different queries are not usually a problem, but can make analysis of those queries more difficult, can cause some very small performance impact since more query plans need to be generated, and in those rare situations where the query plan needs to be pinned, can make that task much more difficult since there can be hundreds of queries depending on the variability of the data.

This problem is solvable by explicitly giving Dapper a data type that matches the column's data type in the database. This way, the query will always be submitted using that type, and only 1 query will be created regardless of how many times the query is executed and the variability of the data. Often, that looks something like this:

```
	var parameters = new DynamicParameters();
	parameters.Add("decimalParameter1", valueForParameter1, DbType.Decimal, ParameterDirection.Input, 5, 9, 2);
	conn.Execute(query, parameters);
```

The 5 that preceeds the `precision` and `scale` values (9,2) in the parameters.Add(...) method represents the `size` of the parameter value, and can be calculated using this code block:

```
	// Values from Transact-SQL Reference > Numeric > decimal & numeric
	// https://docs.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver15
	int result = 5;
	if (precision > 28) result = 17;
	else if (precision > 19) result = 13;
	else if (precision > 9) result = 9;

```

The resulting syntax is not horrible by any stretch of the imagination, but it certainly could be cleaner, especially when there are a number of decimal data types in the same query. Also, the functionality in the code block above that calculates the `size` should probably be encapsulated somewhere so it doesn't have to proliferate. It makes sense to me that this code should live in the Dapper project if possible.

[...]

PR #1547, opened 2020-09-27
Add DbDecimal object

bsstahl:
This simplifies the syntax that allows queries to run with the same parameter types regardless of the actual size of the data as described in issue #1536.

### DIFF
--- a/Dapper.Tests/MiscTests.cs
+++ b/Dapper.Tests/MiscTests.cs
@@ -1196,6 +1196,67 @@ insert TPTable (Value) values (2), (568)");
             Assert.Equal(568, row.Value);
         }
 
+        [Fact]
+        public void TestDbDecimal()
+        {
+            var obj = connection.Query("select sql_variant_property(@a, 'precision') as a_precision, sql_variant_property(@a, 'scale') as a_scale, " +
+                                              "sql_variant_property(@b, 'precision') as b_precision, sql_variant_property(@b, 'scale') as b_scale, " +
+                                              "sql_variant_property(@c, 'precision') as c_precision, sql_variant_property(@c, 'scale') as c_scale, " +
+                                              "sql_variant_property(@d, 'precision') as d_precision, sql_variant_property(@d, 'scale') as d_scale, " +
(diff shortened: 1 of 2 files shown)


mgravell:
We will look when we have a moment. Sorry for delay, but we have lots of
things on the go, and sadly this isn't our day job.

bsstahl:
While I really hope this item is eventually merged into the codebase, I understand and appreciate that working on this project is a labor of love and not anyone's paying job. That said, my DBAs have been _extremely patient_ in tolerating the lack of consistency in Decimal data-types and the impacts to their tooling and processes. As such I need to move forward with a work-around of some type. It seems to me that my options are the following:

1. **Use my fork of Dapper**, which contains this change, until the PR is merged, and hope that it eventually does get merged. The obvious drawback here is that we will either miss-out on any changes made to the main codebase, or I will be responsible for keeping my fork in sync with the main one. I can, of course, do this for some time but would prefer not to be responsible for this in the long term. I would also prefer not to put the team through the need to change the tooling now, then change it again later if the PR is merged. All that said, this is not a terrible work-around and is the one I'm leaning toward.
	
2. **Use syntax that exists today** and skip the _DbDecimal_ object altogether. This is also not a terrible option and it avoids having to utilize a fork of this project, however it has the drawback of requiring somewhat obscure code to be repeated throughout our codebase. Another drawback is, that while it would not be **required** to change the code again if this PR is ever merged, it would make sense to do so since the syntax is _so much_ simpler.
	
3. **Add the functionality to each codebase that contains the queries**, perhaps using an extension method, rather than to Dapper. While this is not a totally unreasonable option, it feels like the wrong one to me since that functionality clearly belongs to Dapper and would need to be repeated in every codebase that needs it. It could be added to a shared package, but that option is similar to option 1 above with many of the same drawbacks.

If anyone has any recommendations or concerns with any of these choices I would appreciate the feedback. Again, I am well aware of the demands that OSS creators have on their time and I don't want this in any way to be seen as a negative toward the project or its maintainers. This project has been and continues to be an important and valuable one to me and my company.

PR #1552, opened 2020-10-02
Added CrossApply method to SqlBuilder

Calidus:
Added CrossApply method to SqlBuilder.

https://github.com/StackExchange/Dapper/issues/1551


Calidus:
Should cross apply have open and closing parentheses?

PR #1553, opened 2020-10-08
Prevent race conditions

SanSYS:
Our application's init process is paralleled for some reasons
And some developers in their own parts of code use AddTypeHandlerImpl for their own handlers
But in really rarelly moments we received exception like this: Invalid cast from 'System.String' to 'Newtonsoft.Json.Linq.JObject' (and yet another random classes)
Why: because typeHandlers didn't contain an appropriate handler)

Of cause race fixed by some changes in our code, but we would like have fix here, if it's possible

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -290,29 +290,32 @@ namespace Dapper
                 }
             }
 
-            var snapshot = typeHandlers;
-            if (snapshot.TryGetValue(type, out ITypeHandler oldValue) && handler == oldValue) return; // nothing to do
+            lock (typeHandlersLocker)
+            {
+                var snapshot = typeHandlers;
+                if (snapshot.TryGetValue(type, out ITypeHandler oldValue) && handler == oldValue) return; // nothing to do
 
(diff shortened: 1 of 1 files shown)


SanSYS:
@mgravell, @NickCraver ?

NickCraver:
Is there any information about what you're trying to resolve here? Normally changes would have some explanation to them, all we have is an empty PR body to go off of :)

SanSYS:
@NickCraver Hi)
Our application's init process is paralleled for some reasons
And some developers in their own parts of code use `AddTypeHandlerImpl` for their own handlers
But in really rarelly moments we received exception like this: `Invalid cast from 'System.String' to 'Newtonsoft.Json.Linq.JObject'` (and yet another random classes)
Why: because typeHandlers didn't contain an appropriate handler)

Of cause race fixed by some changes in our code, but we would like have fix here, if it's possible

SanSYS:
@NickCraver we need the fix, sometimes (1 per month) we have a trouble with it
Please, review, may be test, and try merge 🤞

PR #1578, opened 2020-11-21
Fix Insert Query Returning Statement Column Names Formatting

berkayalcin:
The quote mark required to format the Postgres column names were added to the returning statement and the code was simplified.


NickCraver:
I can't speak to Contrib's future plans here and if this is a good fit, but please see #1658 - Dapper.Contrib is now located at https://github.com/DapperLib/Dapper.Contrib (separate from this repository). Please do be aware, this would still be a breaking-change in behavior as-is.

PR #1586, opened 2020-12-01
Added overloads allowing predicate expressions to limit scope

Simon-Carr:
_DeleteAllAsync()_ and _GetAllAsync()_ overloads accept a predicate expression that can be used to limit the scope of the action. I.E filter records returned or deleted.

This is obviously limited compared to what you can achieve with the QueryAsync or ExecuteAsync methods but does allow you to stay in the Dapper.Contrib world for simple use cases and not require any SQL strings.

For example to return entities where Age = 5:

`await connection.GetAllAsync<User>(x => x.Age == 5) `

For example to delete entities where Age = 5:

`await connection.DeleteAllAsync<User>(x => x.Age == 5) `

Some supporting tests have been added.

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Expression.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Expression.Async.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dapper.Contrib.Extensions
+{
+    public static partial class SqlMapperExtensions
+    {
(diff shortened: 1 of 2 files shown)


Simon-Carr:
I'm not sure why the build failures occur for this PR, can anyone advise?

I submitted this because the functionality would be super-useful for my own projects and therefore, I assume, others would find it useful too. 
Can anyone give any feedback?

NickCraver:
I can't speak to Contrib's future plans here and if this is a good fit, but please see #1658 - Dapper.Contrib is not located at https://github.com/DapperLib/Dapper.Contrib (separate from this repository). If you're wanting predicated functionality on builders, I'd personally recommend EntityFramework Core overall as it has progressed and gotten quite efficient.

PR #1593, opened 2020-12-12
Contributing fixes to GitHub Actions workflow (+ClickHouse container)

DarkWanderer:
This PR:

- Limits frameworks used in GH Action build to .NET 5.0 (build succeeds now)
- Adds ClickHouse container (for which I want to contribute test later)

See [example](https://github.com/DarkWanderer/Dapper/runs/1542047673?check_suite_focus=true) of successful build

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,13 +1,14 @@
 name: Main Build
 
 on:
-  pull_request:
   push:
-    branches:
-    - main
-    paths:
-    - '*'
-    - '!/docs/*' # Don't run workflow when files are only in the /docs directory
+    branches: [ main ]
+    paths-ignore:
+    - '**/*.md'
+  pull_request:
+    branches: [ main ]
+    paths-ignore:
+    - '**/*.md'
 
 jobs:
   vm-job:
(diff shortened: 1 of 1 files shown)


NickCraver:
Thanks for this! It is known that installing any .NET Core SDK breaks all others already on the agent, but we can still build multiple versions. What we'd need to do here is install both .NET 5.0.101 and the 3.1.404 line...but, our problem has also been solved by waiting for the agents to catch up here, now 5.0.101 is on the GitHub agents :) Link here: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

Overall: we don't need any build fixes at this point in time - I'd recommend we move the clickhouse addition to be with tests when you're ready, and have that bundled in a single PR rather than an inexplicable (until then) service container running - thoughts?

DarkWanderer:
Fair enough - at the moment when I crafted this the build was failing. No problem moving the container addition to PR with tests

NickCraver:
Sounds good (and this is appreciated!) Looking forward to ClickHouse, I'm not familiar with it so will learn a bit :)

PR #1595, opened 2020-12-15
Id should not be a Key when we have any ExplicitKey

AartBluestoke:
Id is assumed to be a Key if the object has no Key, (and Id is not an ExplicitKey) 

Instead we need to check for any ExplicitKey. 

The failure case is if you have a column called "ID" and you don't want it be a primary key. 

(yes this is an unusual / not best practice situation, but sometimes you can't control the names of the "attributes" you are requested to add to tables, and the ORM shouldn't have an opinion in this case)


NickCraver:

PR #1597, opened 2020-12-18
fix InsertAsync do not return auto increase primary id when dbconnect…

sampsonye:
InsertAsync Extension Method dosn't return auto increase primary id


NickCraver:

PR #1598, opened 2020-12-20
Add support for Array return value (i.e.: QueryAsync<int[]>))

DarkWanderer:
This change allows to execute queries which return arrays as first-class values, i.e.: `QueryAsync<byte[]>(sql).`

If the type was not guessed exactly (i.e.: `int[]` vs `byte[]`), a conversion is performed

"Encore" of #1467 (branch was deleted accidentally)


NickCraver (approved):
Looking good - thanks again for updating against `main`.

PR #1602, opened 2020-12-29
Add performance benchmarks tests for Norm ORM

vbilopav:
Add performance benchmarks tests for [Norm ORM](https://github.com/vb-consulting/Norm.net).

Includes three methods.

### DIFF
--- a/benchmarks/Dapper.Tests.Performance/Benchmarks.Norm.cs
+++ b/benchmarks/Dapper.Tests.Performance/Benchmarks.Norm.cs
@@ -1,6 +1,4 @@
-﻿#if NET4X 
-#else
-using BenchmarkDotNet.Attributes;
+﻿using BenchmarkDotNet.Attributes;
 using System.ComponentModel;
 using System.Linq;
 using Norm;
@@ -39,4 +37,4 @@ namespace Dapper.Tests.Performance
         }
     }
 }
-#endif
+
--- a/benchmarks/Dapper.Tests.Performance/Benchmarks.Norm.cs
+++ b/benchmarks/Dapper.Tests.Performance/Benchmarks.Norm.cs
@@ -15,21 +15,21 @@ namespace Dapper.Tests.Performance
             BaseSetup();
         }
 
-        [Benchmark(Description = "Read<T> (class)")]
+        [Benchmark(Description = "Read<> (class)")]
         public Post Read()
         {
             Step();
             return _connection.Read<Post>("select * from Posts where Id = @Id", i).First();
         }
 
-        [Benchmark(Description = "Read<T1, T2, ...> (simple values in a tuple)")]
+        [Benchmark(Description = "Read<> (tuples)")]
         public (int, string, DateTime, DateTime, int?, int?, int?, int?, int?, int?, int?, int?) ReadSimpleValues()
         {
             Step();
             return _connection.Read<int, string, DateTime, DateTime, int?, int?, int?, int?, int?, int?, int?, int?>("select * from Posts where Id = @Id", i).First();
         }
 
-        [Benchmark(Description = "Read<(T1, T2, ...)> (named tuple)")]
+        [Benchmark(Description = "Read<()> (named tuples)")]
         public (int Id, string Text, DateTime CreationDate, DateTime LastChangeDate, int? Counter1, int? Counter2, int? Counter3, int? Counter4, int? Counter5, int? Counter6, int? Counter7, int? Counter8) ReadTuple()
         {
             Step();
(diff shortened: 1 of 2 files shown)


benchmarks/Dapper.Tests.Performance/Benchmarks.Norm.cs
```diff
@@ -0,0 +1,42 @@
+﻿#if NET4X
```
Zubaloo, line 1: Why not "#if !NET4X"? then you wouldn't need the empty "then" branch nor "else" statement.

Also, the package reference comes under condition "'$(TargetFramework)' == 'netcoreapp3.1'" which means that if TargetFramework would be set to 'net5.0' then the reference will be missing, but the NormBenchmarks class would be presented and hence there will be a compilation error. So I would propose to have the benchmark file itself in .csproj to be under the same condition that is used for PackageReference instead of having the "#if" directive inside the file.

vbilopav:
>  So I would propose to have the benchmark file itself in .csproj

I love the idea of having a benchmark file in a separate .csproj (if I understood correctly), however, the benchmark tests themselves depend on a `BenchmarkBase` and `Post` classes.

So, referencing those would cause circular reference problems. 

The way around it would be to add a third, shared project called `Dapper.Tests.Performance.BenchmarkBase` that would be shared between the main benchmark project `Dapper.Tests.Performance` and a specific test project, like this one for Norm `Dapper.Tests.Performance.Norm`.

This requires a bit of refactoring, would that be ok to proceed with this?

Or just leave it as is and fix those conditional preprocessors?

Zubaloo:
I did not mean to add a new .csproj file, but to conditionally exclude the benchmark file from compilation in the existing `Dapper.Tests.Performance.csproj`, like:

``` xml
<ItemGroup>
  <Compile Remove="Benchmarks.Norm.cs" Condition="'$(TargetFramework)' != 'netcoreapp3.1'" />
</ItemGroup>
```

benchmarks/Dapper.Tests.Performance/Dapper.Tests.Performance.csproj
```diff
@@ -48,4 +48,12 @@
...
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="5.0.0" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1' Or '$(TargetFramework)' == 'net5.0'">
+    <PackageReference Include="Norm.net">
+      <Version>3.1.0</Version>
+    </PackageReference>
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Remove="Benchmarks.Norm.cs" Condition="'$(TargetFramework)' != 'netcoreapp3.1' And '$(TargetFramework)' != 'net5.0'" />
+  </ItemGroup>
```
NickCraver, line 58: Instead of all these one-off changes, please use the `<ItemGroup Condition="'$(TargetFramework)' != 'net462'">` section above for a normal include, then use `#if !NET4X` in the file itself. This will make it behave like other benchmarks overall.

NickCraver (changes requested):
2 requests here please:
1. (inline comment) Let's normalize the structure with other benchmarks/nuget includes
2. Can we please find shorter names for the description - right now these would be by far the longest names for methods, making the actual results table wrap overall in GitHub.

benchmarks/Dapper.Tests.Performance/Dapper.Tests.Performance.csproj
```diff
@@ -47,5 +47,8 @@
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' != 'net462'">
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="5.0.0" />
+    <PackageReference Include="Norm.net">
+      <Version>3.2.0</Version>
+    </PackageReference>
```
NickCraver, line 52: Tidying/normalizing here a bit!
```suggestion
    <PackageReference Include="Norm.net" Version="3.2.0" />
```

NickCraver (approved):
Looking good - thank you for the contribution!

NickCraver:
I'm getting a new Windows machine together now - plan to update the benchmarks in the README in the next week or two as parts get here. The macOS results are very, very unfair with things in containers and the times being heavily randomized by Docker's attitude at any given time :)

vbilopav:
Thanks @NickCraver 

You might want to include new incoming EF Core 6

https://devblogs.microsoft.com/dotnet/announcing-entity-framework-core-6-0-preview-4-performance-edition/

Let me know if you would require any help with that.

vbilopav:
Hi @NickCraver ,

if new benchmarks aren't released yet, would it be possible to update Norm to new version 3.3.0 or should I create a PR for that?

Thank you very much

PR #1603, opened 2021-01-01
Belgrade benchmark bug fix.

Zubaloo:
The used Map method is asynchronous, hence without waiting for operation to complete the benchmark is only counting the query invocation time.

### DIFF
--- a/benchmarks/Dapper.Tests.Performance/Benchmarks.Belgrade.cs
+++ b/benchmarks/Dapper.Tests.Performance/Benchmarks.Belgrade.cs
@@ -18,30 +18,28 @@ namespace Dapper.Tests.Performance
             _mapper = new QueryMapper(ConnectionString);
         }
 
-        [Benchmark(Description = "ExecuteReader")]
-        public async Task<Post> ExecuteReader()
+        [Benchmark(Description = "FirstOrDefault")]
+        public Task<Post> FirstOrDefault()
         {
             Step();
-            var post = new Post();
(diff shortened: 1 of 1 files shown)


Zubaloo:
@mgravell, @NickCraver

How much time usually a pull request stays opened before it receives any feedback?

NickCraver:
Sorry for the delay - I was on sabbatical and stepped away for a few months. Thanks for alerting us to the issue! The fix here needs to be a bit different though (I'll push a PR in a few) - we need to async/await the result, rather than using `.Wait()` in general so it's a fair test and doesn't hit threading issues, etc. In general, you want to avoid `Task.Wait()` in pretty much every case you can :)

NickCraver:
I've opened #1655 (because I can't push commits to the PR source here) - not trying to remove credit especially from a first time committer though, would you like to update this PR and I can close the other out?

Zubaloo:
> would you like to update this PR

I've updated the PR with your async/await proposal and went beyond that and used a more appropriate method for materialization for that particular benchmark case, which actually eliminates the need for both `Wait()` and async/await.

Zubaloo:
Do not forget to update the `Readme.md` with the benchmark results after applying the fix.

NickCraver (approved):
Thanks!

Zubaloo:
@NickCraver A friendly reminder that the `Readme.md` still shows the benchmark results without that fix.

PR #1622, opened 2021-02-05
Accept an optional table name in Dapper.Contrib operations

### LINKED ISSUE #1621: Dapper.Contrib methods should accept an optional table name
There are scenarios where one could prefer specifiing the table name for that specific insert/update/etc.

My suggestion is to add an optional parameter that when not specified would make the methods work as they currently do.
If specified, the method then uses that table name.

I have considered to configure through `SqlMapperExtensions.TableNameMapper`, but it still is a problem if I need to insert the same entity in two different tables concurrently.

mniak:
Closes #1621

**List of operations:**
- [x] Get
- [x] GetAll
- [x] Insert
- [x] Update
- [x] Delete
- [x] DeleteAll

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -22,13 +22,13 @@ namespace Dapper.Contrib.Extensions
         /// <param name="transaction">The transaction to run under, null (the default) if none</param>
         /// <param name="commandTimeout">Number of seconds before command execution timeout</param>
         /// <returns>Entity of T</returns>
-        public static async Task<T> GetAsync<T>(this IDbConnection connection, dynamic id, IDbTransaction transaction = null, int? commandTimeout = null) where T : class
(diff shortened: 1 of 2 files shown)


NickCraver:
Apologies for the delay - unfortunately this is a breaking change. It's likely we'll split Dapper.Contrib into another repo and find owners, but this can't be merged as-is except on a major version (while respecting SemVer). Ultimately, optional parameter additions break binary compatibility, so we'd either need to duplicate the methods, or go another route.

Overall, we're not headed this direction: we want to allow control of a type in attributes, but the use case on a table name per insert is not something we're likely to add to the core library (weighing confusion and maintenance vs. use). Overall, this conflicts with plans on attributes in the core library...which adds maintenance and complexity for that use case. 

Now in Dapper vNext we are considering `DapperConfig` to be an ambient object (`AsyncLocal<T>`) that you are not limited to 1 of, so instead of overloading methods with a table name, you could override the table name mapping _per config_ and get what you want overall. I hope that sounds reasonable as you'll have the functionality to do what you want, albeit via a very different path.

NickCraver:
See above for how this should likely change on approach, but overall also see #1658 - Dapper.Contrib is not located at https://github.com/DapperLib/Dapper.Contrib (separate from this repository)

PR #1625, opened 2021-02-16
add cancellationToken to async methods, add QueryAsync command method…

AlexeyVasiliev:
… with Array of types in the recordset

### DIFF
--- a/Dapper/SqlMapper.Async.cs
+++ b/Dapper/SqlMapper.Async.cs
@@ -21,9 +21,10 @@ namespace Dapper
         /// <param name="transaction">The transaction to use, if any.</param>
         /// <param name="commandTimeout">The command timeout (in seconds).</param>
         /// <param name="commandType">The type of command to execute.</param>
+        /// <param name="cancellationToken">The cancellation token for the command.</param>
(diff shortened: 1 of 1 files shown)


NickCraver:
Unfortunately - these are all breaking changes. It's something we'll address in the next major version of Dapper, but aren't aiming to do in the current APIs due to the breaking nature (adding optional params would break all libraries compiled on top of Dapper - it's not binary compatible).

I'll leave this open for any questions, but aim to cleanup and close out in a few days if all is understood here.

Herostwist:
> Unfortunately - these are all breaking changes. It's something we'll address in the next major version of Dapper, but aren't aiming to do in the current APIs due to the breaking nature (adding optional params would break all libraries compiled on top of Dapper - it's not binary compatible).

Would adding new overloads with non optional cancellationToken papramater still be considered a breaking change?

NickCraver:
> Would adding new overloads with non optional cancellationToken papramater still be considered a breaking change?

It would not, but it would create a _lot_ of extra methods to maintain. Given the plan to change the overall interface of how we layer in vNext, I think we'd want to add them once...and take this breaking change since people would need to compile against it anyway.

Dasch42:
Would it instead be possible to grant directly access to the currently private methods executing this, that accepts CommandDefinitions _(which does have CancellationTokens)_, in the mean time?

PR #1642, opened 2021-04-01
As far as I aware, Soma related code no longer in this project.

kant2002:
And thus can be removed


NickCraver (approved):
Thanks for the cleanup!

PR #1654, opened 2021-04-27
Use value-type accessors instead of GetValue to avoid boxing

MarkPflug:
This modifies the ILEmit data binder code to use the value-type accessors on IDataReader (GetInt32, GetFloat, GetDateTime, etc) instead of GetValue to avoid boxing the values. It also avoids calling IsDBNull for columns where the schema indicates that the value will never be null.

I added a benchmark which reads the entire (modified) Posts table. The original posts table had a large (2k) text column that seemed to dominate the benchmark time. I reduced this data to "Test text", and made all of the `CounterX` columns non-nullable to try to measure the IsDBNull optimization.

Results of this change as measured on my machine:

NET48:

|    ORM |         Method | MODE |     Mean |   StdDev |    Error |    Gen 0 |    Gen 1 |   Gen 2 | Allocated |
|------- |--------------- |------- |---------:|---------:|---------:|---------:|---------:|--------:|----------:|
| Dapper | 'Query extent' | Before | 11.57 ms | 0.361 ms | 0.545 ms | 462.0000 | 100.0000 | 12.0000 |   2.66 MB |
| Dapper | 'Query extent' | After | 9.533 ms | 0.2729 ms | 0.4126 ms | 234.0000 | 76.0000 | 10.0000 |   1.28 MB |

NET5.0:

|    ORM |         Method | MODE|     Mean |    StdDev |     Error |    Gen 0 |   Gen 1 |   Gen 2 | Allocated |
|------- |--------------- |------- |---------:|----------:|----------:|---------:|--------:|--------:|----------:|
| Dapper | 'Query extent' | Before | 7.886 ms | 0.0470 ms | 0.0790 ms | 428.0000 | 96.0000 | 10.0000 |    2.5 MB |
| Dapper | 'Query extent' | After | 6.497 ms | 0.0373 ms | 0.0714 ms | 204.0000 | 70.0000 | 8.0000 |   1.12 MB |

As expected, this results in a significant reduction in allocation, and also affords a moderate improvement in performance for this specific benchmark.

There is one minor breaking change: conveying the value of value-type fields to the exception catch block is very difficult without boxing it. Rather than doing that, I opted to instead convey the *type* of the value-type only. The effect of this change can be seen in one of the unit tests, the message doesn't display the specific value anymore (for value-type columns only), but will still display the ordinal and the field type. I hope this is an acceptable break. I think the impact should be low, because the majority of exceptions would likely occur when processing string values, which would still have their value available in the exception message because they are ref-types.

All of the Dapper unit tests pass. I do worry a bit about edge cases though, as in my limited experience bugs in this code tend to result in one of InvalidProgramException, VerificationException or AccessViolationException, all of which are bad and the the last of which is catastrophic. Certainly a thorough review by someone more familiar with ILEmit is in order.

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -3628,7 +3628,6 @@ namespace Dapper
             else if ((op = GetOperator(from, to)) != null)
             {
                 // this is handy for things like decimal <===> double
-                //il.Emit(OpCodes.Unbox_Any, from); // stack is now [target][target][data-typed-value]
                 il.Emit(OpCodes.Call, op); // stack is now [target][target][typed-value]
             }
             else
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -3218,6 +3218,8 @@ namespace Dapper
                     LoadDefaultValue(il, targetType);
 
                     il.MarkLabel(finishLabel);
+                    // without this cast an VerificationException is thrown when JITing
+                    il.Emit(OpCodes.Castclass, targetType);
                 }
                 else
                 {
@@ -3513,11 +3515,15 @@ namespace Dapper
             {
                 if (isValueType)
                 {
+                    // TODO: decide if this compromise is acceptable. The only unit test affected was processing a Guid value.
+
                     // if it is a value type, we want to avoid boxing it
-                    // this comes at the cost of not seeing the value in the exception message.
-                    // however, such messages would typically only happen for string values
-                    // which would still be handled.
-                    il.Emit(OpCodes.Ldnull);
+                    // unfortunately, conveying the value to the exception block
+                    // without boxing it is extremely complex (and probably slow).
+                    // Compromise by passing  the type of the column instead of the value.
+                    // Most commonly these exceptions would happen when processing strings,
+                    // in which case the value would still be passed, since it is a ref-type (else block below).
+                    il.Emit(OpCodes.Ldtoken, colType);
                     il.Emit(OpCodes.Stloc, valueCopyLocal); // stack is now [...][value]
                 }
                 else
@@ -3746,13 +3752,22 @@ namespace Dapper
                     }
                     try
                     {
-                        if (value == null || value is DBNull)
+                        // for value-type columns the value isn't available here,
+                        // but the type is. This is to avoid boxing the value.
+                        if (value is Type type)
                         {
-                            formattedValue = "<null>";
+                            formattedValue = "[" + type.Name + " value]";
                         }
                         else
                         {
-                            formattedValue = Convert.ToString(value) + " - " + Type.GetTypeCode(value.GetType());
+                            if (value == null || value is DBNull)
+                            {
+                                formattedValue = "<null>";
+                            }
+                            else
+                            {
+                                formattedValue = Convert.ToString(value) + " - " + Type.GetTypeCode(value.GetType());
+                            }
                         }
                     }
                     catch (Exception valEx)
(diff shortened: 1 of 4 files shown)


NickCraver:
The main issue here is unfortunately an ecosystem one we do not control: some ADO.NET providers do not support all of these methods correctly, which is why Dapper uses the current agnostic approach to support as many providers as possible. While it is a performance hit, it's important that it works. To do better, we'd have to fork generation based on provider to be aware of those that can actually using this reliably.

@mgravell thoughts here? Since we're moving to a generator based approach I wouldn't want to do that split in IL here, but it is an option if we had an allow-list of providers or something in the AOT world.

MarkPflug:
Out of curiosity, what providers are known to have issues here? I would have expected most providers would implement `GetValue` by delegating to the type-specific accessors, in which case `GetValue` would be broken in the same way.

NickCraver:
@MarkPflug I'm sorry I don't have the list in front of me (I wish we had made an issue and written it down), but it's a thing across providers. Various ones do not support various methods. Part of that is because they do not have those data types in the underlying storage (e.g. Oracle has no GUID, SQLite has no...lots of things), the devs didn't know any better, or who knows what.

We tried this direction when first writing Dapper and hit roadblocks and `NotImplementedException` cases across the landscape. While I don't have a list, we know this will break providers and known use cases today. I think it'd be good info to have this list somewhere but that's a) time I don't have, and b) even if we did have it...it's not in our power to fix, unfortunately.

mgravell:
FWIW, my plan is to address some of this in the "child of dapper" (really,
Dapper vNext) work that I'm currently doing in a side repo. It does
emphasize the need for better capability APIs, but it is also very possible
that many of the gaps are now filled.

MarkPflug:
The decision of which accessor to use is determined by the result of `GetFieldType`, so I would expect that Oracle not supporting `Guid` would be fine: `GetFieldType` would never report the type as `Guid`, so the `GetGuid` would never be selected as an accessor. 

Maybe MySql is the odd-ball here. I know it uses a loose type system where a column type is merely a suggestion, you can put a string in an int column for example. I don't know the behavior of the MySQL .NET client though..

A provider reporting a specific type via `GetFieldType` and not supporting the corresponding accessor method seems really broken. But, if Dapper has chosen to support that broken-ness, I guess that ship has sailed.

MarkPflug:
>Maybe MySql is the odd-ball here. I know it uses a loose type system where a column type is merely a suggestion, you can put a string in an int column for example. I don't know the behavior of the MySQL .NET client though..

Sqlite, not MySql. I get those two confused.

MarkPflug:
Yes, SqLite is problematic:

```C#
        var conn = new SQLiteConnection("Data Source=test.db");
        conn.Open();
        var cmd = conn.CreateCommand();
        cmd.CommandText = "create table Test (Id int, Date text, Name text)";
        //cmd.ExecuteNonQuery();

        cmd.CommandText = "insert into Test (Id, Date, Name) values ('a', 'b', 'test')";
        //cmd.ExecuteNonQuery();

        cmd.CommandText = "select * from Test";
        var reader = cmd.ExecuteReader();
        var schema = reader.GetColumnSchema();
        while (reader.Read())
        {
            var t = reader.GetFieldType(0); // int32
            var t2 = schema[0].DataType; // int32
            var a = reader.GetValue(0); // 0 (boxed int)
            var b = reader.GetString(0); // "a"
            var c = reader.GetInt32(0); // throws InvalidCastException
        }
```

Honestly, the `InvalidCastException` is probably the most appropriate from my perspective. `GetValue` returning `0` seems insane to me.

It would be relatively easy to hide the new behavior behind a global switch. If the switch is set `GetAccessor` can be made to always return `GetValue` which should have the same behavior as the old code, even the exception messages should be unchanged. Alternately, inspect the runtime type of the reader, if the full name of the reader type is one of a well-known set (sql, postgres, etc) enable the optimization.

Issue #1648, opened 2021-04-09
C# 9 record vs class with SQL Money type - materialization error


DonThompson:
Issue:  A C# 9 record is not serializing a 'money' type value into a 'double' variable in the same way that a regular class works.
Workaround:  Just use a class, not a record.

I can build up a reproducible code if needed, but here's what I'm seeing:

Database table is something like

```
CREATE TABLE Test (Amount money NOT NULL);
INSERT INTO Test (Amount) VALUES (3.50);

```
---

```
class MoneyDBModel() {
  public double Amount { get; set; }
}

record MoneyDBRecord(double Amount);

string sql = "SELECT Amount FROM Test"; //and some other stuff

var results = DBConnection.QueryMultipleAsync(sql);

//Works fine
var moneyRecord = results.ReadAsync<MoneyDBModel>();

//Runtime ERROR:  A parameterless default constructor is required to allow for dapper materialization
var moneyRecord = results.ReadAsync<MoneyDBRecord>();
```

I've been using Dapper and record types all over, it only seems to be an issue with 'money' type values.

Issue #1496, opened 2020-07-06
Diferent result with QueryAsync and QueryMultipleAsync


Wqawer:
I am using .net core 3.1 with dapper 2.0.35

When i do something like this:
```
var q = "Select count(1)";
var count = await conn.QueryAsync<int>(q);
```

It works ok and returns integer, but when i do something like this: 
```
var q = "Select 'a'; Select count(1); ";
using(var multiple = await cnn.QueryMultipleAsync(q){
    var text= await multiple.ReadAsync<string>();
    var count = await multiple.ReadAsync<int>();
}
```
This throws System.InvalidCastException: Unable to cast object of type 'System.Int64' to type 'System.Int32'. I have found out that second query returns long instead of int and that is causing error.

Is this expected behaviour or is it bug?

mgravell:
Sounds like a bug with us not running all the same conversions in both cases.

NickCraver:
Added a fix for this (seen a few reports, all the same thing) in #1663.

PR #1575, opened 2020-11-19
GridReader.ReadAsync normalization: Convert.ChangeType like other paths

zqlovejyc:
This does proper type conversion like other paths so everything aligns. Fix for #1301, #1496, and #1648. Added tests that fail against current code here.

PR #1663, opened 2021-05-09
GridReader.ReadAsync normalization: Convert.ChangeType like other paths

NickCraver:
This does proper type conversion like other paths so everything aligns. Fix for #1496 and #1648. Added tests that fail against current code here.

We should follow-up here with an aggressively inlined shared approach, but getting everything normalized first.


NickCraver:
Just noticed the conflict with #1575 - will work on moving this test case over and bringing that branch up (it doesn't handle one of the call sites here).

NickCraver:
Closing this after merging in the extra bits to #1575 - we'll get the fix in there.

PR #1672, opened 2021-06-09
Fixed for custom parameter usage in parameter lists

MG-Russell:
Custom implementations of ICustomQueryParameter do not work when used as an "IN" list parameter. 
Fix uses the current implementation for DbString but changes the type check so that it works with any ICustomQueryParameter.


NickCraver:
Can you please add a test or two illustrating this working with other implementations?

PR #1689, opened 2021-07-26
Enables "snowflake" connection

mgravell:
Snowflake uses positional bind variables (`?`), with the additional nuance that the parameters must be *named* by their position, i.e. `"1"`, `"2"`, etc. Adds support and tests for this. Cross-ref: #1687


mgravell:
@NickCraver wouldn't mind your thoughts here; in particular, I'm very tempted to remove the configuration option, and make it *always* generate sequential names in the pseudo-positional case (perhaps using a `string[]` for low integers, and avoid the `ToString()`) - with the justification that every other provider to date simply ignores the name, so... meh.

NickCraver (approved):
:+1:

Issue #1293, opened 2019-07-11
V2 API refactors


mgravell:
So v2 is going to have to happen at some point; we will want to do as much as we can at one "break", so: thoughts here:

- [ ] unify strong-name
- [ ] remove overload-'splosion caused by gradual growth of overloads
- [ ] move from `IDbConnection` etc to `DbConnection` throughout
- [ ] move from `Task<T>` to `ValueTask<T>` throughout
- [ ] make sure `CancellationToken` is used on all async methods
- [ ] move the async-reader API to `IAsyncEnumerable<T>`
- [ ] consider (not definite) `PooledAwait`

other ideas?

mgravell:
heh, turns out we've already started this [here](https://github.com/StackExchange/Dapper/issues/688) - but: always good to get another round of input!

kevin-montrose:
Feels like there might be some value in keeping a `QueryAsync<T>` (or equivalent) that returns `Task<IEnumerable<T>>`.  The case of "I don't want to block on the query, but I don't want to proceed until I've got all the results" (as compared to "I don't want to block, but can proceed once the first row is available") doesn't seem un(reasonble|common).  It _kind_ of maps to buffered/not-buffered, but I don't know that that would make sense as a separate set of methods rather than the parameter it is now.

I, of course, am tempted to want the `MarkedSqlString`-stuff added (opt-in, naturally) since it's so damn useful.  Don't know if there's a clean way to do that though, the SO impl probably has some MS SQL Server assumptions.

mgravell:
That's actually very relevant, @kevin-montrose - because we could also take the opportunity to clarify buffered Vs non-buffered. Buffered => `List<T>` / `ValueTask<List<T>>`, non-buffered => `I[Async]Enumerable<T>`

jnm2:
There are many times where my method's return type is `Task<ImmutableArray<Foo>>` and I don't want to buffer to a `List<Foo>` since I have to do `.ToImmutableArray()` myself anyway. It would be cool if somehow the immutable array builder could be the buffer itself.

mgravell:
@jnm2 we'd still have a non-buffered API - it would just be more explicit in the naming; in your case, you'd presumably just write an extension method that wraps that up for you?

jnm2:
That sounds good.

vdaron:
[...]

PR #1692, opened 2021-08-04
Add StreamAsync alternatives to support IAsyncEnumerable

dwfrancis:
Add `StreamAsync` alternatives to `connecton.QueryAsync` and `gridReader.ReadAsync` to support `IAsyncEnumerable` in a non-breaking way.

Unlike `QueryAsync`, stream will default to unbuffered.

### DIFF
--- a/Dapper.StrongName/Dapper.StrongName.csproj
+++ b/Dapper.StrongName/Dapper.StrongName.csproj
@@ -5,7 +5,7 @@
     <Title>Dapper (Strong Named)</Title>
     <Description>A high performance Micro-ORM supporting SQL Server, MySQL, Sqlite, SqlCE, Firebird etc..</Description>
     <Authors>Sam Saffron;Marc Gravell;Nick Craver</Authors>
-    <TargetFrameworks>net461;netstandard2.0;netstandard2.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>net461;netstandard2.0;net5.0</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
   </PropertyGroup>
@@ -19,6 +19,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="5.0.0" />
     <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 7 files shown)


dwfrancis:
Intentionally omitted cancellation tokens to keep the signatures similar. 

Since these are new I can include them.

dwfrancis:
Re-authored commits (forgot to reset git config from a work account)

dwfrancis:
Completes the "new IAsyncEnumerable<T> API" task #1293

mgravell:
Will review when I'm back at in work-mode.

fdcastel:
> Will review when I'm back at in work-mode.

@mgravell... Are you back from your sabbatical? 😅

@dwfrancis Thanks for the excellent work! I'm using the code in some non-production capacity (going into production next month) and so far it is going smoothly. 👍

Sonic198:
I'm also waiting for that :D

pcooper:
I was able to work around lack of support with the current stable version by using `ExecuteReaderAsync` to get a data reader, then get a row parser from that.

```cs
var reader = await connection.ExecuteReaderAsync(sql, param);
var rowParser = reader.GetRowParser<MyType>();
while (await reader.ReadAsync())
{
  var value = rowParser(reader);
  // Do other sync/async stuff with value.
  yield return value;
}
```

This also worked with `dynamic` typed row parser funcs.

Conbag93:
Just wanted to also add my thanks to @dwfrancis for this PR.
I hope that it gets the offical 'nod' of approval because this has allowed me to experiment with some of the new-ish .net features wiith very minimal code changes.

Combining this with System.Text.Json's newly introduced support for asynchronous Json serialization allows my API to return each row to the client as soon as they are retrieved from the database.
Additionally, if the client also knows the "count" ahead-of-time then I can leverage this to track the progress of a long-running query.

SoftStoneDevelop:
It looks like a different implementation of what I did(https://github.com/DapperLib/Dapper/pull/1882) as part of the solution https://github.com/DapperLib/Dapper/issues/1239 (because in the comments to the issue it was https://github.com/DapperLib/Dapper/issues/1239#issuecomment-1035448069).

The fundamental difference is that I rewrote SqlMapper.Async and SqlMapper.GridReader.Async from pseudo-async to IAsyncEnumerable. And of course added "buffered" parameter.
Because my opinion is that there should be either a synchronous method or an asynchronous one. A pseudo-asynchronous wrapper over a synchronous one has no right to exist.

My point is that it might be worth choosing one of these two pull requests or considering them together.

ronimizy:
Any updates on that? 👀

SoftStoneDevelop:
> Any updates on that? 👀

No, Dapper seems to be abandoned.

mgravell:
Dapper is absolutely not abandoned; see #1909 for status; this feature now essentially provided via #1912 - so: I appreciate the input, but I think we should re-address needs when #1912 is in

PR #1699, opened 2021-08-30
Fix NETSDK1174

CaringDev:
As per https://docs.microsoft.com/en-us/dotnet/core/tools/sdk-errors/netsdk1174 `-p` is deprecated in favor of `--project`
This PR fixes the deprecation warning when running the benchmarks as currently described in the README.


NickCraver (approved):
👍 Thanks for getting this up to date!

PR #1700, opened 2021-08-30
EF Core benchmarks for connections w/ passwords

CaringDev:
`SqlConnection.ConnectionString` returns the original connection string _without_ the password.
Use `BenchmarkBase.ConnectionString` to ensure successful connections.


NickCraver (approved):
:+1: Thanks for the fix here!

PR #1707, opened 2021-09-22
Document usage of IDynamicParameters in more detail

bigjonroberts:
Considering all of the inquiries on StackOverflow about DynamicParameters, there should probably be a simple overview in the documentation.
Examples:

- https://stackoverflow.com/questions/12723922/dapper-adddynamicparams-for-in-statement-with-dynamic-parameter-name
- https://stackoverflow.com/questions/43412455/dapper-with-sql-query-dynamic-parameters
- https://stackoverflow.com/questions/38443739/passing-dynamically-created-sql-parameters-into-dapper-as-an-anonymous-type
- https://stackoverflow.com/questions/62448900/how-do-i-loop-through-dapper-dynamic-query-results-using-strings-as-parameters
- https://stackoverflow.com/questions/36735189/dapper-dynamicparameters-returning-error

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -187,12 +187,35 @@ Alternatively, you might prefer Frans Bouma's [RawDataAccessBencher](https://git
 Parameterized queries
 ---------------------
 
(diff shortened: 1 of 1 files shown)


NickCraver (approved):
Looking good, thanks for helping others here!

PR #1711, opened 2021-10-01
Allow to use  custom TypeHandlers for Enums

### LINKED ISSUE #259: Custom type handlers for enums are ignored

MaxShoshin:
Try to deserialize enum via TypeHandler first (if exists)

Closes #259

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -2942,6 +2942,14 @@ namespace Dapper
             }
 #pragma warning restore 618
 
+            if (typeHandlers.TryGetValue(type, out ITypeHandler handler))
+            {
+                return r =>
+                {
+                    var val = r.GetValue(index);
+                    return val is DBNull ? null : handler.Parse(type, val);
+                };
+            }
             if (effectiveType.IsEnum)
             {   // assume the value is returned as the correct type (int/byte/etc), but box back to the typed enum
                 return r =>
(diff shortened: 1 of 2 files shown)


heilmanBryan (approved):

heilmanBryan (commented):
Trying to remove my approval

SimonCull:
> Trying to remove my approval. I accidentally approved this pr and cannot seem to undo my approval. I'm not an expert on this code base and should not approve.

Can you put this in front of someone who can approve, this issue is 6 years old and the fix in simple enough.

PR #1718, opened 2021-10-15
Adding notes on the sub-packages' purpose to the readme

### LINKED ISSUE #1488: Add summary of the different packages (.Contrib, .SqlBuilder, etc.) to main README?
I'm wondering if it would be a good idea to add a brief summary of what the different sub-packages (Dapper.Contrib and the like) do to the main README. I didn't even realize they provided additional functionality at first, due to the lack of discoverability. Would this be helpful/appreciated?

RSid:
I just paraphrased the summaries from the sub-packages' Nuget Readmes and added them to the grid of packages in the Readme. Is that a good spot for them, and are those summaries sufficient? PR would close #1488


NickCraver:
The additions are appreciated! I'd say they're going to make the table haphazardly sized and hard to read though - I suggest we create a bulleted list below it to say what each package below is perhaps?

RSid:
Sounds good to me, I'll reformat!

NickCraver (approved):
Looking good - thank you!

PR #1723, opened 2021-11-03
Don't specify dbtype for DateTime et al; fixed npgsql 6 issue

mgravell:
Primary change: allow scenarios where we **do not** specify the dbtype on a parameter, but instead let the provider take control; this was causing problems with some scenarios, especially in npgsql 6

additional changes:

- npgsql 6 uses case-sensitive parameter names by default: fix tests accordingly
- (temporary) switch to the current daily of npgsql 6, for netfx fix (else everything npgsql breaks)

---

Note: in some paths, we *used* to generate code that called a helper `GetDbType` method (passing the value), and use that to set the param's `DbType`. With the new change, we'd have to also now add a branch to test for null, and rebalance the stack etc afterwards, so: instead, we have replaced `GetDbType` with a new `SetDbType` that *takes the parameter* (and the value), and does *all* the work internally, so we don't need to emit complex IL (and: "DRY"). This helper method was public, but explicitly hidden and marked "do not use" (via `[Obsolete]`) - not worried about a break there.

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -155,6 +155,12 @@ namespace Dapper
         /// </summary>
         public bool RemoveUnused { get; set; }
 
+        internal static bool ShouldSetDbType(DbType? dbType)
+            => dbType.HasValue && dbType.GetValueOrDefault() != EnumerableMultiParameter;
+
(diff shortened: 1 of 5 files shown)


NickCraver (approved):
:+1: Looking good, and cleaner

PR #1735, opened 2021-12-17
Enhance literal tokens capabilities

echofool:
Enhance `Literal Tokens` capabilities, now we can use `Literal Tokens` as a suffix of table for `Table Sharding`.
before:
```csharp
connection.Query("select * from User_34 where Id= {=Id}", new { TableId = 5201314%64 , Id = 5201314 });
```
after:
```csharp
connection.Query("select * from User_{=TableId} where Id= {=Id}", new { TableId = 5201314%64 , Id = 5201314 });
```


mgravell:
This is an interesting one. I'm always a little twitchy when it comes to SQL rewriting, but: the existing feature exists. What I'd really like to remember (and can't) is what scenarios the existing prefix exclusion was there to care about. Sadly there are no tests for that. It is also interesting that we don't seem to care about any similar exclusion *after* the token.

In principle, I'm fine with this. I just want to try to think of edge conditions where we need to care more about what comes before (or possibly after?)

PR #1753, opened 2022-02-06
support for enum type of db /支持 数据库 enum

liuweizzuie:
hello there, 

here we have an entity like this: 
```
    public enum ContentType
    {
        TXT,

        MD,

        HTML
    }

    [Table("card")]
    public class Card: BaseEntityUInt64
    {
        [Column("uid")]
        public ulong UID { get; set; }

        [Column("create_at")]
        public DateTime CreateAt { get; set; }

        [Column("title")]
        public string Title { get; set; }

        [Column("content")]
        public string Content { get; set; }

        [Column("content_type")]
        public ContentType ContentType { get; set; }

        public Card()
        {
            this.Title = string.Empty;
            this.Content = string.Empty;
            this.ContentType = ContentType.TXT;
        }
    }
```
and the table defines like :
```
CREATE TABLE `card` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT COMMENT 'ID服务中心获取ID，以便将来分库',
  `uid` bigint(20) DEFAULT NULL,
  `create_at` datetime DEFAULT NULL,
  `title` varchar(128) DEFAULT NULL,
  `content` text,
  `content_type` enum('TXT','MD','HTML') NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `id_UNIQUE` (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=11 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
```
we want to support enum type for database,  via TypeHandler: 
```
    public class EnumTypeHandler<T> : SqlMapper.TypeHandler<T> where T : struct, IConvertible
    {
        static EnumTypeHandler()
        {
            if (!typeof(T).IsEnum)
            {
                throw new ArgumentException("T must be an enumeration type");
            }
        }

        public override T Parse(object value)
        {
            return (T)Enum.Parse(typeof(T), Convert.ToString(value));
        }

        public override void SetValue(IDbDataParameter parameter, T value)
        {
            parameter.DbType = DbType.AnsiString;
        }
    }
```

we use 
`SqlMapper.AddTypeHandler(typeof(ContentType), new EnumTypeHandler<ContentType>());`

but it doesn't work.

so I forked the repository and made a little change.

please consider this.

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -367,7 +367,7 @@ namespace Dapper
             handler = null;
             var nullUnderlyingType = Nullable.GetUnderlyingType(type);
             if (nullUnderlyingType != null) type = nullUnderlyingType;
-            if (type.IsEnum && !typeMap.ContainsKey(type))
+            if (type.IsEnum && !typeMap.ContainsKey(type) && !typeHandlers.ContainsKey(type))
             {
                 type = Enum.GetUnderlyingType(type);
             }


liuweizzuie:
sorry for the confict , i will fix it

liuweizzuie:
I reopen this issue , for this commit
https://github.com/DapperLib/Dapper/pull/2060

liuweizzuie:
I reopen this issue , for this commit
https://github.com/DapperLib/Dapper/pull/2060

Issue #1754, opened 2022-02-08
Ability to handle nulls in a custom type handler


jonnymuir:
In https://sodocumentation.net/dapper/topic/13/handling-nulls it states that nulls are handled as follows:

"values read that are null are presented as null, or (in the case of mapping to a known type) simply ignored (leaving their type-based default)"

Propose using the same pattern as ITypeHandlerParse to implement a INullTypeHandler to allow us to handle a null rather than being ignored. If not implemented this will then have no effect and therefore be safe / backwards compatible / no risk and have no runtime performance overhead.

The scenario this is useful for is if a user wants to differentiate between:
- A column being present in the result set but null
- A column not being present in a result set

An example of this would be implementing an Optional pattern

Below is a .net fiddle to demonstrate.  
[DotNetFiddle example](https://dotnetfiddle.net/ph3TQK)

jonnymoo:
**Summary of pull request:**

Allow a developer to implement an additional interface (INullTypeHandler) on a custom type handler.

**Performance assessment:**

Performance overhead = 1 check for INullTypeHandler per mapped ITypeHandler type on generation of IL.
No performance overhead at runtime of IL (because the generated IL is the same if the interface isn't implemented)
If Interface is implemented the same pattern as ITypeHandlerParse is used

**Impact assessment:**

Developer is required to implement a new interface. No change to any existing behaviours.

jonnymoo:
Hi good dapper people - we have been using this now in development and UAT since March and are about to put it into live. It is a really useful addition for us. Dapper is by far the best fit for our projects and this addition is critical for our solution. 

It would be great it this addition could be a candidate to be a pull requested, I think there may be others that find it useful and I think it fits with the ethos of the project (performance, simplicity and quality!)

Is there anything I can do to make it more attractive? Contribute to documentation etc?

Ta.

PR #1756, opened 2022-02-12
Issue #1754 resolution - Allow INullTypeHandler to allow parsing of nulls on  custom type handlers

jonnymoo:
This change allows a developer to implement an additional interface (INullTypeHandler) on a custom type handler if they want to be able to handle the parsing of nulls (as opposed to the default behaviour of ignore nulls).

This has been implemented as an interface rather than extending the behaviour of SqlMapper.Settings.ApplyNullValues. This is to make it fully backwards compatible for existing code and ensure it doesn't impact performance.

There is an example of existing behaviour: [Existing behaviour example](https://dotnetfiddle.net/ph3TQK)

Example of how to implement desired behaviour is in the supplied unit tests.

Apologies - I can't see where to update documentation, if someone could point me in the right direction I will happily contribute to the documentation too.

Changes:

- Unit tests to show desired behaviour and ensure backwards compatible behaviour
- Create INullTypeHandler interface
- Change LoadReaderValueOrBranchToDBNullLabel to check for implementation of INullTypeHandler
- Alter GenerateDeserializerFromMap to produce IL which will use the null handler if required

Performance assessment:

- Performance overhead = 1 check for INullTypeHandler per mapped ITypeHandler type on generation of IL.
- No performance overhead at runtime of IL (because the generated IL is the same if the interface isn't implemented)
- If Interface is implemented the same pattern as ITypeHandlerParse is used

Impact assessment:

- Developer is required to implement a new interface. No change to any existing behaviours. Unit test supplied to uphold this.

### DIFF
--- a/Dapper/SqlMapper.ITypeHandler.cs
+++ b/Dapper/SqlMapper.ITypeHandler.cs
@@ -25,5 +25,18 @@ namespace Dapper
             /// <returns>The typed value</returns>
             object Parse(Type destinationType, object value);
         }
+
+        /// <summary>
+        /// Implement this interface along side ITypeHandler if you want nulls to be passed to your custom value parsing
+        /// </summary>
+        public interface INullTypeHandler
+        {
+            /// <summary>
+            /// Parse a database back to a typed value (or null if not required)
+            /// </summary>
+            /// <param name="destinationType">The type to parse to</param>
+            /// <returns>The typed value</returns>
+            object ParseNull(Type destinationType);
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


jonnymoo:
Hi good dapper people - we have been using this now in development and UAT since March and are about to put it into live. It is a really useful addition for us. Dapper is by far the best fit for our projects and this addition is critical for our solution.

It would be great it this addition could be a candidate to be a pull requested, I think there may be others that find it useful and I think it fits with the ethos of the project (performance, simplicity and quality!)

Is there anything I can do to make it more attractive? Contribute to documentation etc?

Ta.

PR #1758, opened 2022-03-01
upgraded dependencies and added querylimit to commanddefinition

dioptre:
Can now set a byte limit to interrupt the query response (for large queries)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ Dapper.Tests/SqlServerTypes/
 BenchmarkDotNet.Artifacts/
 .idea/
 .DS_Store
+tests/Dapper.Tests/mono_crash.*.json
--- a/Dapper.EntityFramework.StrongName/Dapper.EntityFramework.StrongName.csproj
+++ b/Dapper.EntityFramework.StrongName/Dapper.EntityFramework.StrongName.csproj
@@ -4,7 +4,7 @@
     <Title>Dapper: Entity Framework type handlers (with a strong name)</Title>
     <Description>Extension handlers for entity framework</Description>
     <Authors>Marc Gravell;Nick Craver</Authors>
-    <TargetFrameworks>net461</TargetFrameworks>
+    <TargetFrameworks>net48</TargetFrameworks>
     <AssemblyOriginatorKeyFile>../Dapper.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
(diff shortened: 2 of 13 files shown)


mgravell:
K, let's take a step back here. What exact problem are we trying to solve? Initial thoughts:

1. Changing TFM (and in particular removing TFMs) is a non-trivial step that can have significant consequences; not impossible, but if a specific TFM can't support a feature, it is often preferably to *add an additional TFM*, so that a feature can be made available when possible, but not be available on the down-level TFM (which remains available)
2. Anything that combines the current thread with async - as seen in the GC stuff - may be prone to odd behaviour; threads are not tightly bound when awaiting

There's also questions as to impact of the new tests, which needs to be weighed against what we're trying to achieve. So: to go back to the start: what are we trying to achieve here?

mgravell:
If the driver here is intended to be to avoid large queries; two simpler solutions present themselves:

- use non-buffered mode and interrupt the query at the consumer (noting that this isn't suitable for async currently)
- limit the rows returned by the query itself

dioptre:
Currently there's no way of limiting a 1TB query result until it does the 1TB query. This PR interrupts the query at say 100M.

dioptre:
Yes this overcomes the limit: 
```

    use non-buffered mode and interrupt the query at the consumer (noting that this isn't suitable for async currently)
    limit the rows returned by the query itself
```

mgravell:
If we were to add anything here: a "max rows" property seems a lot simpler, easier to implement, easier to understand and predict, etc? And it wouldn't require any TFM changes...

dioptre:
- This fix allows interruption in async and sync
   - In theory 1GB sized rows would lead to 1TB dataset in 1000 rows (this
   is exaggerated but is a common problem for dapper users)

mgravell:
Another part of me is thinking "if you type `while (true) {}`, bad things happen" - very true, and the most common fix is: "so... don't do that?".

dioptre:
The max rows didn't work for us. Seems it's the same for other users on
stack exchange too...

dioptre:
Yeah our product is dependent on the unknown. Can't stop users from making
wrong decisions!

dioptre:
That's pretty easy to do with pagination, or a count + query? We have very diverse and unknown data, where query time due to the unknown service (Azure) or unknown size per row become a non trivial limit for scalability - I'm sure this will effect a lot of other projects too. No harm in not accepting the PR as it's here for others to use if they wish!

dioptre:
Yeah we found row wasn’t accurate enough for our use case - and I’m
expecting the GC call is almost a no-op as the system should be keeping
tally on memory…

I didn’t love the TypeSize*row (List<T>) estimate as it wasn’t accurate at
all especially for long nvarchar(max) types. The only way we’ve seen we can
accurately protect our clusters is with query size.

NickCraver:
@dioptre I think if you choose to use this path that's perfectly fine, and we wish you well. For the reasons already laid out here, it's not something we'd want to put into the library. If you'd like to link people to a fork here: awesome, but for this PR itself I think we safely say this isn't something for mainstream and the changes themselves (e.g. to TFM) are extremely breaking and not something we want to do.

dioptre:
Makes sense!

For posterity and people trying to find help in the future might add some
keywords:
query size, limits, query memory exceeds machine size, limit, timeout,
command, execution timeout, cluster, scale out dapper, scalability issues,
query timeout, times out, limit, dapper hangs, overflow

PR #1762, opened 2022-03-20
Add option to persist enum names

### LINKED ISSUE #813: Store Enum Value as String
I have a simple enum just like this:

```
public enum Foo
{
  First,
  Second,
  Third
}
```

When saving in the dabase, the value are stored as the indexed position of the enum value.
Is possible to store an enum property as string in the database?

bfriesen:
When the new `PersistEnumsByName` setting is true, an enum is persisted by its name (`Enum.ToString()`) instead of by its numeric value.

This is sort of an alternative solution to issue #259, but it doesn't directly address that issue by changing how custom type handlers are used. Instead, it adds a new setting, `PersistEnumsByName`, which I believe addresses what most users actually want - to be able to persist an enums by their name. When this setting is false (its default value), the current behavior is completely unchanged. Admittedly, there is no granularity with how enums are handled; you couldn't have some enums persisted by name and others by their underlying value. Supporting that scenario would add significant complexity, and I don't think there is enough demand to justify it.

Fixes #813.

### DIFF
--- a/Dapper/SqlMapper.Settings.cs
+++ b/Dapper/SqlMapper.Settings.cs
@@ -99,6 +99,11 @@ namespace Dapper
             /// instead of the original name; for most scenarios, this is ignored since the name is redundant, but "snowflake" requires this.
             /// </summary>
             public static bool UseIncrementalPseudoPositionalParameterNames { get; set; }
+
+            /// <summary>
+            /// Whether to persist an enum by its name (<see cref="Enum.ToString()"/>) instead of by its underlying numeric value.
+            /// </summary>
+            public static bool PersistEnumsByName { get; set; }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


joostmeijles:
Nice improvement! Is there an ETA for this to get into a release?

nyan-cat:
> Nice improvement! Is there an ETA for this to get into a release?

Hello from 2023 eve. There are a bunch of same Dapper enum issues opened for 8 years, and nobody gives a damn about it. I'm thinking of forking the library, merging commits above into it, and publish to NuGet so everybody could finally use a mature version of this ORM 😂

SeppPenner:
Will this ever be in a release?

docs/index.md
SeppPenner, line 1: Just a small thing: There's a typo: `underltying` should be `underlying` I guess ;)

bfriesen: Whoops!

PR #1786, opened 2022-06-16
Adds MatchConstructorParametersWithUnderscores

### LINKED ISSUE #818: DefaultTypeMap.MatchNamesWithUnderscores is ignored for ctor parameters

jo-goro:
Resolves #818.

Adds the static `MatchConstructorParametersWithUnderscores` property for `DefaultTypeMap` which allows matching columns containing underscores to constructor parameters (eg. `user_id` matches `userId`).

This commit also renames `MatchNamesWithUnderscores` to `MatchFieldsAndPropertiesWithUnderscores` for clarity. `MatchNamesWithUnderscores` is marked as obsolete and simply maps to `MatchFieldsAndPropertiesWithUnderscores`.

### DIFF
--- a/Dapper/DefaultTypeMap.cs
+++ b/Dapper/DefaultTypeMap.cs
@@ -74,7 +74,8 @@ namespace Dapper
                 int i = 0;
                 for (; i < ctorParameters.Length; i++)
                 {
-                    if (!string.Equals(ctorParameters[i].Name, names[i], StringComparison.OrdinalIgnoreCase))
+                    var denseName = MatchConstructorParametersWithUnderscores ? names[i].Replace("_", "") : names[i];
+                    if (!string.Equals(ctorParameters[i].Name, denseName, StringComparison.OrdinalIgnoreCase))
                         break;
                     if (types[i] == typeof(byte[]) && ctorParameters[i].ParameterType.FullName == SqlMapper.LinqBinary)
                         continue;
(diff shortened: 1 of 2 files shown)


oliverlevay:
Merge this pleeeaaaasse

AdamTovatt:
Please merge this

b2soft:
Would be great if merger, thanks

mgravell:
that might mean that the preferred option is for `GetConstructorParameter` to attempt an exact match first, and then and only then `if (MatchNamesWithUnderscores) { /* normalize and retry */ }` ?

b2soft:
> that might mean that the preferred option is for `GetConstructorParameter` to attempt an exact match first, and then and only then `if (MatchNamesWithUnderscores) { /* normalize and retry */ }` ?

For me that looks logical. Leave the same variable name, and just exact match + check underscores if setting is true. But I'm open to other options as well.

AdamTovatt:
I just don't want to have to use underscores for the parameter names in my constructors.

If it is implemented as the default behavior without a specific property for just constructors or if it is implemented with a specific property for constructors doesn't matter to me, I think both works.

I think you are saying that it is unnecessary to add the extra property when it could be the default behavior (given the existing `MatchNamesWithUnderscores`), which I agree with. I just assumed it was written that way to avoid changing the behavior of already existing code. I trust that someone smart will make a smart decision here, as long as I don't have to use snake case in my constructors I'm happy :)

mgravell:
Yep; I'll take a look ASAP and see if I can validate the behaviour *without* this setting; if it is what I think, I'll take on the work to tweak the PR (preserving credit) to remove the extra properties, which I *believe* we can side step.

mgravell:
I've validated to my satisfaction that we can add this *without* the additional settings, since any scenario affected currently fails as:

> System.InvalidOperationException : A parameterless default constructor or one matching signature ({signature}) is required for {typename}

I'll see about tweaking the PR

mgravell:
This PR combined and extended in #1947

PR #1801, opened 2022-07-10
Fixed case sensitivity issues.

olivier-spinelli:
This fixes a case sensitive issues.
Instead of using the FactUnlessCaseSensitiveDatabase attribute, I think it's better to fix the test itself.


NickCraver (approved):
Looking good, thanks!

Issue #1104, opened 2018-08-24
Auto Generation of Interface Proxies?


tippmar:
AutoMapper has a wonderfully useful feature where it can auto-generate a "proxy" class for an interface - allowing you to specify a mapping *to* an IFoo. AutoMapper generates the proxy on the fly and reuses it on future calls.

Any chance Dapper could be enhanced to do the same thing? I have a zillion private concrete classes implementing my DTO public interfaces, and the *only* reason those exist is to make Dapper happy - my repositories expose only public interfaces, and it would be a beautiful thing if I could 

            return await _dbConn.QueryAsync<IFoo>(query, new {fooId});

PR #1802, opened 2022-07-10
Support for Type mapping.

olivier-spinelli:
This adds support for Type mapping, typically resolving the issue #1104 by allowing a requested type to be mapped to another type. The core of this PR is rather light since everything is done in the TypeDeserializerCache:

When a type is new to the cache, instead of considering only the requested type, a mapping function is called and if it provides a type, both the requested and mapped types are cached. 

Currently, no check is done (with `IsAssignableFrom`) to enforce the relationship between the requested and mapped types
but this can easily be done. Not sure it's useful though...

### DIFF
--- a/Dapper/SqlMapper.TypeDeserializerCache.cs
+++ b/Dapper/SqlMapper.TypeDeserializerCache.cs
@@ -43,7 +43,15 @@ namespace Dapper
                         found = (TypeDeserializerCache)byType[type];
                         if (found == null)
                         {
-                            byType[type] = found = new TypeDeserializerCache(type);
+                            var mapped = SqlMapper.abstractTypeMap?.Invoke(type);
+                            if( mapped != null && mapped != type )
+                            {
+                                byType[type] = byType[mapped] = found = new TypeDeserializerCache(mapped);
+                            }
+                            else
+                            {
+                                byType[type] = found = new TypeDeserializerCache(type);
+                            }
                         }
                     }
                 }
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -427,6 +427,55 @@ namespace Dapper
             return DbType.Object;
         }
 
+        private static Func<Type, Type> abstractTypeMap;
+
+        /// <summary>
+        /// Gets the current abstract to concrete mapper (can be null).
+        /// Use <see cref="AddAbstractTypeMap(Func{Func{Type, Type}, Func{Type, Type}})"/> to combine
+        /// it with a any new mapping. This function must simply return null (or the type itself) if
+        /// the type has no mapping or must not be mapped.
+        /// </summary>
+        /// <remarks>
+        /// Once a type has been mapped, it will keep its original mapping until <see cref="PurgeQueryCache"/> is called.
+        /// </remarks>
+        public static Func<Type, Type> CurrentAbstractTypeMap => abstractTypeMap;
+
+        /// <summary>
+        /// Updates <see cref="CurrentAbstractTypeMap"/> with a new one that should combine the 
+        /// current one with any new mappings in a thread safe manner.
+        /// </summary>
+        /// <remarks>
+        /// The <paramref name="combiner"/> may be called more than once in case of concurrent calls.
+        /// </remarks>
+        /// <param name="combiner">A function that must combine its input with any rules and returns a new mapper.</param>
+        public static void AddAbstractTypeMap(Func<Func<Type, Type>, Func<Type, Type>> combiner)
+        {
+            var spinWait = new SpinWait();
+            while( true )
+            {
+                var current = abstractTypeMap;
+                if( Interlocked.CompareExchange(ref abstractTypeMap, combiner(current), current) == current )
+                {
+                    return;
+                }
+                spinWait.SpinOnce();
+            }
+        }
+
+        /// <summary>
+        /// Sets the <see cref="CurrentAbstractTypeMap"/> to a new function, regardless of its current 
+        /// value. Use <see cref="AddAbstractTypeMap(Func{Func{Type, Type}, Func{Type, Type}})"/> to safely
+        /// combine a new mapper with the current one.
+        /// </summary>
+        /// <remarks>
+        /// Once a type has been mapped, it will keep its original mapping until <see cref="PurgeQueryCache"/> is called.
+        /// </remarks>
+        /// <param name="map">The new mapping function to set. Null to reset it.</param>
+        public static void SetAbstractTypeMap(Func<Type, Type> map)
+        {
+            abstractTypeMap = map;
+        }
+
         /// <summary>
         /// Obtains the data as a list; if it is *already* a list, the original object is returned without
         /// any duplication; otherwise, ToList() is invoked.
--- a/tests/Dapper.Tests/AbstractTypeMappingTests.cs
+++ b/tests/Dapper.Tests/AbstractTypeMappingTests.cs
@@ -1,0 +1,95 @@
+﻿using System;
+using System.Data;
+using System.Linq;
+using Xunit;
+
+namespace Dapper.Tests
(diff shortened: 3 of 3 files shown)


mgravell:
I think we need to check, here, what the actual *feature* is that we're enabling - the "why", not the "how". Only then can we assess whether this is an appropriate implementation of that "why". You cite #1104, but a key part of that is the generation aspect; that seems like something we could now do very easily over in "AOT". Config APIs should usually be a last resort - what is the common scenario where this would be used, and can we make that easier? Would static factory methods on the interface (where allowed by the runtime) be an option?

Can you elaborate on the perceived main use case for the feature?

olivier-spinelli:
Hello,

Glad that this PR caught the attention!
The "why" is rather basic: I have a IThing at compile time, only at runtime do I have a Thing class that implements the interface. This PR simply enables such mapping to be known and handled by Dapper. 

Whether such mapping can be triggered by an _automagic_ pattern is always possible (including code generation) and I fully agree that when AOT generation is possible, this _configuration_ may be done "more genuinely". However, in my case, the implementation class is not generated in the project/package that declares the IThing but by the final application (IThing is unified with other `ISpecializedThing : IThing` that can exist in other - orthogonal - packages). 

This is one of the reason "why" I choose this implementation. Another reason was that I tried to follow the current design principles of Dapper and tries to minimize the impact as much as possible (and I think that the result is bearable), avoiding reflection/discovery/assembly scans and without platform restrictions...  We use this for years now without any issues. 

I've not yet explored the C#11 "static interface methods". It seems to me that the "strange recurring pattern" (the `Interface<TSelf>`) would unfortunately be required on the top of my hierarchy of types (named `IPoco`) to have the `static abstract TSelf CreateInstance()` be correctly typed. But... since Dapper uses reflection afterwards, it seems that a `static abstract IPoco CreateInstance()` may do the job. I've not yet put a lot of thought on this feature... However here, it may well be an alternative to the _configuration way_...

olivier-spinelli:
BTW, I just realized that you may have addressed this bug: https://github.com/DapperLib/Dapper/issues/1822 because I see now a wrapper around the reader (but the issue is still opened).

I also fixed it  here: https://github.com/DapperLib/Dapper/pull/1824
The explanation is here: https://github.com/DapperLib/Dapper/issues/1822#issuecomment-1254715993.
This fix doesn' use systematically a wrapper (note that the `GetSameReaderForSameShape` test - and its invariant - was preserved).

mgravell:
Right; so - separately, I'm having a conversation about allow factory methods; at the moment we support constructors only, but there is a good case for factories, and there is a PR in AOT that adds support. That prompted the question: should Dapper allow factories? I wonder if this is an example of where a factory might be ideal; consider the following:

``` c#
public interface IFoo // your POCO... POIO?
{
    [SomeAttribute] // from Dapper, TBA
    public static IFoo Create()
    {
        // your logic here; your own project-specific
        // type registry/lookup/whatever
        return new DummyFoo();
    }
}

// just for the example
class DummyFoo : IFoo { }
```

If Dapper recognized this attribute, *everything works*; a quick test shows that this is usable:

``` c#
var factory = (from method in typeof(IFoo).GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static)
               where method.IsStatic && Attribute.IsDefined(method, typeof(SomeAttributeAttribute))
               select method).SingleOrDefault();

if (factory is not null)
{
    var obj = factory.Invoke(null, null);
    Console.WriteLine("Created: " + obj);
}
```

This seems a lot more flexible, and is *less* impact to Dapper - and it simply reuses a feature we were talking about adding *anyway*. Thoughts?

Note: this is framework dependent; it needs .NET 5 or later, and will not work on .NET Framework.

mgravell:
the downside is that it makes an `IFoo.Create` method available *more generally*, but: *it can be `private`*

olivier-spinelli:
I understand the gain. What annoys me is the attribute (not all `IPoco` needs to be databse aware but this concerns is not necessarily known WHERE the Poco is definde). 
Can a static named CreateInstance() that returns a type that is *assignable from* the declaring type be considered an implicit candidate IIF no ctor exist (when the declaring type is an interface).

The *assignable from* here is crucial (for me): this abstract factory will be defined at the top of the hierarchy. Only the ultimate abstraction is available. By design it WILL return an instance that is *assignable to* the actual type, but this cannot be asserted at the type system level.
 
In other words, `static object CreateInstance()` is a valid factory... This seems weird, I know, but this is the key to a truly modular system in which parts of a system interacts with types that can be refined by others: you cannot guaranty type closure "right now", you rely on the fact that types will be "eventually satisfied".

mgravell:
>  By design it WILL return an instance that is assignable to the actual type, but this cannot be asserted at the type system level.

Why can't it return the abstract type? In what valid scenario would it be returning something that *isn't at least that type*? Can you provide an example?

mgravell:
implementation note: there's also static interface methods as an alternative design idea:

``` c#
interface IBar : IFactory<IBar>
{
    static IBar IFactory<IBar>.CreateInstance() => default!;
}

public interface IFactory<T>
{
    abstract static T CreateInstance();
}
```

This presumably has the same problems as the attribute, to you?

olivier-spinelli:
It will be typed (in my case) to return the **most** abstract type (the `IPoco` hierarchy root), but the actual runtime type will necessarily be a type that is assignable to any interface that extends the interface family.

I detail below the context quickly described previously.

- Abstraction package:
```chsarp
public interface IPoco
{
     public static abstract IPoco CreateInstance();
}
```
- UserManagement package (depends on Abstraction). It defines the IUserInfo family:
```chsarp
public interface IUserInfo : IPoco
{
}
```
 - UserHasColor package (depends on UserManagement):
```chsarp
public interface IUserWithColor : IUserInfo
{
}
```
- UserHasFirstName package (also depends on UserManagement but doesn't know UserWithColor at all):
```chsarp
namespace UserHasFirstName;
public interface IUserInfo : UserManagement.IUserInfo
{
     public string FirstName { get; set; }
}
```
 - In AcmeCorp users must have a color and a first name: the project depends on UserHasColor and 
   UserHasFirstName.
   The "Post Code Generation" generates an implementation that unifies the family (there is much more automatic code of 
   course in real life, for instance non nullable fields, collections, etc. are recursively initialized):
 ```csharp
public sealed class IUserInfo_CK: UserManagement.IUserInfo, UserHasColr.IUserWithColor, UserHasFirstName.IUserInfo
{
   public string Name {get;set;}
   public int Color {get;set;} 
   public string FirstName {get;set;} 
}
``` 
This (awful) class is out of reach from any user code.

Any code, in any package, can always safely require the `IPocoFactory<IUserInfo>`, or `IPocoFactory<IUserWithColor>` or any member of the family that is known at its level from the DI (the generated factory implementation is a singleton that works for all the members of the family):
```csharp
var uInfo = services.GetRequiredService<IPocoFactory<IUserWithColor>>().Create();
// or from the PocoDirectory.
var uInfo = services.GetRequiredService<PocoDirectory>().Create<IUserWithColor>();
````
The developper focuses on its Poco, there is absolutely no boilerplate. This example introduces only 2 family members. In practice much more refinements exists.

That being said, back to your interrogation:

The `IFactory<T>` would be needed in the user code, and repeated for each member of the family (can we call it "`TSelf` pollution"?). 
Each typed `CreateInstance()` would have to be generated (this is easlity doable) but today, only one exposed generic 
methods is required that simply casts the `object` in `T : where T : class, IPoco`.

I totally understand your concern. Note that I'v always considered the design I propose as a low level, advanced, feature. This is why I exposed the thread safe "combiner" as the only API: this is not inteded to be used by mundane developers but by architects/lead devs who understand what they do :). In my mind, this would have been used by the infrastructure of a code base (just like I do), not by everyone (how much Dapper users configure the `SqlMapper` mapping capabilities?).

PR #1816, opened 2022-08-11
When there are too many parameters, parameterized insertion has performance problems.

0611163:
Used for batch insert by one SQL statement.

issue: [https://github.com/DapperLib/Dapper/issues/1817](https://github.com/DapperLib/Dapper/issues/1817)

Myself Test Engineering：https://github.com/0611163/Dapper.LiteSql --> PostgreSQLTest --> public void Test2InsertList() --> conn.Execute(strSql.ToString(), ToDynamicParameters(parameters), _tran)

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -170,8 +170,6 @@ namespace Dapper
         {
             var literals = SqlMapper.GetLiteralTokens(identity.sql);
 
-            HashSet<string> cleanNames = new HashSet<string>();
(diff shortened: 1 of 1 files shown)


0611163:
Used batch insert by one SQL statement..

0611163:
After my tested, the command.parameters.contains method in Npgsql.dll version 5.0.14 and earlier has performance problem. But the Npgsql.dll latest version 6.0.6 has no problem.

PR #1868, opened 2022-12-27
add: tests for CockroachDB

dikshant:
This PR adds tests for CockroachDB. The tests are largely borrowed from the existing PostgreSQL tests to ensure basic functionality. [Due to the way SERIAL](https://www.cockroachlabs.com/docs/stable/serial.html) primary keys work in CockroachDB some tests had to be removed because they were no longer deterministic.

The official CockroachDB docker image [has issues](https://github.com/cockroachdb/cockroach/discussions/83749) starting in single node mode in Actions CI, so an [alternate image](https://github.com/cockroachlabs-field/cockroachdb-single-node) had to be used.

Finally, I was running into issues with .NET framework versions missing for 3.x and 5.x so I modified the CI to include those versions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,13 +1,9 @@
 name: Main Build
 
 on:
-  pull_request:
   push:
     branches:
     - main
-    paths:
-    - '*'
-    - '!/docs/*' # Don't run workflow when files are only in the /docs directory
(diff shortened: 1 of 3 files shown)


dikshant:
Closing this because based it in wrong branch.

PR #1869, opened 2022-12-27
add: tests for CockroachDB

dikshant:
This PR adds tests for CockroachDB. The tests are largely borrowed from the existing PostgreSQL tests to ensure basic functionality. [Due to the way SERIAL](https://www.cockroachlabs.com/docs/stable/serial.html) primary keys work in CockroachDB some tests had to be removed because they were no longer deterministic.

The official CockroachDB docker image https://github.com/cockroachdb/cockroach/discussions/83749 has trouble starting in single node mode in Actions CI, so an [alternate image](https://github.com/cockroachlabs-field/cockroachdb-single-node) had to be used.

Finally, I was running into issues with .NET framework versions missing for 3.x and 5.x so I modified the CI to include those versions.

You can see that the tests pass on my fork here:
https://github.com/dikshant/Dapper/actions/runs/3790538626/jobs/6445231573

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,10 @@ jobs:
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: test
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      cockroachdb:
+        image: timveil/cockroachdb-single-node:latest
+        ports:
+          - 26257:26257
       sqlserver:
         image: mcr.microsoft.com/mssql/server:2019-latest
         ports:
@@ -38,8 +42,15 @@ jobs:
           MYSQL_ROOT_PASSWORD: root
           MYSQL_DATABASE: test
     steps:
+
     - name: Checkout code
       uses: actions/checkout@v1
+    - name: .NET setup
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: |
+          3.1.x
+          5.0.x
     - name: .NET Build
       run: dotnet build Build.csproj -c Release /p:CI=true
     - name: Dapper Tests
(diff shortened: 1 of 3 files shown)


dikshant:
Ah it seems the CI/CD has moved to Appveyor. I don't see a way to setup CockroachDB through the AppVeyor CI, although I do see configuration for Postgres and MYSQL. Any help there would be appreciated.

.github/workflows/main.yml
```diff
@@ -23,6 +23,10 @@ jobs:
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: test
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      cockroachdb:
+        image: timveil/cockroachdb-single-node:latest
```
NickCraver, line 27: Let's try and use the official image here please - everything else looks good at a quick glance, but we'd want to rely on the official image in the builds.

.github/workflows/main.yml
```diff
@@ -38,8 +42,15 @@ jobs:
           MYSQL_ROOT_PASSWORD: root
           MYSQL_DATABASE: test
     steps:
+
```
NickCraver, line 45:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -38,8 +42,15 @@ jobs:
...
     - name: Checkout code
       uses: actions/checkout@v1
+    - name: .NET setup
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: |
+          3.1.x
+          5.0.x
```
NickCraver, line 53: Dapper needs a lot of updating love we haven't had time for - this _shouldn't_ be needed, but needs to be paired with a net6.0 update _on test projects only_ (not changing core library targets). We can do that in this PR or separately.

Issue #1239, opened 2019-04-13
QueryAsync missing buffered parameter


vitidev:
I can

`db.Query<Person>(sql, parameters, buffered: false)`

But I can't 

`db.QueryAsync<Person>(sql, parameters, buffered: false)`

At the same time, the multimap overload of QueryAsync has parameter "buffered"

`Task<IEnumerable<TReturn>> QueryAsync<TReturn>(this IDbConnection cnn, string sql, Type[] types, Func<object[], TReturn> map, object param = null, IDbTransaction transaction = null, bool buffered = true, `

mgravell:
Well that's odd. Is there an overload that takes the struct? Maybe that has
it?

But: when async iterators lands in c# vNext, we'll need to revisit the
non-buffered query-async *anyway*, as it is only pseudo-async at the moment
(by necessity).

vitidev:
@mgravell 
Yes. I found 
`QueryAsync<T>(this IDbConnection cnn, CommandDefinition command) `

But it was expected that synchronous and asynchronous versions will have "mirror" signatures. Sometimes you have to convert Query into QueryAsyncand it's very nice when it's enough to add Async(+await) and that is all

>as it is only pseudo-async at the moment

That all the same it is good, it is possible not to hold a current thread waiting for the response from a database

>when async iterators lands in c# vNext

Async enumerators are good, but a little incomprehensible why you can not do without them like

     var reader = await db.ExecuteAsyncReaderAsync(); // 
     while (await reader.ReadAsync())
     {
           var value = reader.GetChar(0);
     }

     var reader = await db.QueryReaderAsync<Person>();
     while (await reader.MoveNext())
     {
          var value = reader.Current;
     }

Or is it already there and I do not read the documentation well?

mgravell:
You absolutely *can* do that as long as you're happy for the consuming code
to do that work but a execute-reader and getting the per-T materializer
separately. The point here is that until C# vNext, neither "foreach" nor
"yield return" support this scenario, so you can't *just* return a
sequence. That's what vNext gives us: "await foreach".

ravimp:
@mgravell await foreach is out there now. Any plans to support this now?

ravimp:
@mgravell, Dapper has been such a help, I'd like to return favor. I think I can contribute if someone points me in right direction. Meanwhile, I got this from an SO post. Do you think it is a reliable workaround until we get an API for it?

```
[...]

PR #1882, opened 2023-02-06
Issue #1239 resolution - QueryAsync missing buffered parameter

SoftStoneDevelop:
This change resolves https://github.com/DapperLib/Dapper/issues/1239.

I had to switch to net48;netstandard2.1 because of the Microsoft.Bcl.AsyncInterfaces and System.Linq.Async packages, but I think it's worth it.

### DIFF
--- a/Dapper.EntityFramework.StrongName/Dapper.EntityFramework.StrongName.csproj
+++ b/Dapper.EntityFramework.StrongName/Dapper.EntityFramework.StrongName.csproj
@@ -4,7 +4,7 @@
     <Title>Dapper: Entity Framework type handlers (with a strong name)</Title>
     <Description>Extension handlers for entity framework</Description>
     <Authors>Marc Gravell;Nick Craver</Authors>
-    <TargetFrameworks>net461</TargetFrameworks>
+    <TargetFrameworks>net462</TargetFrameworks>
     <AssemblyOriginatorKeyFile>../Dapper.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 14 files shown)


SoftStoneDevelop:
It would be nice to add an asynchronous connection close to the method:
```csharp
private static async IAsyncEnumerable<T> QueryAsync<T>(this IDbConnection cnn, Type effectiveType, CommandDefinition command)
```

But I didn’t want to do something like TryOpenAsync, so I made a proposal in .NET to create an IAsyncDbConnection interface:
https://github.com/dotnet/runtime/issues/81723

mgravell:
Right; so: I don't think we can take this as-is; there's *way* too much API breakage here, and honestly, taking `buffered` as a parameter here is just a bad choice, as buffered and unbuffered are fundamentally incompatible shapes

IMO the correct change here is *just* the addition of a single API:

``` c#
public static IAsyncEnumerable<T> QueryUnbufferedAsync<T>(this DbConnection cnn, string sql, object param = null, DbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
```

key points here:

- there's no point pretending `IDbConnection` will work; it won't; that was a bad API choice in the past
- no `buffered` parameter

if you care to refactor to *just* the above, we should be able to take that, but I *suspect* it may be easier to start from a clean slate and just re-apply the interesting bits; alternatively I'm happy to take this and re-do from scratch

mgravell:
on the TFM change: I don't see why we need `System.Linq.Async` - if we just use `Microsoft.Bcl.AsyncInterfaces`, we get to keep all the TFMs IIRC

mgravell:
this feature now essentially provided via #1912; I propose we close this and re-evaluate any missing requirements when that is merged

Issue #1886, opened 2023-03-10
PetaPoco performance tests fail when running on net6.0


verrigo:
Hello,
When running `PetaPoco` performance test on `net6.0`, instead of a successful run, we get an exception:
`System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidProgramException: Common Language Runtime detected an invalid program.
   at System.Runtime.CompilerServices.RuntimeHelpers.CompileMethod(RuntimeMethodHandleInternal method)
   at System.Reflection.Emit.DynamicMethod.CreateDelegate(Type delegateType)
   at PetaPoco.Database.PocoData.GetFactory[T](String key, Boolean ForceDateTimesToUtc, IDataReader r) in /_/benchmarks/Dapper.Tests.Performance/PetaPoco/PetaPoco.cs:line 1248
   at PetaPoco.Database.Fetch[T](String sql, Object[] args) in /_/benchmarks/Dapper.Tests.Performance/PetaPoco/PetaPoco.cs:line 472
   at Dapper.Tests.Performance.PetaPocoBenchmarks.FetchFast() in /_/benchmarks/Dapper.Tests.Performance/Benchmarks.PetaPoco.cs:line 38
   at BenchmarkDotNet.Autogenerated.Runnable_0.WorkloadActionUnroll(Int64 invokeCount) in /_/benchmarks/Dapper.Tests.Performance/bin/Release/net6.0/5531dc99-89b7-4dd0-97ef-7fcc70a75721/5531dc99-89b7-4dd0-97ef-7fcc70a75721.notcs:line 950
   at BenchmarkDotNet.Engines.Engine.RunIteration(IterationData data)
   at BenchmarkDotNet.Engines.EngineFactory.Jit(Engine engine, Int32 jitIndex, Int32 invokeCount, Int32 unrollFactor)
   at BenchmarkDotNet.Engines.EngineFactory.CreateReadyToRun(EngineParameters engineParameters)
   at BenchmarkDotNet.Autogenerated.Runnable_0.Run(IHost host, String benchmarkName) in /_/benchmarks/Dapper.Tests.Performance/bin/Release/net6.0/5531dc99-89b7-4dd0-97ef-7fcc70a75721/5531dc99-89b7-4dd0-97ef-7fcc70a75721.notcs:line 167
   --- End of inner exception stack trace ---
[...]

PR #1887, opened 2023-03-11
Added support for net462 and net6.0/net7.0 

verrigo:
Dropped support for net461 and net5.0 as they are out of service
Added support for net462 and net6.0. And for Dapper and Dapper.Strongname also for net7.0
Fixed issue https://github.com/DapperLib/Dapper/issues/1886
Added possibility to debug the Benchmark Project from an IDE
Moved appveyor image up to Visual Studio 2022


mgravell:
This doesn't "add support" for net6/net7 - they were already perfectly well supported. All this does is *remove* support for anyone using net5 - or at least, pushes them to use the netstandard2.0 build.

Now, I can *sort of* get behind removing the obsolete ones and migrating to current, but there's no point having both net6 and net7 **unless** there is a difference between them, usually via `#if`. But I'll need to review on a PC (on mobile at the moment) to see more.

mgravell:
I actually missed the important bit "Fixed issue #1886" on first review. Honestly, I think that should be a separate PR, rather than bundling everything all together.

mgravell:
The petapoco change: is this simply an update of current code from the petapoco repo? If so, fine - but: we shouldn't be "fixing" petapoco here

verrigo:
> is this simply an update of current code

Yes, you are (of course) right. I can remove net7 and upate the pull request. Also remove the add support for net6 and mention that net461 and net5 will no longer be supported since they are out of support

verrigo:
> 

So, there are two petapoco changes: One is removing the old petapoco package and using instead petapoco.compiled, since the current one is obsolete and the other change is in the PocoData class, where the change fixed issue https://github.com/DapperLib/Dapper/issues/1886

NickCraver:
I think there's often a disconnect between "out of support" and "aren't being used" - those are not the same thing. What do we gain by dropping the support in the library? If there's no tangible benefit to dropping those working-as-is consumers, we err on the side of not doing that.

verrigo:
I saw an Isuse created in 2019 where someone asked for net461 to be added again because of a situation where they could not upgrade to a newer version of visual studio. Reconsidering this PR, it looks like I didn't think it properly.

NickCraver:
Going to close this out to tidy up - the removal isn't tenable for support on the breaking front.

Issue #1909, opened 2023-06-08
Announcement/Input - DapperAOT progress and Dapper API / feature update


mgravell:
There's a lot of preamble and back-story here. If you're short of time, scroll down to "Call to action".

---

You would be forgiven for thinking

> The devs aren't iterating Dapper any more, is it dead?

I'm happy to say that the answer is an emphatic "no, it isn't dead", and "yes, we are still here" - and while ADO.NET isn't quite as central to our current roles, it underpins a lot of the teams that *use* the things that we  (@mgravell and @NickCraver) work on.

So: I've been a little hesitant to invest heavily in extending Dapper in its current form, for a few reasons, mostly related to metaprogramming. A lot of my thoughts are [captured here](https://blog.marcgravell.com/2021/05/is-era-of-reflection-heavy-c-libraries.html). The key impact of this is:.

- hard to debug at runtime 
- lots of runtime work, both initially (emitting strategies) and each time (strategy cache lookups, metadata verification, etc)
- hard to develop, with few people able to contribute
- in particular, hard to extend async code, since ref-emit and async are not good friends
- expensive to maintain - a "small" change can have huge impact to the ref-emit code, and cause instability
- absolutely categorically does not work with AOT or other ref-emit constrained runtimes

So: I've wanted to improve that. Initially I spent a while looking at a "reimagining of Dapper" that used the extended partial methods support in C# 9? 10? and "generators", so that an analyser understood the *intent* of your code (which looked very different to Dapper today), and wrote the missing pieces. It worked, but it was inelegant, and frankly "just use this completely different API" is a terrible story re adoption.

But last week, I had an epiphany. I learned about "iterators", which are hopefully going to be released in net8. Iterators are a new C# vNext feature that allow additional code (typically via a generator) to say "that method call there? yeah, just ignore where that's going - come *here* instead" - effectively, it allows code to retarget what arbitrary method calls do.

So how is this relevant to Dapper? Well, if an analyzer can find your `connection.QueryAsync<Customer>(...)` call, it can spit out an interceptor for *that specific call* (or multiple call-sited if it wants), that is a method that *achieves the same result* (probably forwarding the same arguments, unless it wants to optimize), but using an API that allows *generated* code to deal with all the parameter packing and row parsing. To make a long story short ("too late!"), it makes your existing Dapper code work without requiring *any* ref-emit at runtime, with *zero* ongoing strategy cache checks, in a fully AOT way.

[...]

PR #1910, opened 2023-06-09
add support for fetching values via GetFieldValue<T>

### LINKED ISSUE #1907: No way to store big numbers with a lot of decimal spaces (without storing it in the database as text)
We have MSSQL databases that have a column typed as numeric(30,15).

The database works with a Java application (which has the column used as BigDecimal in its entity).

Now I'm trying to make a C# microservice that puts a number into the same column, but I'm striking some limits here.

The numeric type in C# has a precision of only 29 and we have 30, which means that numeric cannot be used and .NET / C# doesn't have an equivalent to Java's BigDecimal yet.

So far I have tried the Extended.Numerics.BigDecimal NuGet package, but that doesn't work with Dapper.

Trying to insert it as a string gives me: 
```
Arithmetic overflow error converting nvarchar to data type numeric.
      The statement has been terminated.
```

How to test: Try storing 999999999999999.999999999999999 into an MSSQL numeric(30,15) column. Works in Java with Hibernate, can't do it in C# with Dapper. :'(

Why such big numbers? Chemical measurements.

The kitty is sad.

:max_bytes(150000):strip_icc():focal(899x0:901x2)/21042210_264995290674140_8840525631411191808_n-530848c0d1134a31bc03861ea9ddd700.jpg)

mgravell:
this requires `DbDataReader`, but in reality the reader is *always* `DbDataReader`; acknowledge this, and stop pretending to use `IDataReader` internally

The primary use for this is things like `SqlDecimal` which **cannot** be accessed other than via `DbDataReader.GetFieldValue<T>` or via `SqlDataReader.GetSqlDecimal`

fix #1907

the most relevant parts are:

- for things like `Query<SqlDecimal>`, the addition of `ReadViaGetFieldValueFactory` which is a per-T cached *factory* method that works a lot like the old logic (now enshrined in `ReadViaGetValue`), but using `IsDbNull` and `GetFieldValue<T>`
- the addition of `LoadReaderValueViaGetFieldValue`, which reads a value in IL via `IsDbNull` and `GetFieldValue<T>`; the use of `IsDbNull` means that we won't have anything to pop on the stack (unlike the old with did a read, dup, test, branch - leaving the original copy on the stack if we branched), which is tracked via `popWhenNull`

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -197,7 +197,7 @@ namespace Dapper
 
         static SqlMapper()
         {
-            typeMap = new Dictionary<Type, TypeMapEntry>(21)
+            typeMap = new Dictionary<Type, TypeMapEntry>(41)
             {
                 [typeof(byte)] = DbType.Byte,
                 [typeof(sbyte)] = DbType.SByte,
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -8,6 +8,8 @@ using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Data;
+using System.Data.Common;
+using System.Data.SqlTypes;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
@@ -31,7 +33,7 @@ namespace Dapper
         {
             public int Compare(PropertyInfo x, PropertyInfo y) => string.CompareOrdinal(x.Name, y.Name);
         }
-        private static int GetColumnHash(IDataReader reader, int startBound = 0, int length = -1)
+        private static int GetColumnHash(DbDataReader reader, int startBound = 0, int length = -1)
         {
             unchecked
             {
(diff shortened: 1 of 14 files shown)


Dapper/SqlMapper.cs
```diff
@@ -2970,12 +3063,41 @@ private static IDataReader ExecuteReaderImpl(IDbConnection cnn, ref CommandDefin
...
+                    .MakeGenericMethod(type));
+                lock (s_ReadViaGetFieldValueCache)
+                {
+                    s_ReadViaGetFieldValueCache[type] = factory;
+                }
+            }
+            return factory(index);
         }
+        // cache of ReadViaGetFieldValueFactory<T> for per-value T
+        static readonly Hashtable s_ReadViaGetFieldValueCache = new Hashtable();
```
NickCraver, line 3094: Are we worried about unbounded growth on the factory cache here?

mgravell: Great question, but honestly: no; this is only used for things seen in the type-map, so this is no different than "are we worried about unbounded growth of the type-map if people keep calling `AddTypeMap` with lots of types" - entirely self inflicted (and will never happen)

NickCraver: Yeah that sounds fair 👍. You already taking a look at the test stuff? Happy to fix up GitHub runs here shortly if not - thinking we bump to 6.x on the suite but may need further love. I looked over the rest, looking good!

mgravell:
Emphasis: this PR proves everything I am saying in #1909 - maintaining IL emit code is **hard**

Dapper/SqlMapper.cs
NickCraver: See bottom of PR :) Ubuntu doesn't ship with .NET 5.x, but changing suite to net6.0 should resolve

mgravell:
all CI is green @NickCraver

PauloHMattos:
@mgravell This should also close https://github.com/DapperLib/Dapper/issues/1644?

PR #1912, opened 2023-06-09
add missing QueryUnbufferedAsync<T> API

mgravell:
I am *not* proposing to add the myriad of multi-read `<T0, T1, ..., T5>` APIs that go with this

### DIFF
--- a/Dapper/SqlMapper.Async.cs
+++ b/Dapper/SqlMapper.Async.cs
@@ -102,7 +102,7 @@ namespace Dapper
         /// A sequence of data of <typeparamref name="T"/>; if a basic type (int, string, etc) is queried then the data from the first column is assumed, otherwise an instance is
         /// created per row, and a direct column-name===member-name mapping is assumed (case insensitive).
         /// </returns>
-        public static IAsyncEnumerable<T> QueryUnbufferedAsync<T>(this IDbConnection cnn, string sql, object param = null, IDbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
+        public static IAsyncEnumerable<T> QueryUnbufferedAsync<T>(this DbConnection cnn, string sql, object param = null, DbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
         {
             // note: in many cases of adding a new async method I might add a CancellationToken - however, cancellation is expressed via WithCancellation on iterators
             return QueryUnbufferedAsync<T>(cnn, typeof(T), new CommandDefinition(sql, param, transaction, commandTimeout, commandType, CommandFlags.None, default));
--- a/Dapper/SqlMapper.Async.cs
+++ b/Dapper/SqlMapper.Async.cs
@@ -5,6 +5,7 @@ using System.Data;
 using System.Data.Common;
 using System.Globalization;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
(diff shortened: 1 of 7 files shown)


mgravell:
I think this provides the core features wanted in #1692 and #1239, while being simpler

plaisted:
I see a new release for this but nothing published to nuget, is this planned to be pushed?

Dapper/SqlMapper.Async.cs
```diff
@@ -1217,5 +1218,79 @@ private static async Task<T> ExecuteScalarImplAsync<T>(IDbConnection cnn, Comman
...
+                        {
+                            yield break;
+                        }
+                        tuple = info.Deserializer = new DeserializerState(hash, GetDeserializer(effectiveType, reader, 0, -1, false));
+                        if (command.AddToCache) SetQueryCache(identity, info);
+                    }
+
+                    var func = tuple.Func;
+
+                    var convertToType = Nullable.GetUnderlyingType(effectiveType) ?? effectiveType;
```
0xced, line 1275: This `convertToType` variable is not used. I guess it was meant to be used 4 lines below?

```cs
yield return GetValue<T>(reader, convertToType, val);
```

mgravell: You're almost certainly correct; thanks, good eyes! I'll try to look at this tomorrow (I guess I need some tests that do `QueryUnbuffered<int>`, `QueryUnbuffered<int?>`, `QueryUnbuffered<string>`, `QueryUnbuffered<SomeValueType>` and `QueryUnbuffered<SomeClass>`

0xced: Or maybe it should just be deleted because apparently the `convertToType` is already computed inside `GetValue<T>(…)`.

mgravell: added for tracking: https://github.com/DapperLib/Dapper/issues/1920

0xced: > good eyes

Actually, it's JetBrains Rider dimming unused variables. 😉

image

Dapper/SqlMapper.Async.cs
```diff
@@ -1217,5 +1218,79 @@ private static async Task<T> ExecuteScalarImplAsync<T>(IDbConnection cnn, Comman
...
+        /// <param name="sql">The SQL to execute for the query.</param>
+        /// <param name="param">The parameters to pass, if any.</param>
+        /// <param name="transaction">The transaction to use, if any.</param>
+        /// <param name="commandTimeout">The command timeout (in seconds).</param>
+        /// <param name="commandType">The type of command to execute.</param>
+        /// <returns>
+        /// A sequence of data of <typeparamref name="T"/>; if a basic type (int, string, etc) is queried then the data from the first column is assumed, otherwise an instance is
+        /// created per row, and a direct column-name===member-name mapping is assumed (case insensitive).
+        /// </returns>
+        public static IAsyncEnumerable<T> QueryUnbufferedAsync<T>(this DbConnection cnn, string sql, object param = null, DbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
```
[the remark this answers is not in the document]
NickCraver: Async doesn't work on IDbConnection, it never has. We should have done this with the old methods but can't break things now - IDbConnection simply predates async being a thing, so it never had support. The methods needed are on the abstract base class.

PR #1917, opened 2023-06-13
Re-enable Firebird tests (Issue178_Firebird)

0xced:


0xced:
Well, it looks like running Firebird in a docker container on AppVeyor is [not as easy](https://help.appveyor.com/discussions/problems/25465-pipedocker_engine-access-denied) as it ought to be. 😣

mgravell (approved):

mgravell:
I'm fine with it, but appveyor says "nope"

0xced:
Unfortunately I'm not familiar enough with AppVeyor.

There's this [//./pipe/docker_engine access denied?](https://help.appveyor.com/discussions/problems/25465-pipedocker_engine-access-denied) discussion on the AppVeyor forum but it doesn't seem to really offer a solution.

Alternatively, Firebird could probably be downloaded, configured and executed with scripts, [like nhibernate-core does](https://github.com/nhibernate/nhibernate-core/blob/90d50c03475628d10336cbad368f071375056fef/appveyor.yml#L42-L62). But that's too much YAML and PowerShell for me. 😅

mgravell:
Me too :)

I think this is one of those things where if it is going to fight us, it is going to stay disabled (unless @NickCraver has some magic CI-jitsu to employ)

NickCraver:
Indeed this isn't something we can really pull in unless the platform already runs the server, we just don't have the test environment from it. Even in say GitHub Actions we have another version of the problem: we can't build and test .NET Framework (Windows) because service containers can only run on Ubuntu. Clients testing against server bits isn't something well supported on most test platforms, unfortunately :(

PR #1923, opened 2023-06-19
Change the argument of QueryUnbufferedAsync from a concrete type to an interface

xin9le:
# Summary
Dapper.NET primarily provides extension methods for `IDbConnection`. However, the recently added `QueryUnbufferedAsync` method follows the form of an extension method for `DbConnection`, which lacks consistency. This Pull-Request aims to enhance consistency and convenience by changing the argument from a concrete type to an interface type.

### DIFF
--- a/Dapper/SqlMapper.Async.cs
+++ b/Dapper/SqlMapper.Async.cs
@@ -1234,7 +1234,7 @@ namespace Dapper
         /// A sequence of data of <typeparamref name="T"/>; if a basic type (int, string, etc) is queried then the data from the first column is assumed, otherwise an instance is
         /// created per row, and a direct column-name===member-name mapping is assumed (case insensitive).
         /// </returns>
-        public static IAsyncEnumerable<T> QueryUnbufferedAsync<T>(this DbConnection cnn, string sql, object param = null, DbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
+        public static IAsyncEnumerable<T> QueryUnbufferedAsync<T>(this IDbConnection cnn, string sql, object param = null, IDbTransaction transaction = null, int? commandTimeout = null, CommandType? commandType = null)
         {
             // note: in many cases of adding a new async method I might add a CancellationToken - however, cancellation is expressed via WithCancellation on iterators
             return QueryUnbufferedAsync<T>(cnn, typeof(T), new CommandDefinition(sql, param, transaction, commandTimeout, commandType, CommandFlags.None, default));


mgravell:
This was not accidental; the reality is that the async APIs *do not work* on things that aren't `DbConnection`, so: let's stop lying - I'm more concerned with obviousness and correctness

drasticactions:
> This was not accidental; the reality is that the async APIs _do not work_ on things that aren't `DbConnection`, so: let's stop lying - I'm more concerned with obviousness and correctness

@mgravell Is there a reason for this? It seems strange that an interface type couldn't work for that API since it would be on me (the person writing to that interface) to ensure it works correctly if I'm using it. Why would it not work?

@xin9le was not accusing you of anything, nor did they imply any "wrongdoing," so your tone seems very strong for someone making an earnest attempt at contributing.

mgravell:
The async APIs don't exist on the interface. Interfaces are closed for extension (at least in pre-DIM platforms) and pre-date async. The abstract base classes are open for extension.

drasticactions:
I apologize for my ignorance. I'm confused by this.

https://github.com/DapperLib/Dapper/blob/02c92d99c4445f37b15b723bdde9ff3180d47950/Dapper/SqlMapper.Async.cs#L26

This is using the IDbConnection interface and it's Async. As far as I can tell, all of the others do too, apart from `QueryUnbufferedAsync,` which then _does_ use the Interface version inside of the private function. 

image

So I'm unclear what "Interfaces are closed for extension" means in this context. If I pass through a `DbConnection` into that public static method, it will go into a private static method that uses `IDbConnection`. Are there specific functions in `DbConnection` not in `IDbConnection`?

drasticactions:
Reading through the documentation for [DbConnection](https://learn.microsoft.com/ja-jp/dotnet/api/system.data.common.dbconnection?view=net-7.0) and [IDbConnection](https://learn.microsoft.com/ja-jp/dotnet/api/system.data.idbconnection?view=net-7.0), and reading the rest of the file, I think I understand what's happening here.

https://github.com/DapperLib/Dapper/blob/02c92d99c4445f37b15b723bdde9ff3180d47950/Dapper/SqlMapper.Async.cs#L382-L389

`IDbConnection` does not have Async methods, but the abstract class `DbConnection` does. The code above checks if the `IDbConnection` is a `DbConnection` and, if not, throws. Since this extension method is not in `netstandard` and is new (So, not a breaking change), you've elected to use `DbConnection` directly. Since you already know that the command being passed in is a `DbConnection` could you change:

https://github.com/DapperLib/Dapper/blob/02c92d99c4445f37b15b723bdde9ff3180d47950/Dapper/SqlMapper.Async.cs#LL1261C44-L1261C64

To be `DbConnection`, directly invoke the API and avoid the reflection since you know that's the only thing it could be.

Is this correct? If so, maybe we could write documentation for it to explain that. That would help anyone who sees this going forward why this is the case.

PR #1928, opened 2023-06-26
attempt NRT makeover

mgravell:
Functionally, the only real changes here are in a few API surfaces (not a real change, but NRT changes things slightly):

NOTE: the intention is to do a minor rev here, i.e. ship this as 2.1.0

- type handlers take and return `?`
- `{Query|Read}{First|Single}OrDefault[Async]` return `?`
- `ExecuteScalar[Async]` return `?`
- update libs
- package `readme.md` (this is the current NuGet recommendation)
- enable `[SkipLocalsInit]` everywhere; simplify the remoting warning logic

for Dapper.AOT purposes (although everything here applies to any consumer), I'm also applying some tweaks to make `GridReader` more usable from externals; `GridReader` is not `sealed`, but was mostly unusable from custom implementations - added some things:
  - `protected` constructor
  - switched to `Action<object?>?`/`object?` callback rather than type-specific (so external consumers aren't bound to same types)
  - added `protected int OnBeforeGrid()` (pre-grid book-keeping)
  - added `protected void OnAfterGrid(int)` and `protected Task OnAfterGridAsync(int)` (post-grid book-keeping)
  - made all existing usage consistently use that `OnBeforeGrid`/`OnAfterGrid` API
  - exposed the reader and cancellation-token via `protected` APIs
  - allow the `Identity` to be constructed on-the-fly if/when needed

(the NRT issue was also partly driven from AOT)


mgravell:
@NickCraver  I've added API proof to this; do you have an opinion on this? otherwise, ima merge

NickCraver:
@mgravell I can this afternoon or tonight - but if not doing a release and this is blocking, happy to comment after the merge too and we can tidy to parallel - your call!

mgravell:
Not immediately blocking; however it is a pain to have to keep rebasing it (which is why it is heavily squashed and force-pushed), as it touches *a lot*

Dapper/CustomPropertyTypeMap.cs
```diff
@@ -29,13 +29,13 @@ public CustomPropertyTypeMap(Type type, Func<Type, string, PropertyInfo> propert
...
         public ConstructorInfo FindConstructor(string[] names, Type[] types) =>
-            _type.GetConstructor(Array.Empty<Type>());
+            _type.GetConstructor(Array.Empty<Type>())!;
 
         /// <summary>
         /// Always returns null
         /// </summary>
         /// <returns></returns>
-        public ConstructorInfo FindExplicitConstructor() => null;
+        public ConstructorInfo FindExplicitConstructor() => null!;
```
NickCraver, line 38: This...seems like a lie?

Dapper/DefaultTypeMap.cs
```diff
@@ -29,22 +29,22 @@ public DefaultTypeMap(Type type)
 
         internal static MethodInfo GetPropertySetter(PropertyInfo propertyInfo, Type type)
```
NickCraver, line 30: Not sure how it interacts with the direct emits, but feels like this is more correct as `MethodInfo?`

Dapper/CustomPropertyTypeMap.cs
```diff
@@ -29,13 +29,13 @@ public CustomPropertyTypeMap(Type type, Func<Type, string, PropertyInfo> propert
         /// <param name="types">DataReader column types</param>
         /// <returns>Default constructor</returns>
         public ConstructorInfo FindConstructor(string[] names, Type[] types) =>
```
NickCraver, line 31: FWIW we mismatch here on DefaultTypeMap vs. this for nullability - intentional?

Dapper/DefaultTypeMap.cs
```diff
@@ -127,7 +127,7 @@ public ConstructorInfo FindExplicitConstructor()
         /// <returns>Mapping implementation</returns>
         public SqlMapper.IMemberMap GetConstructorParameter(ConstructorInfo constructor, string columnName)
         {
-            ParameterInfo param = MatchFirstOrDefault(constructor.GetParameters(), columnName, static p => p.Name);
+            ParameterInfo param = MatchFirstOrDefault(constructor.GetParameters(), columnName, static p => p.Name)!;
```
NickCraver, line 130: This feels weird because if we are null we'll immediately throw downstream - should we have a null check with better messaging here?

Dapper/SqlMapper.DapperRow.Descriptor.cs
```diff
@@ -98,7 +98,7 @@ public RowBoundPropertyDescriptor(Type type, string name, int index) : base(name
                 public override object GetValue(object component)
                     => ((DapperRow)component).TryGetValue(_index, out var val) ? (val ?? DBNull.Value): DBNull.Value;
                 public override void SetValue(object component, object value)
-                    => ((DapperRow)component).SetValue(_index, value is DBNull ? null : value);
+                    => ((DapperRow)component).SetValue(_index, value is DBNull ? null! : value);
```
NickCraver, line 101: Should DapperRow accept a null? The whole nullability of DapperRow seems to represent not-nullable, but we can fail in places...I think it's really object? rather than object? in DapperRow (and we can remove the `!` here)

Dapper/SqlMapper.DapperRow.cs
```diff
@@ -46,14 +46,14 @@ internal bool TryGetValue(int index, out object value)
             {
                 if (index < 0)
                 { // doesn't exist
-                    value = null;
+                    value = null!;
```
NickCraver, line 49: I think we should represent this as nullable throughout the class, e.g. on TryGetValue, e.g. `[NotNullWhen(true)]` on `TryGetValue` and `object?` throughout - thoughts?

Dapper/SqlMapper.Link.cs
```diff
@@ -13,11 +13,10 @@ public static partial class SqlMapper
         /// <typeparam name="TValue">The value type of the cache.</typeparam>
         internal class Link<TKey, TValue> where TKey : class
         {
-            public static void Clear(ref Link<TKey, TValue> head) => Interlocked.Exchange(ref head, null);
-
-            public static bool TryGet(Link<TKey, TValue> link, TKey key, out TValue value)
+            public static void Clear(ref Link<TKey, TValue>? head) => Interlocked.Exchange(ref head, null);
+            public static bool TryGet(Link<TKey, TValue>? link, TKey key, out TValue value)
```
NickCraver, line 17: Can also use `[NotNullWhen(true)]`

Dapper/SqlMapper.cs
```diff
@@ -91,12 +91,12 @@ private static void CollectCacheGarbage()
         private static int collect;
         private static bool TryGetQueryCache(Identity key, out CacheInfo value)
```
NickCraver, line 92: `[NotNullWhen(true)]`

Dapper/SqlMapper.cs
```diff
@@ -463,7 +464,7 @@ public static void SetDbType(IDataParameter parameter, object value)
                         try
                         {
                             handler = (ITypeHandler)Activator.CreateInstance(
-                                typeof(SqlDataRecordHandler<>).MakeGenericType(argTypes));
+                                typeof(SqlDataRecordHandler<>).MakeGenericType(argTypes))!;
```
NickCraver, line 467: Should this use `(ITypeHandler?)` instead?

Dapper/SqlMapper.cs
```diff
@@ -1525,45 +1530,45 @@ private static T GetValue<T>(DbDataReader reader, Type effectiveType, object val
...
             bool wasClosed = cnn?.State == ConnectionState.Closed;
             try
             {
-                if (reader == null)
+                if (reader is null)
                 {
-                    ownedCommand = command.SetupCommand(cnn, cinfo.ParamReader);
-                    if (wasClosed) cnn.Open();
+                    ownedCommand = command.SetupCommand(cnn!, cinfo.ParamReader);
+                    if (wasClosed) cnn!.Open();
```
NickCraver, line 1563: Wrap in `cnn is not null`?

Dapper/SqlMapper.cs
```diff
@@ -2000,7 +2005,7 @@ private static Exception MultiMapException(IDataRecord reader, string splitOnCol
                         values[0] = r.GetValue(startBound);
                         if (values[0] is DBNull)
                         {
-                            return null;
+                            return null!;
```
NickCraver, line 2008: IMO this isn't accurate...but it's also viral, realistically we have a `Func<DbDataReader, object?>` throughout here - thoughts?

NickCraver:
@mgravell Tried to give good eyes here, admittedly glossed over the NRT exclusions in the Emit section as the last reviewed thing, I'll review it fresh next round (just a lot to consume in one go - nothing wrong on PR structure)

Dapper/CustomPropertyTypeMap.cs
mgravell: I... hmmm, I can't fault you. I'm curious why this is. Not why this is anything specific, just: *why this is*

Dapper/DefaultTypeMap.cs
mgravell: Yeah, I wondered about that. I was trying to avoid too many *functionally observable* changes, but it might be warranted here.

Dapper/CustomPropertyTypeMap.cs
mgravell: Almost certainly not

Dapper/SqlMapper.cs
mgravell: I think the issue there was TFMs, but I guess I can declare them as file-local or something when needed

Dapper/SqlMapper.cs
mgravell: No, because Activator.CreateInstance returns non-null for this case. Probably just `Nullable<T>` when it doesn't.

Dapper/SqlMapper.cs
mgravell: I guess I was happy for it to fail in previously observed ways if `null` happens, but: we are allowed to change that

Dapper/SqlMapper.cs
mgravell: It is a tricky one, because then it makes the core APIs things like `IEnumerable<T?> Query<T>` - which is wrong more than it is right. The exact nuance isn't expressable, sadly

Dapper/SqlMapper.cs
NickCraver: I did this in SE.Redis FWIW, think I just declared attribute based on TFM

Dapper/SqlMapper.cs
mgravell: ah, this is a "we know this but can't convince the compiler", because of:

``` csharp
bool wasClosed = cnn?.State == ConnectionState.Closed;
```

so if `wasClosed` is `true` we **know** `cnn` is not null; the additional test is genuinely redundant, and I don't think we should add it just to make the compiler happy

mgravell:
addressed all, I think

Dapper/SqlMapper.TypeHandler.cs
```diff
@@ -16,28 +16,28 @@ public abstract class TypeHandler<T> : ITypeHandler
             /// </summary>
             /// <param name="parameter">The parameter to configure</param>
             /// <param name="value">Parameter value</param>
-            public abstract void SetValue(IDbDataParameter parameter, T value);
+            public abstract void SetValue(IDbDataParameter parameter, T? value);
```
jnm2, line 19: @mgravell Since this causes implementers of the class to have to annotate their `T value` parameter as nullable, how would we _handle_ null being passed here? What would it mean?

I know for SqlParameter.Value, `DBNull` means "set to null" and `null` means "unset." Does `null` carry the same "unset" connotation if implementers see it here?

Thanks!

mgravell: Since there isn't a constraint on `T`, this doesn't strictly change the API shape - in particular, a `TypeHandler<SomeValueType>` still has `SetValue(..., SomeValueType)` and not `SetValue(..., SomeValueType?)` ; however, to answer your question: it should mean `DBNull`, but with the opportunity to configure the parameter correctly, and mostly just applies in the `TypeHandler<SomeReferenceType>` scenario, it acknowledges that the incoming parameter *might be* `null`.

Issue #565, opened 2016-07-20
List support fails with MySQL Server 5.6


bgrainger:
MySQL Server 5.6.29
Dapper 1.50.1
MySql.Data 6.9.9

Execute the following code (query adapted from https://github.com/StackExchange/dapper-dot-net#list-support):

```
var csb = new MySqlConnectionStringBuilder { Server = "localhost", UserID = "root", Password = "..." };
using (var connection = new MySqlConnection(csb.ConnectionString))
{
	connection.Open();
	connection.Query<int>("select * from (select 1 as Id union all select 2 union all select 3) as X where Id in @Ids", new { Ids = new int[0] }).ToList();
}
```

It throws a MySqlException: "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'WHERE 1 = 0)' at line 1".

For this statement, Dapper sets the CommandText to `select * from (select 1 as Id union all select 2 union all select 3) as X where Id in (SELECT @Ids WHERE 1 = 0)`

The MySql.Data connector interpolates the command parameters and rewrites this as `select * from (select 1 as Id union all select 2 union all select 3) as X where Id in (SELECT NULL WHERE 1 = 0)` before sending it to MySQL Server.

MySQL Server 5.6 rejects this as invalid SQL syntax.

Upgrading to MySQL Server 5.7.11 fixes the exception, but may not always be practical. For MySQL Server 5.6, the following query would be valid (and selects 0 rows): `select * from (select 1 as Id union all select 2 union all select 3) as X where Id in (SELECT NULL)`.

bgrainger:
Another workaround is to check `.Any()` on the list parameter in the caller before executing the query (if having 0 items in the list would trivially return 0 rows from the query).

mgravell:
@bgrainger  that would be a perfectly sensible thing to do for the calling app-code to do; not really very easy for dapper to do, because it would have to parse and understand the intent - it isn't always the case that an empty input means an empty output

bgrainger:
Yes, that's only a workaround for the caller (not Dapper) if they can infer that "0 items in list" == "0 rows returned". For a `NOT IN @ids` query, it wouldn't be applicable, for example.

bgrainger:
Another workaround that works no matter the structure of the SQL you're trying to execute.

[...]

PR #1931, opened 2023-07-05
Allow configuration of empty parameter list SQL

andrewjsaid:
Dapper is currently hardcoded to embed `(select null where 1=0)` when it detects empty list parameters. While it works for most db servers, some times that sql is invalid or sub-optimal. This change allows it to be configured by the library consumer.

This issue was discussed in #565 and I could not find a pull request which addressed it.

Thanks

### DIFF
--- a/Dapper/SqlMapper.Settings.cs
+++ b/Dapper/SqlMapper.Settings.cs
@@ -64,6 +64,7 @@ namespace Dapper
             {
                 CommandTimeout = null;
                 ApplyNullValues = false;
+                EmptyParameterListFormat = "(select null where 1=0)";
             }
 
             /// <summary>
@@ -99,6 +100,11 @@ namespace Dapper
             /// instead of the original name; for most scenarios, this is ignored since the name is redundant, but "snowflake" requires this.
             /// </summary>
             public static bool UseIncrementalPseudoPositionalParameterNames { get; set; }
+
+            /// <summary>
+            /// The SQL used when a list parameter is empty.
+            /// </summary>
+            public static string EmptyParameterListFormat { get; set; }
         }
     }
 }
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -2437,7 +2437,7 @@ namespace Dapper
                             }
                             if (first)
                             {
-                                return "(select null where 1=0)";
+                                return Settings.EmptyParameterListFormat;
                             }
                             else
                             {


mgravell:
Out of curiosity: what alternative construct is working well for your environment? I'm wondering if there's some common pattern that works similarly well everywhere, as an alternative approach

andrewjsaid:
@mgravell thanks for replying.

For our use-case of MySQL 8 `(null)` seems to work well and doesn't even match records with null fields. The following (with redacted times/rowcounts) shows the difference between the two queries.

```
EXPLAIN ANALYZE
SELECT * FROM Bills WHERE Id IN (select NULL WHERE 1=0)

-- -> Filter: <in_optimizer>(Bills.Id,<exists>(select #2))  (cost=xxxxxxxx rows=xxxxxxxx) (actual time=xxxxxxxx..xxxxxxxx rows=0 loops=1)
--     -> Table scan on Bills  (cost=xxxxxxxx rows=xxxxxxxx) (actual time=xxxxxxxx..xxxxxxxx rows=xxxxxxxx loops=1)
--     -> Select #2 (subquery in condition; dependent)
--         -> Zero rows (Impossible WHERE)  (actual time=0.000..0.000 rows=0 loops=xxxxxxxx)

EXPLAIN ANALYZE
SELECT * FROM Bills WHERE Id IN (NULL)

-- -> Zero rows (no matching row in const table)  (actual time=0.000..0.000 rows=0 loops=1)

```

Our developers often forget to check that their parameters are empty lists and that results in full table scans when it happens. We quickly just add the check after our DBAs notice perf hits but that requires hotfixing or waiting until the next working day which is not ideal.

mgravell:
Question: how does this behave:

``` sql
EXPLAIN ANALYZE
SELECT * FROM Bills WHERE 1 = 0
```

?

andrewjsaid:
```
-> Zero rows (Impossible WHERE)  (actual time=0.000..0.000 rows=0 loops=1)
```

mgravell:
so IMO that looks like a better solution to me, in that it is more universal, performs fine in any reasonable scenario, and wouldn't require configuration

andrewjsaid:
I agree! However I would need to understand how Dapper transforms SQL as we are changing scope from changing just the parameter to the clause.
e.g. previously
```sql
SELECT * FROM Bills WHERE Id IN @ids
                                ^^^^
```
and now the suggestion is to rewrite

```sql
SELECT * FROM Bills WHERE Id IN @ids
                          ^^^^^^^^^^
```

Detecting the `Id IN` clause is perhaps a much more risky change than this. However while thinking through it I have realised that
the following two queries do not result in the same data set.
```
SELECT * FROM Bills WHERE Id NOT IN (SELECT NULL WHERE 1 = 0); -- full set
SELECT * FROM Bills WHERE Id NOT IN (NULL); -- empty set
```

So maybe more thought is required.

mgravell:
> Detecting the Id IN clause is perhaps a much more risky change than this.

Agreed; that doesn't mean we should rule it out, though - although differences in SQL variants may make this undesirable and infeasible

> However while thinking through it I have realised that the following two queries do not result in the same data set.

That's a great catch, and I think means we still need to keep thinking; the "full set" is semantically the correct outcome here, and our first priority should be correctness; until we can achieve that, the topic of performance is somewhat moot; I'm not a MySQL person - does there exist a suitable variant here?

andrewjsaid:
Totally agree on correctness. Unfortunately from the online research I've done it looks like MySQL has no way to represent `IN ()` without going down the riskier route. Your examples plus another one below makes it beyond my risk appetite for a library this widely used.

```sql
SELECT * FROM Bills WHERE NOT(   Id IN (SELECT NULL WHERE 1 = 0)   ); -- full set
SELECT * FROM Bills WHERE NOT(   Id IN (NULL)   ); -- empty set, using "NULL" method would be a mistake

SELECT * FROM Bills WHERE NOT(   COALESCE(Id IN (NULL),false)   ); -- full set, exemplifies complexity of the solution
```

Maybe creating a source analyser would be the simplest route for our team. Thank you so much for your time.

jscottPTC:
We use Actian Zen, Mariadb, and Sql Server.  We have an application that communicates with all 3.  The default behavior in Dapper as of version 2.1.24 works fine with Mariadb and Sql Server but not with Actian Zen.  In Actian Zen a syntax error is thrown.  '(null)' does seem to work with all 3.  We have a temporary workaround of skipping the db call when the list is empty but it would be nice if dapper had a generic solution to this problem!

mgravell:
Ultimately, Dapper by design has limited knowledge of individual SQL variants - and I am not familiar enough with Active Zen (I've not heard of it, honestly) to even suggest what might be a *valid* construct for the same. I doubt a single "works everywhere" solution exists, but: since you have Active Zen, can you propose one that at least works for Active Zen ?

jscottPTC:
> @mgravell thanks for replying.
> 
> For our use-case of MySQL 8 `(null)` seems to work well and doesn't even match records with null fields. The following (with redacted times/rowcounts) shows the difference between the two queries.
> 
> ```sql
> EXPLAIN ANALYZE
> SELECT * FROM Bills WHERE Id IN (select NULL WHERE 1=0)
> 
> -- -> Filter: <in_optimizer>(Bills.Id,<exists>(select #2))  (cost=xxxxxxxx rows=xxxxxxxx) (actual time=xxxxxxxx..xxxxxxxx rows=0 loops=1)
> --     -> Table scan on Bills  (cost=xxxxxxxx rows=xxxxxxxx) (actual time=xxxxxxxx..xxxxxxxx rows=xxxxxxxx loops=1)
> --     -> Select #2 (subquery in condition; dependent)
> --         -> Zero rows (Impossible WHERE)  (actual time=0.000..0.000 rows=0 loops=xxxxxxxx)
> 
> EXPLAIN ANALYZE
> SELECT * FROM Bills WHERE Id IN (NULL)
> 
> -- -> Zero rows (no matching row in const table)  (actual time=0.000..0.000 rows=0 loops=1)
> ```
> 
> Our developers often forget to check that their parameters are empty lists and that results in full table scans when it happens. We quickly just add the check after our DBAs notice perf hits but that requires hotfixing or waiting until the next working day which is not ideal.

In my testing (null) seems to work for Actian Zen.

PR #1932, opened 2023-07-06
Fix issue 222: MultiMap does not return null object

Mblkolo:
Fix issue https://github.com/DapperLib/Dapper/issues/222: MultiMap does not return null object in left-outer-join if splitting on a surrogate key that is not a property on the target object.


ailtonbueno:
Approve this fix, please.
The world needs

PR #1952, opened 2023-08-23
Support $ as a parameter prefix

Giorgi:
This PR adds support for $ as a parameter prefix. I made the changes as described in this comment: [https://github.com/DapperLib/Dapper/issues/1687#issuecomment-885027948](https://github.com/DapperLib/Dapper/issues/1687#issuecomment-885027948)

I need support for $ because I'm building [ADO.NET Provider for DuckDB](https://github.com/Giorgi/DuckDB.NET) and DuckDB uses $: [Add support for named parameters in prepared statements](https://github.com/duckdb/duckdb/pull/7113)

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -2108,7 +2108,7 @@ namespace Dapper
 
         private static string GetInListRegex(string name, bool byPosition) => byPosition
             ? (@"(\?)" + Regex.Escape(name) + @"\?(?!\w)(\s+(?i)unknown(?-i))?")
-            : ("([?@:]" + Regex.Escape(name) + @")(?!\w)(\s+(?i)unknown(?-i))?");
+            : ("([?@:$]" + Regex.Escape(name) + @")(?!\w)(\s+(?i)unknown(?-i))?");
 
         /// <summary>
         /// Internal use only.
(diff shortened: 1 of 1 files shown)


mgravell:
Does the existing pseudo-positional parameter support not work? By that I mean: looking at DuckDB, it seems to use positional rather than named parameters - and the `?` isn't a prefix - it is the entire thing. Dapper can't work well with that by itself, because it doesn't know how to order things - however, Dapper supports a syntax variant:

```` sql
where Name = ?name?
```

which it rewrites as

``` sql
where Name = ?
```

but uses the token `name` to figure out which parameter to add in that position. This works with multiple providers, and it looks like it should work here, without any other changes?

mgravell:
Secondary: is there any integration test possible that would should this proposed change actually working against DuckDB? I don't know how DuckDB deploys in terms of whether it can work on the CI server, but something that can at least run on a dev box is desirable.

Giorgi:
> By that I mean: looking at DuckDB, it seems to use positional rather than named parameters

This is not correct, the upcoming version of DuckDB (0.9.0, scheduled to be released in September) will support named parameters too. This PR added support for named parameters: [Add support for named parameters in prepared statements](https://github.com/duckdb/duckdb/pull/7113)

As for how to run tests against DuckDB, DuckDB is an embedded database and you don't need to deploy anything. The simplest option is to clone the https://github.com/Giorgi/DuckDB.NET repo, checkout Named-Parameters branch and run `dotnet build /p:BuildType=Full` BuildType=Full tells the build to fetch necessary DuckDB binaries so you don't need to do anything else.

Inside the `ParameterCollectionTests.cs` file, you will find this commented assertion:

`//connection.Execute(queryStatement, new { foo = 1, bar = "test" }).Should().BeGreaterOrEqualTo(1, "Passing parameters as object should work");`

The changes in this PR make that assertion successful.

mgravell (approved):

Giorgi:
Thanks for merging the PR. Do you want me to add test cases to Dapper that use DuckDB and test the new prefix?

Giorgi:
I already have a test case for it in my repo: https://github.com/Giorgi/DuckDB.NET/commit/c5c2e5d6e593fca3ce780dfb7b69492940b7e59c

mgravell:
Ideally "yes", but the real question is: how do we *execute* those tests? What is required to spin up a duckdb instance? Can it run in CI, or is it dev machine only?

mgravell:
https://github.com/DapperLib/Dapper/releases/tag/2.1.11

PR #1970, opened 2023-10-03
Add DuckDB tests

Giorgi:
Test for #1952

The NuGet package includes native libraries too.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,6 +1,7 @@
 <Project>
     <ItemGroup>
       <!-- note: 6.2.0 has regressions; don't force the update -->
+      <PackageVersion Include="DuckDB.NET.Data.Full" Version="0.9.0" />
       <PackageVersion Include="EntityFramework" Version="6.1.3" />
       <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
       <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,7 +1,7 @@
 <Project>
     <ItemGroup>
       <!-- note: 6.2.0 has regressions; don't force the update -->
-      <PackageVersion Include="DuckDB.NET.Data.Full" Version="0.9.0" />
+      <PackageVersion Include="DuckDB.NET.Data.Full" Version="0.9.0.3" />
       <PackageVersion Include="EntityFramework" Version="6.1.3" />
       <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
       <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" />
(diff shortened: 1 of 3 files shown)


mgravell (approved):
This is probably fine - I will need to try it locally - will try to do tomorrow

mgravell:
I'm happy to merge as-is, but FYI: this doesn't actually work for netfx - I see that there's a `netstandard2.0` package in the nuget, but: it doesn't seem to work; it doesn't work in the tests here, and doesn't work in standalone packages either; example:

``` xml
<Project Sdk="Microsoft.NET.Sdk">
    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>net472</TargetFramework>
        <ImplicitUsings>enable</ImplicitUsings>
        <LangVersion>12</LangVersion>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="DuckDB.NET.Data.Full" Version="0.9.0.3"/>
    </ItemGroup>
</Project>
```

with:

``` csharp
using DuckDB.NET.Data;

using var conn = new DuckDBConnection("Data Source=:memory:");
conn.Open();
```

gives:

``` txt
Unhandled Exception: System.DllNotFoundException: Unable to load DLL 'duckdb': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
   at DuckDB.NET.NativeMethods.Startup.DuckDBOpen(String path, DuckDBDatabase& database, DuckDBConfig config, IntPtr& error)
   at DuckDB.NET.Data.Internal.ConnectionManager.GetConnectionReference(DuckDBConnectionString connectionString)
   at DuckDB.NET.Data.DuckDBConnection.Open()
   at Program.<Main>$(String[] args) in C:\Code\ConsoleApp7\ConsoleApp7\Program.cs:line 4
```

The packaging might not be right. I haven't looked to see *what* you're packaging and how - it doesn't look like there are native binaries *in the nupkg itself*, so I'm guessing that they are embedded resources in the dll? More of a topic for DuckDB, but you *might* prefer to embed them in the nupkg rather than the dll; I also haven't checked to see how the unmanaged dll works re cross-platform (normally I'd expect to see a bunch of platform-specific native dlls in the nupkg, for this purpose)

(let me know if you need any help with any of this)

mgravell:
see also:

image

where the failure is:

``` txt
DuckDB is unavailable: Unable to load DLL 'duckdb': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
```

Giorgi:
It doesn't work for netfx but works for net core. The native libraries are under `runtime` folder but netfx doesn't know how to load them: https://stackoverflow.com/questions/56154884/nuget-package-not-copying-native-dlls-on-build

Giorgi:
The native libraries are in another package that the top level depends on [DuckDB.NET.Bindings.Full](https://www.nuget.org/packages/DuckDB.NET.Bindings.Full/)

image

mgravell:
logged here: https://github.com/Giorgi/DuckDB.NET/issues/151 - I guess the key bit is "it didn't work"

mgravell:
@Giorgi  ah, it is 32-bit vs 64-bit; see 151

mgravell:
(I still can't get the tests to run in netfx, but my standalone console works :shrug:)

Issue #1971, opened 2023-10-03
Need a setting to disable unreliable OLEDB parameter check


StruanT:
The smellsLikeOleDb regex on SqlMapper doesn't work correctly in all kinds of scenarios. If you are passing a model that has properties that you don't intend to parameterize and cannot be parameterized this results in a NotSupportedException

I think this would most easily solved by having a setting turn it off on this line (just make filterParams true and ignore this regex if a setting is enabled)
https://github.com/DapperLib/Dapper/blob/47ef1e7100b7ecdb68bcf97432b9674373a6f75c/Dapper/SqlMapper.cs#L2538

Some examples that break this "smell check".

- Question marks in SQL Comments 
`-- Why is the question mark in this comment treated as a parameter ?` 

- Question marks in strings. https://github.com/DapperLib/Dapper/issues/1914

- Using openquery execute LDAP Queries that start with `<LDAP://:>;`

- Postgres type casts https://github.com/DapperLib/Dapper/issues/1299

This would all be solved by having a way to turn it off (make filterParams true and ignore this regex)

I see no reason not to have the option to disable this footgun completely for anyone not using OLEDB at all. For that matter... why not default it to off and let anyone using OLEDB turn it on?

mgravell:
I can't disagree with anything you've said. There are no scenarios when the pseudo-positional parameters (`?abc?`) wouldn't be preferable. Agree, we should:

- add a property to turn it off, but leave the default behaviour as-is for now
- add release notes advertising that this default will be changed in a future version, and that anyone using naked `?` parameters should manually set it
- release a "major" maybe start of 2024 with the default swapped, again with release notes highlighting it

Giorgi:
Should the new setting live in `SqlMapper.Settings` ?

mgravell:
Yes. I was planning to take a look when I get a moment, but I'd happily consider a PR if that's your plan.

mgravell:
https://github.com/DapperLib/Dapper/releases/tag/2.1.11

Issue #1914, opened 2023-06-11
Issue with handling '?' character in SQL strings when using SQL Server


eatmeatHJ:
I've encountered an issue when using Dapper with SQL Server. In my SQL queries, I sometimes include '?' characters inside the string values, as a part of the string itself, not as a parameter placeholder. For example, a query might look like this:

```C#
string sql = @"INSERT INTO test_table2 ([name0],[name1],[name2]) VALUES ('?a?', '3.?a?', @name)";
var dp = new Dapper.DynamicParameters();
dp.Add("name", "qq");
conn.Execute(sql, dp);
```

In this example, '?a?' and '3.?a?' are string values to be inserted into the table, and @name is the only actual parameter in the query.

However, when I execute such a query with Dapper, it throws an exception with the message: "When passing parameters by position, each parameter can only be referenced once."

Is there any way to configure Dapper to correctly handle the '?' characters in SQL strings when using SQL Server?

Any help would be greatly appreciated.

mgravell:
That is an unfortunate false positive. Sorry, right now that isn't configurable. The only immediate fix I have for you is to break the literal in your SQL, for example '?' + 'a?', '3.?' + 'a?', which most SQL implementations will recombine trivially (so no real overhead, other than ugliness)

Definitely something for us to consider for the future.

PR #1974, opened 2023-10-10
Add a setting to turn off Ole Db "smell check"

Giorgi:
Implements https://github.com/DapperLib/Dapper/issues/1971

### DIFF
--- a/Dapper/SqlMapper.Settings.cs
+++ b/Dapper/SqlMapper.Settings.cs
@@ -122,9 +122,11 @@ namespace Dapper
             }
 
             /// <summary>
-            /// If true, Dapper will try to determine if the current sql statement is ole db statement and not filter
+            /// Indicates whether single-character parameter tokens (<c>?</c> etc) will be detected and used where possible;
+            /// this feature is not recommended and will be disabled by default in future versions;
+            /// where possible, prefer named parameters (<c>@yourParam</c> etc) or Dapper's "pseudo-positional" parameters (<c>?yourParam? etc</c>).
             /// </summary>
-            public static bool OleDbCheckEnabled { get; set; } = true;
+            public static bool SupportLegacyParameterTokens { get; set; } = true;
 
             private static long s_FetchSize = -1;
         }
--- a/Dapper/SqlMapper.Settings.cs
+++ b/Dapper/SqlMapper.Settings.cs
@@ -121,6 +121,11 @@ namespace Dapper
                 }
             }
 
+            /// <summary>
+            /// If true, Dapper will try to determine if the current sql statement is ole db statement and not filter
+            /// </summary>
+            public static bool OleDbCheckEnabled { get; set; } = true;
+
             private static long s_FetchSize = -1;
         }
     }
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -2533,9 +2533,16 @@ namespace Dapper
             }
 
             bool filterParams = false;
-            if (removeUnused && identity.commandType.GetValueOrDefault(CommandType.Text) == CommandType.Text)
+            if (Settings.OleDbCheckEnabled)
             {
-                filterParams = !smellsLikeOleDb.IsMatch(identity.sql);
+                if (removeUnused && identity.commandType.GetValueOrDefault(CommandType.Text) == CommandType.Text)
+                {
+                    filterParams = !smellsLikeOleDb.IsMatch(identity.sql);
+                } 
+            }
+            else
+            {
+                filterParams = true;
             }
             var dm = new DynamicMethod("ParamInfo" + Guid.NewGuid().ToString(), null, new[] { typeof(IDbCommand), typeof(object) }, type, true);
 
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -2532,18 +2532,13 @@ namespace Dapper
                 throw new NotSupportedException("ValueTuple should not be used for parameters - the language-level names are not available to use as parameter names, and it adds unnecessary boxing");
             }
 
-            bool filterParams = false;
-            if (Settings.OleDbCheckEnabled)
+            bool filterParams = removeUnused && identity.commandType.GetValueOrDefault(CommandType.Text) == CommandType.Text;
+            
+            if (Settings.SupportLegacyParameterTokens && filterParams)
             {
-                if (removeUnused && identity.commandType.GetValueOrDefault(CommandType.Text) == CommandType.Text)
-                {
-                    filterParams = !smellsLikeOleDb.IsMatch(identity.sql);
-                } 
+                filterParams = !smellsLikeOleDb.IsMatch(identity.sql);
             }
-            else
-            {
-                filterParams = true;
-            }
+            
             var dm = new DynamicMethod("ParamInfo" + Guid.NewGuid().ToString(), null, new[] { typeof(IDbCommand), typeof(object) }, type, true);
 
             var il = dm.GetILGenerator();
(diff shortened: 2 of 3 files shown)


Dapper/SqlMapper.cs
```diff
@@ -2533,9 +2533,16 @@ internal static IList<LiteralToken> GetLiteralTokens(string sql)
             }
 
             bool filterParams = false;
-            if (removeUnused && identity.commandType.GetValueOrDefault(CommandType.Text) == CommandType.Text)
+            if (Settings.OleDbCheckEnabled)
```
mgravell, line 2536: it is important that we don't change the behaviour in unrelated cases; this *looks* like it is now going to filter sprocs, which would be very bad; suggest simplifying to just:

``` diff
- filterParams = !smellsLikeOleDb.IsMatch(identity.sql);
+ filterParams = !(Settings.OleDbCheckEnabled && smellsLikeOleDb.IsMatch(identity.sql);
```

(please double-check my boolean logic)

Dapper/SqlMapper.Settings.cs
```diff
@@ -121,6 +121,11 @@ public static long FetchSize
                 }
             }
 
+            /// <summary>
+            /// If true, Dapper will try to determine if the current sql statement is ole db statement and not filter
+            /// </summary>
+            public static bool OleDbCheckEnabled { get; set; } = true;
```
mgravell, line 127: not loving this name; naming is hard; maybe `SupportLegacyParameterTokens` ?

also, more comments would help here, I think; perhaps:

``` txt
Indicates whether single-character parameter tokens (<c>?</c> etc) will be detected and used where possible;
this feature is not recommended and will be disabled by default in future versions; where possible, prefer named
parameters (<c>@yourParam</c> etc) or Dapper's "pseudo-positional" parameters (<c>?yourParam? etc</c>).
```

mgravell (changes requested):
a few comments inline; needs some kind of test; this is a little awkward because of concurrency concerns, but we should at least try

Dapper/SqlMapper.Settings.cs
Giorgi: Yeah, naming is hard

Dapper/SqlMapper.cs
Giorgi: Even in the case of legacy parameter tokens, shouldn't the regex be evaluated only in the case of `CommandType.Text`

Giorgi:
I think it now should work correctly in both cases but I'm not sure how to test it.

Dapper/SqlMapper.cs
mgravell: yes, but I was referring to the value of `filterParams` - with this setting disabled, it would have changed from `false` to `true` in the original version, no?

Dapper/SqlMapper.cs
```diff
@@ -2532,11 +2532,13 @@ internal static IList<LiteralToken> GetLiteralTokens(string sql)
                 throw new NotSupportedException("ValueTuple should not be used for parameters - the language-level names are not available to use as parameter names, and it adds unnecessary boxing");
             }
 
-            bool filterParams = false;
-            if (removeUnused && identity.commandType.GetValueOrDefault(CommandType.Text) == CommandType.Text)
+            bool filterParams = removeUnused && identity.commandType.GetValueOrDefault(CommandType.Text) == CommandType.Text;
+            
+            if (Settings.SupportLegacyParameterTokens && filterParams)
```
mgravell, line 2537: nit; can we check the local first?

Dapper/SqlMapper.cs
Giorgi: It would filter for sprocs too which is now fixed.

Dapper/SqlMapper.cs
Giorgi: Sorry, I'm not sure I follow. Check what?

mgravell:
Re testing; IIRC in the original ticket there was an example where it did the wrong thing; can we not just use something based on that to show it continuing to fail when set to true and working when set to false? The only caveat is that we should use try/finally to set the value back to *what it was originally* after the test

Dapper/SqlMapper.cs
mgravell: Swap the order of the two operands -`if (filterParams && ...)`

Giorgi: Sure, will do. Out of curiosity, why does it matter?

Giorgi:
I tried https://github.com/DapperLib/Dapper/issues/1914 and it still throws the same exception.

Dapper/SqlMapper.cs
mgravell: it isn't huge, hence "nit" - but checking a local in the current stack-frame is faster than pretty much anything else; now, compared to going out-of-process and probably off-box for the actual SQL: this is nothing - we've spent more time talking about it here than will *ever* be saved by the change, but... meh, I guess I just prefer it that way around, despite there not being an objectively demonstrable "this will genuinely help" :p

mgravell:
> I tried #1914 and it still throws the same exception.

well yes, but that is a different glitch; I wouldn't expect that to change, so it is good that it didn't - however, we should be able to validate with a scenario more like #1971 - for example:

``` c#
        [Theory]
        [InlineData(true)]
        [InlineData(false)]
        public void OleDbParamFilterFails(bool yourNewSettingHere)
        {
            var oldValue = SqlMapper.Settings.YourNewSettingHere;
            try
            {
                SqlMapper.Settings.YourNewSettingHere = yourNewSettingHere;

                if (yourNewSettingHere) // OLE DB parameter support enabled; can false-positive
                {
                    Assert.Throws<NotImplementedException>(() => GetValue(connection));
                }
                else // OLD DB parameter support disabled; more reliable
                {
                    Assert.Equal("this ? could be awkward", GetValue(connection));
                }
            }
            finally
            {
                SqlMapper.Settings.YourNewSettingHere = oldValue;
            }

            static string GetValue(DbConnection connection)
                => connection.QuerySingle<string>("select 'this ? could be awkward'",
                    new TypeWithDodgyProperties());
        }
        class TypeWithDodgyProperties
        {
            public string Name => throw new NotSupportedException();
        }
```

I haven't executed that at all, note

mgravell (approved):

mgravell:
looks good; can you add a release-note entry? for an example see index.md here: https://github.com/DapperLib/Dapper/commit/a37b151bb745ed0763441aa50f29c504221aa7d3#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6

mgravell:
also, can you confirm you're free and willing to contribute this code irrevocably etc in line with the project license, i.e. "don't sue me later"

mgravell:
Note: re the pseudo-positional parameter support (#1914) - that's one that Dapper.Analyzer will definitely be helping with ASAP; the problem is that we don't want to have to constantly parse SQL, so for `@foo` and `?foo?`, at runtime we kinda have to do things the "probably" way; however, Dapper.Analyzer can take the time at build to do a full parse, find the *actual* parameters, and offer a warning and advice on workarounds. So if that one is hurting you: worry not!

Giorgi:
> looks good; can you add a release-note entry? for an example see index.md here: [a37b151#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6](https://github.com/DapperLib/Dapper/commit/a37b151bb745ed0763441aa50f29c504221aa7d3#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6)

Under the `### unreleased` section?

Giorgi:
> also, can you confirm you're free and willing to contribute this code irrevocably etc in line with the project license, i.e. "don't sue me later"

I confirm

mgravell:
Merged with thanks

PR #1975, opened 2023-10-10
auto-detect stored procedures as anything without whitespace

mgravell:
also, we have public fields? (hangs head in shame)

### DIFF
--- a/Dapper/CommandDefinition.cs
+++ b/Dapper/CommandDefinition.cs
@@ -2,7 +2,6 @@
 using System.Data;
 using System.Reflection;
 using System.Reflection.Emit;
-using System.Text.RegularExpressions;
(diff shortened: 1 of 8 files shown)


kirylbokach:
Unfortunately this change breaks all single word commands, like `COMMIT` and `ROLLBACK`

mgravell:
@kirylbokach https://github.com/DapperLib/Dapper/pull/1995 thanks!

crayzd92:
This change also breaks the `REVERT` command

mgravell:
K, will fix

SSMKittel:
Also affects the "analyze" command in postgres. https://www.postgresql.org/docs/current/sql-analyze.html

PR #1980, opened 2023-10-13
change NRT for ITypeMap.GetConstructorParameter

### LINKED ISSUE #1969: Should `ITypeMap.GetConstructorParameter` be allowed to return null?
We're currently using a custom type map like this:

```csharp
internal class RemoveSpacesFromColumnNames : SqlMapper.ITypeMap
{
    private readonly Type _type;

    public RemoveSpacesFromColumnNames(Type type)
    {
        _type = type;
    }

    public ConstructorInfo FindConstructor(string[] names, Type[] types) =>
        _type.GetConstructors().Single();

    public ConstructorInfo? FindExplicitConstructor() => null;

    public SqlMapper.IMemberMap? GetConstructorParameter(ConstructorInfo constructor, string columnName)
    {
        var nameWithoutSpaces = columnName.Replace(" ", "");
        var parameter = constructor.GetParameters().FirstOrDefault(x => x.Name == nameWithoutSpaces);
        return parameter == null ? null : new CustomParameterMap(columnName, parameter);
    }

    public SqlMapper.IMemberMap? GetMember(string columnName) => null;
}
```

where we'd added the nullable reference annotations ourselves. (Hopefully this code doesn't look too crazy :)

However when updating to dapper 2.1.4 this code now fails to compile since `GetConstructorParameter` is expected to return `SqlMapper.IMemberMap` instead of `SqlMapper.IMemberMap?`.

In `SqlMapper.GenerateDeserializerFromMap` it seems like the output of this method is checked for null in the same way that `GetMember` is, so I guess returning null is probably valid here?

mgravell:
this change is just the release notes for NRT tweak; it should also have included https://github.com/DapperLib/Dapper/commit/b8d119fa635b535574593daf07aea831259ab698, but I was a moron and forgot to branch - my bad

fixes https://github.com/DapperLib/Dapper/issues/1969

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -27,7 +27,7 @@ Note: to get the latest pre-release build, add ` -Pre` to the end of the command
 - infer command text without any whitespace as stored-procedure (#1975 via @mgravell)
 - add global `SupportLegacyParameterTokens` setting to enable or disable single-character parameter tokens (#1974 via @Giorgi)
 - revert `$` addition for legacy parameter tokens (#1979 via @mgravell)
-- change NRT annotation on `GetConstructorParameter` (fixes #1969)
+- change NRT annotation on `GetConstructorParameter` (#1980 via @mgravell, fixes #1969)
 
 ### 2.1.4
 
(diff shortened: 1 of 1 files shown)


mgravell:
[2.1.11](https://github.com/DapperLib/Dapper/releases/tag/2.1.11)

Issue #1976, opened 2023-10-12
Concrete initial v3 API change proposal


mgravell:
Dapper has grown organically over time, and not all decisions were good ones. We have worked hard to try to avoid runtime or compile-time breaks, but this means we have been constrained by those poor decisions.

I would like to get to work on shipping a V2->V3 transition that fixes this, and get some breaks over and done with. As [discussed here](https://github.com/DapperLib/Dapper/issues/1909), some *feature* work may be shunted to the AOT project, but we need the right API for things to be effective.

## What are we trying to solve?

The key problems in the API in V2 are:

- the signed/unsigned split is a nuisance - there should be one and only one Dapper
- starting from the interfaces (`IDbConnection` etc) rather than the concrete base types (`DbConnection`) was a mistake
- `CommandDefinition` was well intentioned, but again: a mistake - we should expose cancellation via a standard `CancellationToken` parameter 
- the shared API between buffered and unbuffered data is suboptimal (at best)
- we should add support for `DbDataSource` as a root primitive

Contraints on changes:

- where possible, we should *avoid* runtime breaks, and *minimize* compile-time breaks

## Planned implementation

My proposal, effectively, is to retire the existing API but leave it in place (so: no missing methods) by removing the `this` modifier (so they no longer resolve as extension methods), and create a *new* set of extension methods - probably on a new `static class` that expose the *preferred* API, meaning: the shape we want, and the most generalized version only (no hive of forwarded methods with increasing numbers of parameters reflecting our incremental understanding of the required API)

For example:

``` diff
  public static class SqlMapper {
-     public static IEnumerable<T> Query<T>(this IDbConnection conection, string sql, object? parameters = null, bool buffered = false, ...);
+     public static IEnumerable<T> Query<T>(IDbConnection conection, string sql, object? parameters = null, bool buffered = false, ...);
  }
+  public static class DbConnectionExtensions {
+    public static List<T> Query<T>(DbConnection conection, string sql, object? parameters = null, ...);
+    public static IEnumerable<T> QueryUnbuffered<T>(DbConnection conection, string sql, object? parameters = null, ...);
+ }
```

Key changes:

- root type changes from `IDbConnection` to `DbConnection`
- `buffered` is now explicit on the method signature rather than a parameter (see the existing `QueryUnbufferedAsync`)
- `async` methods will gain `CancellationToken` parameters

However:

- all code will continue to run without recompile (ignoring the signed/unsigned fun)
[...]

PR #1983, opened 2023-10-15
first stab at v3

mgravell:
Initial changes:

- tfm net461 -> net462, so allow:
- use BCL async interfaces on netfx
- remove `this` on SqlMapper connection methods
- new DbConnectionExtensions
- new DbDataSourceExtensions
- add CommandFlags to all APIs
- add CancellationToken to all async APIs
- subst CommandType? = null to CommandType = default
- subst int? = null (timeout) to int? = {some default TBD}


mgravell:
Xref #1976

iamcarbon:
.NET 7.0 will be EOL on May 14, 2024. It may be worth skipping the .NET 7.0 target and requiring .NET8.0 for DbDataSource.

mgravell:
@iamcarbon for now I'm happy to keep net7 - that also gives us the ability to use the regex codegen tool (adding that to "main" right now, no API change except a net7 tfm); if net7 adds cost, and/or we want to add net8 APIs: that becomes more pressing, but right now: meh

PR #1998, opened 2023-11-13
AppVeyor: Restore MySQL

NickCraver:
Looks like the image was borked with MySQL 8 went in (because it's still trying to start 5.7), see https://github.com/appveyor/ci/issues/3894.


mgravell (approved):
looks good, but I note `MYSQL_PATH: C:\Program Files\MySql\MySQL Server 5.7` - does that matter?

NickCraver:
Oh woops, I guess it doesn't but let's fix to match

Issue #2011, opened 2023-12-08
CustomPropertyTypeMap propertySelector does now allow null


uitmaurik:
The constructor of `CustomPropertyTypeMap` accepts two parameters. The propertySelector parameter is of type` Func<Type, string, PropertyInfo>`. If I understand correctly this means that the function must always return a PropertyInfo. Returning `null` is not allowed due to the project setting <Nullable>enable</Nullable>.
As the method `public SqlMapper.IMemberMap? GetMember(string columnName)` is the only method that uses the propertySelector and is already capable of handling a `null` return value, I would like to request to modify the constructor to (so changing the return type of the function from `PropertyInfo` to `PropertyInfo?`):

`        /// <summary>
        /// Creates custom property mapping
        /// </summary>
        /// <param name="type">Target entity type</param>
        /// <param name="propertySelector">Property selector based on target type and DataReader column name</param>
        public CustomPropertyTypeMap(Type type, Func<Type, string, PropertyInfo?> propertySelector)
        {
            _type = type ?? throw new ArgumentNullException(nameof(type));
            _propertySelector = propertySelector ?? throw new ArgumentNullException(nameof(propertySelector));
        }
`

PR #2012, opened 2023-12-08
Made result of CustomPropertyTypeMap parameter propertySelector nullable

uitmaurik:
To support sql statements for which the target type does not have a property for every column returns by the sql statement.
#2011


mgravell (approved):

rabiaabrar:
Any chance this is getting merged?

PR #2029, opened 2024-01-19
Demonstrate Issue with Tests

Crisfole:
There's some level of setup I'm missing to be able to run the tests (futzing around with that right now). But I think this documents my theory correctly so I can find out which of these options this is:

 1. a misunderstanding to be corrected with documentation
 2. a feature request
 3. a bug

I'm not clear on what the expected behavior is for "simple" records. I'd expect the parameter names to get matched up the same way "normal" records do...

### DIFF
--- a/tests/Dapper.Tests/MiscTests.cs
+++ b/tests/Dapper.Tests/MiscTests.cs
@@ -74,6 +74,10 @@ namespace Dapper.Tests
             public PositionalCarRecord() : this(default, default, default) { }
         }
 
+        private record SimpleRecord(int InternalId, int ExternalId, string? Name);
+
+        private record SimpleRecordReordered(int InternalId, string? Name, Guid ExternalId);
+
         private record NominalCarRecord
         {
             public int Age { get; init; }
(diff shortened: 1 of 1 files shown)


Crisfole:
I get 675 failures when trying to run this because Kerberos is not setup. I'm not sure how to run _just_ my new specs as `dotnet test --filter Name~SimpleRecord` says it can't find any specs like it...I'm happy to try something else.

mgravell:
Right; so I think the key topic here is:

- simple records don't have default constructors
- if there's no default constructor, there must currently need to be a constructor that matches the column order

Is this a correct restatement of what you're seeing? If so: the good news here is that I believe this is already addressed, but: over in Dapper.AOT; I'm trying hard to resist major functionality changes in vanilla Dapper, because touching the IL emit is a major pain point. I'm busy for a few days, but: I can check how these tests behave in AOT after that - or you could give it a whirl and let me know :)

Crisfole:
@mgravell I think this is the crux of it. Again, I can't run the tests locally for reasons I haven't figured out yet. Sounds like I could add AOT to my project and see if that resolves it. I can give it a whirl on Monday.

PR #2032, opened 2024-01-22
Update SqlMapperExtensions.Async.cs

AlexTian2012:
Extract type judgment to private method

### DIFF
--- a/Dapper.Contrib/SqlMapperExtensions.Async.cs
+++ b/Dapper.Contrib/SqlMapperExtensions.Async.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


Dapper.Contrib/SqlMapperExtensions.Async.cs
```diff
@@ -143,24 +143,7 @@ public static partial class SqlMapperExtensions
...
-                    typeInfo.ImplementedInterfaces.Any(ti => ti.IsGenericType && ti.GetGenericTypeDefinition() == typeof(IEnumerable<>)) ||
-                    typeInfo.GetGenericTypeDefinition() == typeof(IEnumerable<>);
-
-                if (implementsGenericIEnumerableOrIsGenericIEnumerable)
-                {
-                    isList = true;
-                    type = type.GetGenericArguments()[0];
-                }
-            }
+            isList = IsArrayAndGetElementType(ref type);
```
mgravell, line 146: initialize inline with decl

mgravell (changes requested):
looks good - nice catch of a dup; one minor nit, but: +1

PR #2038, opened 2024-02-05
Exposing the transaction used in class Database.

### LINKED ISSUE #1944: dapper.rainbow external transaction support
How to best use Dapper.rainbow code with transactions? For example, I may do operations that are not supported by dapper.rainbow in the same transaction as dapper.rainbow operations.

Dapper.rainbow  Database class has a field `private DbTransaction _transaction;` that cannot be accessed. This means there is no way to "enlist" the rainbow database in an existing transaction. 

There is a method `void BeginTransaction()` which allows me to create a new transaction. But since it is not returned, I cannot piggyback on top of it. 

I was thinking that maybe we need to extend the code base.
*  Either by making an overload  of `Init()` taking a transaction parameter from which the transaction and the connection can be accessed. 
* or by creating a method  `void EnlistTransaction(DbTransaction tx) { _dbTransaction = tx; } `
* Or maybe just expose the transaction variable. The very least make it `protected` such that subclasses can access it.
* 
Will this be acceptable? I'm happy to create a pull request if we can agree on a design.

kbilsted:
This enables to do queries not supported by rainbow within the same transaction

Closes #1944

### DIFF
--- a/Dapper.Rainbow/Database.cs
+++ b/Dapper.Rainbow/Database.cs
@@ -175,7 +175,11 @@ namespace Dapper
 
         private DbConnection _connection;
         private int _commandTimeout;
-        private DbTransaction _transaction;
+
+        /// <summary>
+        /// Get access to the underlying transaction
+        /// </summary>
+        public DbTransaction Transaction { get; }
 
         /// <summary>
         /// Get underlying database connection.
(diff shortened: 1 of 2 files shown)


kbilsted:
The build pipeline is failing on installing dependencies so it cannot run the tests

kbilsted:
still errors in the pipeline. please help 

```
Packages requiring reboot:
 - vcredist140 (exit code 3010)
The recent package changes indicate a reboot is necessary.
 Please reboot at your earliest convenience.

Command exited with code 3010```

File: Dapper.Rainbow/Database.cs [hunk 1]
```diff
@@ -175,7 +175,11 @@ public Table(Database<TDatabase> database, string likelyTableName)
 
         private DbConnection _connection;
         private int _commandTimeout;
-        private DbTransaction _transaction;
+
+        /// <summary>
+        /// Get access to the underlying transaction
+        /// </summary>
+        public DbTransaction Transaction;
```
mgravell, line 182: property please - i.e. add `{get;}`

mgravell:
gotta love CI servers :/ working on it: https://github.com/DapperLib/Dapper/pull/2042

mgravell:
CI issue fixed

Dapper.Rainbow/Database.cs
kbilsted: done

mgravell (approved):

PR #2045, opened 2024-02-17
Fix build error caused by read only property #2044

### LINKED ISSUE #2044: Build error due to read only Transaction in Dapper.Rainbow
Hi,
I'm getting an error when attempting to build the Dapper.Rainbow project 

`Property or indexer 'Database<TDatabase>.Transaction' cannot be assigned to -- it is read only` 
[Error in github Actions](https://github.com/DapperLib/Dapper/actions/runs/7937770474/job/21675575092)

[The Transaction property was recently changed to Read-only causing the problem ](
https://github.com/DapperLib/Dapper/blob/4f8e92fba53e789bf55d3054b30b3a5555a40363/Dapper.Rainbow/Database.cs#L182).

adding `private set; `will allow the Transaction to be updated within the class while keeping it protected from being updated outside of the class
 
```c#
/// <summary>
/// Get access to the underlying transaction
/// </summary>
public DbTransaction Transaction { get; private set; }
```

DarylJtn:
Resolves #2044

### DIFF
--- a/Dapper.Rainbow/Database.cs
+++ b/Dapper.Rainbow/Database.cs
@@ -179,7 +179,7 @@ namespace Dapper
         /// <summary>
         /// Get access to the underlying transaction
         /// </summary>
-        public DbTransaction Transaction { get; }
+        public DbTransaction Transaction { get; private set; }
 
         /// <summary>
         /// Get underlying database connection.


mgravell (approved):

mgravell:
Appreciated, thanks

PR #2068, opened 2024-04-03
fix(TypeHandler): use the custom type handler when packing a list parameter

### LINKED ISSUE #2067: "No mapping exists" error when using a TypeHandler and a collection parameter
Dapper is throwing an error when running a query in the following scenario:
- the parameter is a collection of objects of a given type T
- and the type T is not natively supported by the SQL client
- and a TypeHandler is defined for this type T

Example:

```csharp
// a type that is not natively supported by SqlClient
public struct LocalDate
{
    public int Year { get; set; }
    public int Month { get; set; }
    public int Day { get; set; }
}

// a type handler
public class LocalDateHandler : SqlMapper.TypeHandler<LocalDate>
{
    private LocalDateHandler() { /* private constructor */ }

    // Make the field type ITypeHandler to ensure it cannot be used with SqlMapper.AddTypeHandler<T>(TypeHandler<T>)
    // by mistake.
    public static readonly SqlMapper.ITypeHandler Default = new LocalDateHandler();

    public override LocalDate Parse(object? value)
    {
        var date = (DateTime)value!;
        return new LocalDate { Year = date.Year, Month = date.Month, Day = date.Day };
    }

    public override void SetValue(IDbDataParameter parameter, LocalDate value)
    {
        parameter.DbType = DbType.DateTime;
        parameter.Value = new DateTime(value.Year, value.Month, value.Day);
    }
}

// the failing method
public void RunQueryWithParamCollectionAndCustomTypeHandler()
{
    SqlMapper.AddTypeHandler(typeof(LocalDate), LocalDateHandler.Default);

    var singleParameter = new { SingleDate = new LocalDate { Year = 2014, Month = 7, Day = 25 } };

    // this succeeds
    var resultSingle = connection.Query<int>("SELECT 1 WHERE '2014-07-25' = @SingleDate", singleParameter).Single();

    var parameters = new
    {
        ListOfDates = new List<LocalDate>
        {
            new() { Year = 2014, Month = 7, Day = 25 },
            new() { Year = 2014, Month = 7, Day = 26 },
        }
    };

    // this fails with 'No mapping exists from object type LocalDate to a known managed provider native type.'
    var result = connection.Query<int>("SELECT 1 WHERE '2014-07-25' IN @ListOfDates", parameters).Single();
}
```

tlecomte:
Fixes #2067

The code for `SqlMapper.PackListParameters` was not using the custom type handler. This caused an error like `No mapping exists from object type xxxxx to a known managed provider native type.` when passing a collection parameter to a query, when the type of the collection items is not natively supported by the SQL client, and when a custom `TypeHandler` is registered for it.

Here this is fixed by using the same logic that is already in `DynamicParameters.AddParameter`.

A unit test is added to cover the scenario.

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -2168,6 +2168,7 @@ namespace Dapper
                 if (list is not null && !viaSplit)
                 {
                     object? lastValue = null;
+                    SqlMapper.ITypeHandler? handler = null;
                     foreach (var item in list)
                     {
                         if (++count == 1) // first item: fetch some type info
@@ -2178,7 +2179,7 @@ namespace Dapper
                             }
                             if (!isDbString)
                             {
-                                dbType = LookupDbType(item.GetType(), "", true, out var handler);
+                                dbType = LookupDbType(item.GetType(), "", true, out handler);
                             }
                         }
                         var nextName = namePrefix + count.ToString();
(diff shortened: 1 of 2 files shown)


tlecomte:
Dear @mgravell, please see here a PR to fix an issue with list parameters and custom type handlers. Thank you very much for considering this!

SimonAfek:
I was just about to post an issue regarding this very thing when I saw this pull request. Hope this gets merged soon.

I have a custom type handler for `Guid`s that get stored as `binary(16)` in my database. Works fine for single parameters in queries like `SELECT * FROM [TableName] WHERE [Id] = @Id`, but when I supply a list parameter for a query like `SELECT * FROM [TableName] WHERE [Id] IN @Ids;`, Dapper reverts back to treating them with the default handler instead.

aquamoth:
A year later, and we have the exact same issue @SimonAfek is describing.
What prevents this from being merged? If it is a quality issue, I am sure we can pitch in to resolve outstanding issues.

PauloHMattos:
@mgravell Just a gentle bump on this one when you have the bandwidth!
I'm really hoping to see this land so I can avoid spinning up a custom fork to work around this issue.

If the author is unavailable to make any requested changes, please let me know. I'm happy to take over to get this merged.

Issue #2072, opened 2024-04-15
DateOnly "off by one" column error


edwardmjackson:
Hi,

  Using Microsoft.Data.SqlClient 5.2.0 and Dapper 2.1.44, we've noticed an issue with the new DateOnly type support from [#2051](https://github.com/DapperLib/Dapper/pull/2051). 
   Using a simple class such as 
```
  public class PersonTest
  {
    public string FirstName { get; set; }
    public DateOnly FromDate { get; set; }
  }
```
and a QueryAsync with 
```
    public async Task<PersonTest> GetPerson(string Username)
    => (await QueryAsync<PersonTest>(AdminProcs.GetPerson, new { Username })).FirstOrDefault();
```

we get an error response of 
```
{
  "type": "System.Data.DataException",
  "title": "Dapper",
  "status": 500,
  "detail": "Error parsing column 1 (FromDate=Ed - String)",
  "traceId": "00-ae3a4c30e843c63d927302fde275a754-ca340c148ee9f15f-00"
}
```

The data, returned by MS SQL Server, as a varchar and DateTime type, are as follows:
```
FirstName                                          FromDate
-------------------------------------------------- -----------------------
Ed                                                 2019-10-01 00:00:00.000
```

The issue seems to be that it is trying to parse column 1 as the FromDate with the value of Ed, which is in column 0. Hence the "off by one" error.

Running this with Dapper 2.1.35, using the the previous custom solution from [#1715](https://github.com/DapperLib/Dapper/issues/1715#issuecomment-1149665776), does not have this error. 

Staying on 2.1.44, if we remove `FromDate` from `PersonTest`, it works fine with just the FirstName. 
Changing FromDate in PersonText to a more open type, such as string, also works fine.
Changing the SQL type, as e.g., as MS SQL Date, or an int or varchar in ISO8601 format, doesn't have any effect.

quiian:
Also ran into this, dapper attempts to parse DateOnly from column #11 instead of column #12.

![Screenshot 2024-04-15 141405](https://github.com/DapperLib/Dapper/assets/103261442/2376aa6d-a279-419c-9a5a-b8125109b992)
![image](https://github.com/DapperLib/Dapper/assets/103261442/fd4b3c91-78dc-41e0-aee3-935b7b9b9a02)

mgravell:
Well that sounds very bad; I'm hiding 2.1.44 while I investigate - hopefully should have a fix out today or tomorrow

[...]

Issue #2071, opened 2024-04-15
Problem with DateOnly mapping when updating form v2.1.35 to 2.1.44


glebron-woffu:
I used to have this custom type handler configured:
`SqlMapper.AddTypeHandler(new DapperSqlDateOnlyTypeHandler());`

Mapping class is:

```
public class DapperSqlDateOnlyTypeHandler : SqlMapper.TypeHandler<DateOnly>
{
    public override void SetValue(IDbDataParameter parameter, DateOnly date)
        => parameter.Value = date.ToDateTime(new TimeOnly(0, 0));

    public override DateOnly Parse(object value)
        => DateOnly.FromDateTime((DateTime)value);
}

```
This was working fine, but looks like the DateOnly support was added in [2.1.37](https://github.com/DapperLib/Dapper/releases/tag/2.1.37) but
Now I am getting this error:
```
System.Data.DataException
Error parsing column 1 (Date=669745 - Int32)
   at Dapper.SqlMapper.ThrowDataException(Exception ex, Int32 index, IDataReader reader, Object value) in /_/Dapper/SqlMapper.cs:line 3948
Inner exception:
InvalidCastException: Unable to cast object of type 'System.DateTime' to type 'System.DateOnly'.
```
   Thing is, the Date column is column 1, but somehow is parsing column 0 value
 Query result: 
`   669745	2023-11-01	NULL	NULL	...`

So I am a bit confused, in v2.1.35 I didn't have this problem. I removed custom mapper SqlMapper.AddTypeHandler(new DapperSqlDateOnlyTypeHandler()); and still same problem.
Edit:
In another query, I set date only field as first one:
`2023-11-01	0 	0`
The error is:
```
System.Data.DataException
Error parsing column 0 (Date=n/a - Unable to cast object of type 'System.DateTime' to type 'System.DateOnly'.)
   at Dapper.SqlMapper.ThrowDataException(Exception ex, Int32 index, IDataReader reader, Object value) in /_/Dapper/SqlMapper.cs:line 3948
   at Deserialize68929759-6271-4665-a68e-ff25a40601eb(DbDataReader)
   at Dapper.SqlMapper.QueryAsync[T](IDbConnection cnn, Type effectiveType, CommandDefinition command)
```
The class is:
```

public record TestClass
{
    public DateOnly Date { get; set; }
    public int First { get; set; }
    public int Second { get; set; }
}

```

mkorsukov:
[...]

PR #2080, opened 2024-04-26
disable DateOnly / TimeOnly support

mgravell:
see #2071 / #2072


RobK410:
what is the recommended solution for DateOnly support for .44 users?

mgravell:
I guess: stay on .44 (or try AOT, which now has Date only/TimeOnly). I will try to find the time to figure out exactly how the cogs exploded

PR #2102, opened 2024-07-16
Changed the 'Add' methods return type to "DynamicParameters".

CodeHawkz:
This would allow these methods to be used for method chaining, which is an simple and yet effective way. An example would be,

```csharp
    var parameters1 = new DynamicParameters()
                        .Add("Parameter 1", 123)
                        .Add("Parameter 2", "Hello")
                        .Add("Parameter 3", "c8e68c5d-0d20-4aad-841b-f9ce0436acc2");
    var parameters2 = new DynamicParameters()
                        .AddRange(
                              ("Parameter 1", 123),
                              ("Parameter 2", "Hello"),
                              ("Parameter 3", "c8e68c5d-0d20-4aad-841b-f9ce0436acc2")
                        );
```

Changes:
	- Modified:   Dapper/DynamicParameters.cs

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -44,7 +44,7 @@ namespace Dapper
         /// EG: AddDynamicParams(new {A = 1, B = 2}) // will add property A and B to the dynamic
         /// </summary>
         /// <param name="param"></param>
-        public void AddDynamicParams(object? param)
+        public DynamicParameters AddDynamicParams(object? param)
         {
             var obj = param;
             if (obj is not null)
@@ -84,6 +84,8 @@ namespace Dapper
                     }
                 }
             }
+
+            return this;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


CodeHawkz:
@mgravell may I know why this is a hard breaking change please?

mgravell:
Because the return type is part of the signature required for binary compatibility. Think in stack semantics: existing code is going to expect Add to leave zero items on the stack. With this change, 1 item will be left on the stack, which would now require a "pop" instruction for code that wanted to actively ignore it. But code that compiled against the old version would now have an imbalanced stack, which the JIT will detect and reject in either a bind exception or an invalid program exception.

CodeHawkz:
Thank you for the reply. Your reply gave me something to think and learn about. Appreciate it. 

Last question: does the "AddRange" method fall victim to the same issue? It feels like it is not.

mgravell:
The problem is *changing* the return type. AddRange already has this return behaviour, so that's fine - we're not changing anything.

CodeHawkz:
This PR is abandoned, due to design architecture conflict.

Issue #1815, opened 2022-08-09
SqlMapper.AddTypeHandler not thread safe


StephanMoeller:
SqlMapper.SetTypeMap is thread safe. This is achievede by:

lock (_typeMaps)
{
    ...
}

But SqlMapper.AddTypeHandler is not thread safe.

I think it would be valuable to make AddTypehandler thread safe as it enables different data access to be self contained instead of having a central place for registering all custom types.

Scenario:

Using 3rd party package A and another third party package B.
These both use Dapper. No problem, it is the same version.

These two packages each register some custom type handlers lazily, the first time they need them.

In AddTypeHandler there is no locking, the flow is as follows:

...
var snapshot = typeHandlers;
... // <= In this time span, multiple simultanious calls will trigger lost updates.
typeHandlers = newCopy;

I felt this the hard way in production yesterday.

PR #2107, opened 2024-08-03
Make AddTypeHandlerImpl in SqlMapper thread safe

mobilebilly:
As discussed in issue #1815, the current implementation of SqlMapper.AddTypeHandlerImpl is not thread-safe. When typeHandlers is cloned during concurrent access to this method, a race condition occurs because each thread maintains its own copy of typeHandlers. This can result in handlers added by some threads being lost. This fix addresses the issue to ensure thread safety.

Additionally, we propose to obsolete below overload version of AddTypeHandlerImpl:

```
void AddTypeHandlerImpl(Type type, ITypeHandler? handler, bool clone)
```

Also suggest making all overloads of AddTypeHandlerImpl non-public, as their method names indicate that they are intended for internal implementation, but that could be handled separately in the future.

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -255,22 +255,19 @@ namespace Dapper
                 [typeof(TimeOnly?)] = TypeMapEntry.DoNotSetFieldValue,
 #endif
             };
-            ResetTypeHandlers(false);
+            ResetTypeHandlers();
         }
 
         /// <summary>
         /// Clear the registered type handlers.
         /// </summary>
-        public static void ResetTypeHandlers() => ResetTypeHandlers(true);
-
         [MemberNotNull(nameof(typeHandlers))]
-        private static void ResetTypeHandlers(bool clone)
-        {
+        public static void ResetTypeHandlers() {
             typeHandlers = [];
-            AddTypeHandlerImpl(typeof(DataTable), new DataTableHandler(), clone);
-            AddTypeHandlerImpl(typeof(XmlDocument), new XmlDocumentHandler(), clone);
-            AddTypeHandlerImpl(typeof(XDocument), new XDocumentHandler(), clone);
-            AddTypeHandlerImpl(typeof(XElement), new XElementHandler(), clone);
+            AddTypeHandlerImpl(typeof(DataTable), new DataTableHandler());
+            AddTypeHandlerImpl(typeof(XmlDocument), new XmlDocumentHandler());
+            AddTypeHandlerImpl(typeof(XDocument), new XDocumentHandler());
+            AddTypeHandlerImpl(typeof(XElement), new XElementHandler());
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


billrob:
Hey @mgravell I don't think a clone is necessary.  I can fix all the threading issues if cloning is confirmed to not be necessary.  I see this is old code from 2015.  There is comment `// use clone, mutate, replace to avoid threading issues` that doesn't work the author thought it would.

Dapper/SqlMapper.cs
```diff
@@ -373,10 +378,21 @@ public static void AddTypeHandlerImpl(Type type, ITypeHandler? handler, bool clo
                 }
             }
 
-            var snapshot = typeHandlers;
-            if (snapshot.TryGetValue(type, out var oldValue) && handler == oldValue) return; // nothing to do
-
-            var newCopy = clone ? new Dictionary<Type, ITypeHandler>(snapshot) : snapshot;
+            lock (typeHandlers)
+            {
+                if (typeHandlers.TryGetValue(type, out var oldValue) && handler == oldValue) return; // nothing to do
```
billrob, line 383: Can you do this as a ConcurrentDictionary instead of locking?

billrob (changes requested):

mgravell:
The existing code perhaps has risk of lost updates, but this API is not expected to be used aggressively on-the-fly, so in reality: this shouldn't be a problem. If we're concerned, then some simple synchronization in those few methods should be fine.

The proposed changes, however, look to be *more* unsafe - in that it allows competing threads doing routing query work (not adding handlers) to attempt to read the dictionary *while it is being mutated* (only the write is guarded by the lock). This can break in much worse ways.

The old code deliberately avoided this by never allowing the "live" dictionary to be mutated.

Potentially this can be mitigated using things like concurrentdictionary, but: I wonder if this would thing can be much simpler by just locking on a separate sync token for the duration of any admin operations, but maintaining the snapshot/mutate/swap approach.

mgravell:
Regarding the "impl" method; damn, that method was never intended to be public. Separately, I'll deal with that (in a different PR).  Just pretend that public API doesn't exist.

billrob:
@mgravell Right, this PR code currently isn't ideal.  I can run this PR to get it into a good shape.  Or if @mobilebilly is okay with it, I can just do my own PR tidying this section up.  Even the existing "clone operation" can run afoul of issues during the iteration for the clone.

I was thinking about setting it as a `ConcurrentDictionary<T,K>` and removing all the clone parameters and updating to TryAdd / TryGet.  As far as two handlers for the same type coming in a threaded situation, do you want the first or last to win?

mgravell:
I've proposed https://github.com/DapperLib/Dapper/pull/2129, which I think does everything needed in a safer way, with fewer moving parts. Thoughts?

mgravell:
dammit appveyor, stop messing with pgsql!

mgravell:
Closing out with preference to #2129

PR #2112, opened 2024-09-17
🐛 [QueryAsync] `NoCache` flag ignored when doing MultiMap

Dot-H:
As the `AssertNoCacheWorksForMultiMap` tests shows, when the `QueryAsync` method is passed a `splitOn` and `map` arguments, the `NoCache` flag is actually ignored.

The issue seems to come from [here](https://github.com/DapperLib/Dapper/blob/main/Dapper/SqlMapper.Async.cs#L942)

### DIFF
--- a/tests/Dapper.Tests/AsyncTests.cs
+++ b/tests/Dapper.Tests/AsyncTests.cs
@@ -1013,5 +1013,24 @@ SET @AddressPersonId = @PersonId", p).ConfigureAwait(false))
             await MarsConnection.QueryAsync<int, int, (int, int)>(cmdDef, splitOn: "b", map: (a, b) => (a, b));
             Assert.Equal(0, SqlMapper.GetCachedSQLCount());
         }
+
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


tests/Dapper.Tests/AsyncTests.cs
```diff
@@ -995,5 +995,42 @@ public void AssertNoCacheWorksForQueryMultiple()
...
+                a,
+                b
+            }, commandType: CommandType.Text, flags: CommandFlags.NoCache | CommandFlags.Buffered);
+
+            SqlMapper.PurgeQueryCache();
+            var before = SqlMapper.GetCachedSQLCount();
+            Assert.Equal(0, before);
+
+            await MarsConnection.QueryAsync<int, int, (int, int)>(cmdDef, splitOn: "b", map: (a, b) => (a, b));
+            Assert.Equal(0, SqlMapper.GetCachedSQLCount());
```
Dot-H, line 1014: This fails. `SqlMapper.GetCachedSQLCount()` actually returns 1

Issue #2117, opened 2024-10-02
Possible concurrency issues with caches?


JamalQasem:
We are using SqlKata/Dapper(2.1.35) to implement a Web-API for a Postgres database with
around 20 tables. The API contains methods to query and manipulate data. It does not use 
stored procedures. We test the repository methods of the API with around
200 unit tests, which run transactions against a test database rolling them back instead of committing. 

Recently one of our tests started failing with the following error message:
```Npgsql.PostgresException : 23503: insert or update on table "___" violates foreign key constraint "___"```

We then annotated the offending test case with

```[TestFixture, RequiresThread(ApartmentState.STA)]```

and the problem vanished. All our repository methods rely on functions like

```
    protected async Task<IEnumerable<TEntity>> QueryAsync(Query query, IDbTransaction transaction, CancellationToken token)
    {
        var command = CompileQuery(query, transaction, token);
        return await transaction.Connection!.QueryAsync<TEntity>(command).ConfigureAwait(false);
    }

    private async Task<int> ExecuteAsync(Query query, IDbTransaction transaction, CancellationToken token)
    {
        var command = CompileQuery(query, transaction, token);
        return await transaction.Connection!.ExecuteAsync(command).ConfigureAwait(false);        
    }
```

We first checked that the queries use different connections of the connection pool and experimented 
with the command flag `NoCache`to no avail. 
We finally decided  to serialize the access using a semaphore.
This together with a call of `SqlMapper.PurgeQueryCache();` made the failing test working again in MT.

The current implementation looks as follows 
```
    protected async Task<IEnumerable<TEntity>> QueryAsync(Query query, IDbTransaction transaction, CancellationToken token)
    {
        await _semaphore.WaitAsync().ConfigureAwait(false);
        try
        {
            var command = CompileQuery(query, transaction, token);
            return await transaction.Connection!.QueryAsync<TEntity>(command).ConfigureAwait(false);
        }
        finally
        {
            SqlMapper.PurgeQueryCache();
            _semaphore.Release();
        }
    }
```
We suspect that we somehow manage to use different statements which map
to the same cache `Identity` so that concurrently running queries may use 
a wrong cache entry. Leaving us with 2 unanswered questions:
- Did anyone run into the same issue?
- Is there a way to completely avoid caching?

mgravell:
[...]

PR #2118, opened 2024-10-03
Fix #2113

### LINKED ISSUE #2113: 🐛 [QueryAsync] NoCache flag ignored when doing MultiMap
It seems like `QueryAsync` when being passed the `map` and `splitOn` argument ends up ignoring the `NoCache` command flag. 

You can see https://github.com/DapperLib/Dapper/pull/2112 for a repro

goerch:
We also can't use `NoCache` to work around the problems in #2117. This PR seems to make `NoCache`work for our use case. Thanks to @Dot-H for providing test cases.

### DIFF
--- a/Dapper/CommandDefinition.cs
+++ b/Dapper/CommandDefinition.cs
@@ -11,16 +11,9 @@ namespace Dapper
     /// </summary>
     public readonly struct CommandDefinition
     {
-        internal static CommandDefinition ForCallback(object? parameters)
+        internal static CommandDefinition ForCallback(object? parameters, CommandFlags flags)
         {
-            if (parameters is DynamicParameters)
-            {
-                return new CommandDefinition(parameters);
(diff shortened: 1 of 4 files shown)


Dapper/CommandDefinition.cs
```diff
@@ -11,16 +11,9 @@ namespace Dapper
...
+        internal static CommandDefinition ForCallback(object? parameters, CommandFlags flags)
         {
-            if (parameters is DynamicParameters)
-            {
-                return new CommandDefinition(parameters);
-            }
-            else
-            {
-                return default;
-            }
```
goerch, line 23: On a second thought: is it neccessary to mask out some of then command flags?

mgravell (approved):

mgravell:
CI looks to be a postgresql change on the default image. Checking.

PR #2122, opened 2024-10-12
Properly handle value types when setting properties on dynamic objects returned by Dapper queries

alatanza:
This PR resolves an issue where assigning unboxed value types to dynamic properties in `DapperRow` objects caused an exception. The problem occurred because value types are not automatically boxed when using expression trees, resulting in errors like:

`System.ArgumentException: Expression of type 'System.Int32' cannot be used for parameter of type 'System.Object' of method 'System.Object SetValue(System.String, System.Object)'`

The fix ensures that value types are properly boxed before being assigned.

Tests have been added to validate the behavior, covering scenarios with reference types, boxed value types, and unboxed value types.

Please review and provide any feedback or suggestions. Thanks for your time and consideration!


mgravell (approved):
LGTM, thanks

PR #2123, opened 2024-10-15
Add DynamicParametersExtensions

Behrad87:
in this extension we can pass DTO class as parameter and add the members which are not null.


mgravell:
The intended use-case for this is unclear. Can you perhaps give some examples of why/when this would be useful? Also note (although this is secondary to understanding the intent): there's no tests, and these are `internal`, so: not actually available for external callers (as such, they also don't need declaring in the ship files, although I imagine that they are meant to be `public`?)

PR #2129, opened 2024-10-31
Mark AddTypeHandlerImpl as obsolete and prevent lost updates via AddTypeHandler

### LINKED ISSUE #1815: SqlMapper.AddTypeHandler not thread safe

mgravell:
`AddTypeHandlerImpl` was never intended to be `public`; oops

also drive-by fix to prevent lost updates via AddTypeHandler

- mark AddTypeHandlerImpl as obsolete
- syncronize type-handler mutates, to prevent lost writes

fix #1815

alternate to #2107

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -266,11 +266,14 @@ namespace Dapper
         [MemberNotNull(nameof(typeHandlers))]
         private static void ResetTypeHandlers(bool clone)
         {
-            typeHandlers = [];
-            AddTypeHandlerImpl(typeof(DataTable), new DataTableHandler(), clone);
-            AddTypeHandlerImpl(typeof(XmlDocument), new XmlDocumentHandler(), clone);
-            AddTypeHandlerImpl(typeof(XDocument), new XDocumentHandler(), clone);
-            AddTypeHandlerImpl(typeof(XElement), new XElementHandler(), clone);
+            lock (typeHandlersSyncLock)
+            {
+                typeHandlers = [];
+                AddTypeHandlerCore(typeof(DataTable), new DataTableHandler(), clone);
+                AddTypeHandlerCore(typeof(XmlDocument), new XmlDocumentHandler(), clone);
+                AddTypeHandlerCore(typeof(XDocument), new XDocumentHandler(), clone);
+                AddTypeHandlerCore(typeof(XElement), new XElementHandler(), clone);
+            }
         }
 
         /// <summary>
@@ -339,7 +342,7 @@ namespace Dapper
         /// </summary>
         /// <param name="type">The type to handle.</param>
         /// <param name="handler">The handler to process the <paramref name="type"/>.</param>
-        public static void AddTypeHandler(Type type, ITypeHandler handler) => AddTypeHandlerImpl(type, handler, true);
+        public static void AddTypeHandler(Type type, ITypeHandler handler) => AddTypeHandlerCore(type, handler, true);
         /// <summary>
         /// Determine if the specified type will be processed by a custom handler.
         /// </summary>
@@ -353,7 +356,16 @@ namespace Dapper
         /// <param name="type">The type to handle.</param>
         /// <param name="handler">The handler to process the <paramref name="type"/>.</param>
         /// <param name="clone">Whether to clone the current type handler map.</param>
+        [Obsolete("Please use " + nameof(AddTypeHandler), error: true)]
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public static void AddTypeHandlerImpl(Type type, ITypeHandler? handler, bool clone)
+        {
+            // this method was accidentally made public; we'll mark it as illegal, but
+            // preserve existing usage in compiled code; sorry about this!
+            AddTypeHandlerCore(type, handler, clone);
+        }
+
+        private static void AddTypeHandlerCore(Type type, ITypeHandler? handler, bool clone)
         {
             if (type is null) throw new ArgumentNullException(nameof(type));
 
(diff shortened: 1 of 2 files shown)


Dapper/SqlMapper.cs
```diff
@@ -373,39 +385,46 @@ public static void AddTypeHandlerImpl(Type type, ITypeHandler? handler, bool clo
...
             }
 
-            var snapshot = typeHandlers;
-            if (snapshot.TryGetValue(type, out var oldValue) && handler == oldValue) return; // nothing to do
+            // synchronize between callers mutating type-handlers; note that regular query
+            // code may still be accessing the field, so we still use snapshot/mutate/swap;
+            // the synchronize is just to prevent lost writes
+            lock (typeHandlersSyncLock)
+            {
+                var snapshot = typeHandlers;
```
billrob, line 393: This assignment is not necessary.

Dapper/SqlMapper.cs
```diff
@@ -373,39 +385,46 @@ public static void AddTypeHandlerImpl(Type type, ITypeHandler? handler, bool clo
...
+                if (snapshot.TryGetValue(type, out var oldValue) && handler == oldValue) return; // nothing to do
 
-            var newCopy = clone ? new Dictionary<Type, ITypeHandler>(snapshot) : snapshot;
+                var newCopy = clone ? new Dictionary<Type, ITypeHandler>(snapshot) : snapshot;
```
billrob, line 396: If not using Concurrent, always make a copy, but I think there is still a thread issue during the iteration of the ctor for the `new Dictionary<>`.  If some other code touches the dictionary, which today it might not, but in the future...

Dapper/SqlMapper.cs
mgravell: because...?

Dapper/SqlMapper.cs
mgravell: In a separate push I've suppressed the bad `Impl`'s usage of `clone: false`, so: the only scenario where `clone` is `false` is in the .cctor, which is synchronized (over all field access etc) by the runtime. So the only scenario at real runtime is when we've duplicated *before* mutating. Concurrent reads with zero writes are fine.

"if we add broken code later" is a scenario that we can't plan for other than "try not to do that", but that's always true; we know the dictionary needs concurrency consideration

Dapper/SqlMapper.cs
mgravell: Ah yes, with the new `lock` we know no-one else is swapping this field behind our backs; fine

mgravell: [ea273e0](https://github.com/DapperLib/Dapper/pull/2129/commits/ea273e0297a2af78c3802ed5629651181db7d6d8)

mgravell:
(CI fails: appveyor keep changing the pgsql setup; trying to fix)

billrob (approved):
Still not 100% perfect, but 99% on an admin / housekeeping operation will be fine.

mgravell:
> Still not 100% perfect, but 99% on an admin / housekeeping operation will be fine.

what concrete scenario are you concerned about?

mgravell:
giving up on fighting appveyor - summoning the one, the only, @NickCraver - any clue? appveyor keep changing the base image; I fixed once [here](https://github.com/DapperLib/Dapper/commit/70e57f820dc5d845ecacb92b5628a4f5f8fc63ae), but can't get to work now

billrob:
Probably more a "not how I would do it" 1%.  The iteration when newing the dictionary is still potentially an exception.  At least it will throw an exception rather than build a non-desirable dictionary cache.  But verily, if someone is running multi-threading add/reset/remove type handlers we could just tell them to not do it. 

I don't see that currently possible with the current code base, but more future proofing, because in 2 years we will remember this discussion?  lol

mgravell:
> The iteration when newing the dictionary is still potentially an exception. 

I don't see how:

1. we never mutate the dictionary once assigned, *except for* in the `.cctor` (that's what `clone` controls)
2. it doesn't matter because of 1, but we're inside the `lock` for this iterator, so we know nobody is doing anything except reading

So : in what scenario can we compete? Happy to consider a concrete scenario, but : I don't see it currently.

As for the 2 years thing: we might not remember this specific discussion in the future, but yes: we will know that this is static state that needs concurrency consideration.

billrob:
Any potential issues does not exist currently in the code base and this PR is threaded sound.  I'm confident this currently fixes the known threading issues and you should push it.  (appveyor issues aside)

Larger story here, I want Dapper to survive EF consuming this project as they often consume popular projects.  Should we consider a `.UseDapper(with type handler registration)` that can internally register this static dictionary as a singleton with DI?  That would eliminate all these redo scenarios we potentially could face in the future and fit more inline with .net core direction.

mgravell:
I'm not sure that's the best option for configuration in this case. Over in the AOT side, we allow declarative and contextual configuration via attributes, which allows config to be as granular or global as desired without the need for additional registration or service tracking.

billrob:
I had a hard time doing static typemap configuration as Dapper and .net evolved regarding testing.  AOT might be a great solution, but it is so far from on-the-ground developers just wanting things to work.

The dotnet `.UseXXX` has taken off.  Dapper is infinitely more configurable than the current EF implementation mimicking Dapper's performance.   I might be myopic here, but why shouldn't we want a startup configuration that handles application specific type mappings because almost all scenarios don't involve changing type map configurations while your application is running.

mgravell:
Because you can already do that today without requiring a new API. Service registration kinda needs a service state to be "proper" (not just static state), and the existing Dapper bits lack that. As for AOT: you don't need to be deploying via AOT for Dapper's AOT bits to be useful - the aim there is to remove all the reflection code from the critical path.

NickCraver:
Postgres fixed up in #2130 - the docs are just totally missing past Postgresql 13, so you're not crazy. Even more reason we need to get off AppVeyor.

Issue #2125, opened 2024-10-22
QueryAsync with custom mapping AND CancellationToken


HrvojeJuric:
I want to use QueryAsync with multiple types mapped onto a single return type WITH a cancellation token but none of the QueryAsync functions allow for that.

My current solution is to use the private `MultiMapAsync` method but this is not ideal and I don't know what the right solution to this would be but I'd like to talk about it and if you're down I can make a pull request.

goerch:
Maybe [relevant](https://stackoverflow.com/questions/25540793/cancellationtoken-with-async-dapper-methods)?

haras-unicorn:
no because that one doesnt do multi mapping

goerch:
I think we use multi mapping in a similar way. Let me see

```
    protected CommandDefinition CompileQuery(Query query, IDbTransaction transaction, 
        CancellationToken token, CommandFlags flags = CommandFlags.Buffered)
    {
        var statement = _compiler.Compile(query).ToString();
        return new CommandDefinition(statement, transaction:transaction, 
            cancellationToken:token, flags:flags);
    } 

    protected async Task<IEnumerable<TEntity>> QueryAsync<TFirst, TSecond>(Query query, IDbTransaction transaction, CancellationToken token,
        Func<TFirst, TSecond, TEntity> map, string splitOn = "id")
        where TFirst: new() where TSecond: new()
    {
        var command = CompileQuery(query, transaction, token, CommandFlags.Buffered);
        return await transaction.Connection!.QueryAsync<TFirst, TSecond>(command, map, splitOn).ConfigureAwait(false);
    }
```

But I might be missing something.

HrvojeJuric:
this is for a static number of objects in the result
i want dynamic number of objects in the result and cancellation token

goerch:
> i want dynamic number of objects in the result and cancellation token

It could be helpful if you point us to the signature which is missing the `CommandDefinition` parameter?

HrvojeJuric:
public static Task<IEnumerable<TReturn>> QueryAsync<TReturn>(this IDbConnection cnn, string sql, Type[] types, Func<object[], TReturn> map, object? param = null, IDbTransaction? transaction = null, bool buffered = true, string splitOn = "Id", int? commandTimeout = null, CommandType? commandType = null)
    
    this one

Issue #1938, opened 2023-07-18
Add CancellationToken optional parameter to all Async methods


mcenzato:
I saw that a lot of Dapper async methods (eg. QueryAsync) miss the CancellationToken parameter.

Rendojack:
You can workaround this by passing CommandDefinition with assigned CancellationToken in most of the simple queries, except more complicates ones, like with multi-mapping.

digibum:
A friendly reminder for this. Especially for multi-mapping and a simpler implementation without the need for CommandDefinition. Just add an optional parameter to the basic Query methods. Thank you!

PR #2138, opened 2024-12-30
Add CancellationToken support to QueryAsync overload

itcotta:
This change introduces a CancellationToken parameter to the QueryAsync method, enabling support for cancellation of asynchronous database queries. The modification ensures better resilience and control over long-running or potentially cancelable operations.

This fixes issues #2125 and #1938.


PauloHMattos:
This would be a breaking change. I don't think this can me added until a new major release of Dapper

worming004:
This is in detail, and most C# developers accept this change as it is not breaking in a common way. But there is edge cases adding an optionnal parameter is a breaking change. For example:

Here is startup case
```csharp
public class Foo
{
    public static async Task Bar()
    {
        await Task.Delay(5);
        System.Console.WriteLine("Bar");
    }

}

public class MyProgram()
{
    public static async Task MyMain()
    {
        Func<Task> MyFunc = Foo.Bar;

        await MyFunc();
    }
}
```

When modifying to `public static async Task Bar(Cancellation Token = default)`, then MyProgram cannot compile anymore. `MyFunc` expect `Foo.Bar` to have a definition matching `Func<Task>`. But the only definition is now  `Func<CancellationToken, Task>` 

Here is the change without breaking change:

```csharp
public class Foo
{
    public static Task Bar()
    {
        return Bar(default);
    }

    public static async Task Bar(CancellationToken cts)
    {
        await Task.Delay(5, cts);
        System.Console.WriteLine("Bar");
    }
}

public class MyProgram()
{
    public static async Task MyMain()
    {
        Func<Task> MyFunc = Foo.Bar;

        await MyFunc();
    }
}
```

This is not breaking anymore.

mgravell:
To be clear: any change to the parameters is *always* a breaking change.

Fundamentally, there are two very different kinds of "break" we need to consider:

- source breaking changes - things that prevent new compile runs from completing
- runtime / "binary" breaking changes - things that prevent existing code compiled against an old version of the library from running against a newer version of the library

Changes can be either or both (or neither) kind of break.
We need to be mindful of both kinds of breaks, but since we're a commonly used library we need to be *especially* mindful of the second kind: the first kind of break causes a developer to have a bad 20 minutes while they change 3 lines of code with whatever change needs applying, but the second kind causes a *department or entire company* to have a bad *day or days* when a service randomly goes offline when they updated an innocent looking unrelated package that caused a transitive library update 3 steps removed, leading to MissingMethodException or similar.

So: we don't want that.

Adding a parameter - even an optional parameter - is a binary breaking: existing code compiled against the old method *will simply stop working* if the library is updated in this way, unless all code using it is recompiled. It is not always (or even often) the case that every package is rebuilt for every deployment - and often, there may be intermediate 3rd-party wrapper packages in the mix, meaning that it isn't even possible to simply rebuild it - you need to petition the 3rd party (who might be busy or otherwise unavailable) to update and redeploy their package - and there might even be multiple chained 3rd parties that need to update in the right order!

So, yeah. We cannot and will not just add optional parameters: instead, we need to add a new overload with the additional parameter. Note, however, that where appropriate it *is* possible to retire old overloads by simply removing the "this" modifier; this is *not* a binary break, and is not a source break as long as the remaining overloads over the existing and new use-cases. Most of the time, however, both the old and new overloads retain the "this".

I hope that explains why this is a breaking change *as written*.

mhov:
After using Dapper for years and years I finally needed to integrate cancellation tokens to discover it's a hot topic.  When's Dapper 3.x?

WolfyUK:
Could this be achieved with a `CommandDefinition` overload instead, then? e.g. (in `SqlMapper.Async.cs`):

```cs
public static Task<IEnumerable<TReturn>> QueryAsync<TReturn>(this IDbConnection cnn, CommandDefinition command, Type[] types, Func<object[], TReturn> map, string splitOn = "Id")
{
   return MultiMapAsync(cnn, command, types, map, splitOn);
}
```

lassevk:
I was going to ask for something similar for ExecuteAsync, which also lacks the CancellationToken parameter. Would a PR with an additional overload, not touching the existing one(s), be more welcome for that?

ExecuteAsync with parameters simply construct a CommandDefinition behind the scenes, with a `default` value for the cancellation token, it would be a simple matter to add an additional overload with this parameter lifted.

0xced:
I tried to implement what @mgravell suggested.

> So, yeah. We cannot and will not just add optional parameters: instead, we need to add a new overload with the additional parameter. Note, however, that where appropriate it is possible to retire old overloads by simply removing the "this" modifier.

It looks like it could work but I'm still currently fighting with the public API analyzers. Currently using `<NoWarn>RS0016;RS0017</NoWarn>` to get the project to compile. 😅

I think I should be able to open a new pull request soon to add _natural_ CancellationToken support to all Dapper methods.

**Edit:** I just pushed https://github.com/0xced/Dapper/tree/CancellationToken if anyone is interested.

0xced:
I have followed Marc's advice for preserving binary and source compatibility. I just opened #2178 which should supersede this pull request.

Issue #647, opened 2016-11-28
OrWhere() causing bug in usage of Where()


CrescentFresh:
[A prior issue](https://github.com/StackExchange/dapper-dot-net/issues/573) resulted in [this fix](https://github.com/StackExchange/dapper-dot-net/pull/577).

I guess no one bothered to check the original test case. It is still not "fixed":

    var sql = new SqlBuilder();
    var tpl = sql.AddTemplate("select * from foo /**where**/");

    sql.Where("MyField = @Param");
    sql.OrWhere("MyField2 = @Param2");

@rafakwolf expected: 

    select * from foo WHERE  ( MyField = @Param OR MyField2 = @Param2 ) 

However **both the previous and *fixed* SqlBuilder** produces the following:

    select * from foo WHERE MyField = @Param AND  ( MyField2 = @Param2 ) 

This is because only the first call to one of `Where()` or `OrWhere()` is the one that defines the `joiner` variable. That is, the variable that dictates what string to place *between* clauses. If `Where()` is called first the joiner will be "AND" but if `OrWhere()` is called first the joiner will be "OR".

Consider this: with the "fix" we get the following behavior (imagine these are in separate functions that don't know about each other):

    sql.Where("a = @a");

    sql.OrWhere("b = @b1");
    sql.OrWhere("b = @b2");

result:

    select * from foo WHERE a = @a AND (b = @b1 OR b = @b2)

But switch the order of the calls:

    sql.OrWhere("b = @b1");
    sql.OrWhere("b = @b2");

    sql.Where("a = @a");

result:

    select * from foo WHERE a = @a OR (b = @b1 OR b = @b2)

And that is a new bug introduced by the change. IMO under no circumstance should `.Where()` result in a clause getting OR'd into the query.

As a side comment, with the previous code it was at least consistent in that it always produced the first result irrespective of order. I believe the code change was made in haste and that the original behavior was correct, albeit confusing.

NickCraver:
Sorry I haven't gotten to this. I agree the old change was not the correct fix, and I don't plan on releasing the library to NuGet again until we fix this properly (possibly in a major version). I'll try and find time to revert that change for safety this weekend.

Can you open an issue for how this *should* behave, for all cases? There's certainly a big community using this far more than we are - let's figure out *what it should be*, even if that change requires a major version as a breaking change. Let's just get it right rather than trying to fix one symptom. That could definitely include extending SqlBuilder for other common cases people are hitting.

[...]

PR #2149, opened 2025-02-19
Fix and clarify OrWhere caveats in SqlBuilder docs

jnoordsij:
When looking into documentation of the `SqlBuilder` about combining `Where` and `OrWhere`, followed by reading up on it on #647, I noticed the behavior described in the readme differs from the behavior I'm currently observing (and which is also observed in #647).

This PR updates the documentation example to match with current observed behavior and also adds some additional wording for better readability.


mgravell (approved):

mgravell:
thanks, appreciated

PR #2156, opened 2025-04-06
OutputExpression wasn't working for methods that use QueryRowAsync

ri-rgb:
Added command.OnCompleted() to QueryRowAsync


mgravell (approved):

mgravell:
This looks like a good catch, thanks. I'll eyeball all the command paths tomorrow to see if we may have missed any others. Appreciated.

PR #2163, opened 2025-04-30
CI: serialize DB-dependent tests

kbth:
## Problem
Automated tests for the main branch are intermittently failing (though they sometimes succeed).
See: https://ci.appveyor.com/project/StackExchange/dapper

Example: 
```
  Failed SystemSqlClientParameterTests.TestTVPWithAnonymousEmptyObject [33 ms]
  Error Message:
   System.Data.SqlClient.SqlException : Could not find the type 'int_list_type'. Either it does not exist or you do not have the necessary permission.
```

## Cause

This is likely due to database-dependent tests being executed in parallel.

On December 22, 2024, AppVeyor updated their build image, which now includes the .NET 9 SDK:
https://www.appveyor.com/updates/2024/12/22/

Since Dapper targets multiple frameworks, this update caused tests to start running in parallel by default.

> For multi-targeted projects, tests are run for each targeted framework. The test host and the unit test framework are packaged as NuGet packages and are restored as ordinary dependencies for the project. Starting with the .NET 9 SDK, these tests are run in parallel by default.

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test?tabs=dotnet-test-with-vstest#description-1

## Solution

Following the official documentation, disable parallel execution of tests across TargetFrameworks.

> To disable parallel execution, set the TestTfmsInParallel MSBuild property to false.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,7 @@ jobs:
     - name: .NET Build
       run: dotnet build Build.csproj -c Release /p:CI=true
     - name: Dapper Tests
-      run: dotnet test tests/Dapper.Tests/Dapper.Tests.csproj -c Release --logger GitHubActions /p:CI=true
+      run: dotnet test tests/Dapper.Tests/Dapper.Tests.csproj -c Release --logger GitHubActions -p:CI=true -p:TestTfmsInParallel=false
       env:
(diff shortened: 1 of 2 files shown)


mgravell (approved):

mgravell:
CI is my personal nemesis. Appreciated.

PR #2167, opened 2025-05-13
This PR addresses issue #1573 by modifying the GetDeserializer logic to map nullable decimal values to 0m instead of null.

### LINKED ISSUE #1573: Just wonder reason of the casting?
`var dog = connection.Query<Dog>("select Age = @Age, Id = @Id", new { Age = (int?)null, Id = guid });`

Why do you need to cast `null `to `(int?)` here?

I can understand the reason for the casting for the below one, but yours.
```
Main()
{
	T.A(null);
}

public class T {
	public static void A(string x) => Console.WriteLine("str");
	public static void A(List<int> x) => Console.WriteLine("list");
}
```

AlexGreatDev:
- Added explicit check for `decimal?` types in GetDeserializer
- Returns 0m when encountering a NULL value
- Maintains existing behavior for other types

Closes #1573

### DIFF
--- a/Dapper/SqlMapper.cs
+++ b/Dapper/SqlMapper.cs
@@ -1973,6 +1973,11 @@ namespace Dapper
                 }
                 return GetTypeDeserializer(type, reader, startBound, length, returnNullIfFirstMissing);
             }
+            //  Patch for nullable decimal issue
+            if (type == typeof(decimal?))
+            {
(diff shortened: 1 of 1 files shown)


mgravell:
Returning zero for null seems ... just wrong; unless I'm mistaken (and that does happen), we don't usually null-coalesce like this; what is the reason / scenario here?

Issue #2166, opened 2025-05-13
[Optimization] Proposal: Span-based mapping for reduced allocations


AlexGreatDev:
## Problem
Current object mapping in Dapper uses reflection which causes memory overhead for bulk operations.

## Proposed Solution
Implement a `Span<T>`-based mapper as an alternative for scenarios like:
- High-volume data processing
- AOT compatibility

## Benchmark Expectations
Expected 20-30% reduction in memory allocations (based on prototype).

mgravell:
For the reflection discussion: https://aot.dapperlib.dev/gettingstarted

The underlying API that DapperAOT uses already supports span sources, however at the current time the normal usage piggy-backs off the regular Dapper API. I doubt the span usage Vs object is going to be relevant when compared to a DB fetch, but if you have a particular scenario in mind: let me know.

mgravell:
In particular: span usage and reflection are both salient points, but are orthogonal. It would be good to understand which your actual concern is.

mgravell:
Maybe we should take a step back, then, and be *very clear* about the scenario you're trying to describe. 

> Our actual concern is mostly the memory allocation from object materialization,

Right; let's discuss that. There are five main sources of allocations here:

1. metadata discovery and ref-emit; elided in AOT, and mostly amortized in "vanilla" via the strategy cache, so let's ignore that for now
2. raw ADO.NET overheads - largely unavoidable, but AOT has some features to allow optimization  (`[CacheCommand]`)
3. DTOs - might be avoidable by using `struct` DTOs in some cases
4. non-trivial fields, in particular `string` and `byte[]`; not simple to avoid
5. collection overheads, in particular `List<T>` array backers

So: which are you interested in? I assume we can skip 1 ("switch to AOT") and 2 ("enable command caching, then move on with life"); I assume for 3 you're already using `struct` DTOs, and there aren't easy options for 4.

Since you mention spans, can I assume your concern is actually 5? Is that what you're wanting to improve here? Note that if you use non-buffered mode, you can control how the buffering works there, but yes: there are definitely things we could do relatively easily to improve 5, so that we at least don't pay the resizing overhead - however, it is hard to remove the final array without having some API that exchanges a buffer lifetime.

There's also options in AOT to offer a row-count hint there (`[RowCountHint(...)]`), which can make things more efficient *today* for large sets, by starting the buffer with that size.

[...]

PR #2168, opened 2025-05-16
Use custom collector to avoid intermediate array allocations

mgravell:
Context #2166

`List<T>` uses initial-size plus doubling, which means that for large result sets you can get multiple arrays, for example for 90 rows you might have arrays with sizes 16, 32, 64, and 128 for 90, with the final result list using the oversized 128-element array, with 90 elements.

This change uses a custom collector for this part - same logic, but it uses the array-pool so that those intermediate arrays are recycled, and the final array is right-sized, i.e. a list with count 90 and capacity 90, and all intermediate arrays reused.

### DIFF
--- a/Dapper.StrongName/Dapper.StrongName.csproj
+++ b/Dapper.StrongName/Dapper.StrongName.csproj
@@ -19,10 +19,12 @@
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net461'">
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
+    <PackageReference Include="System.Memory" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
     <PackageReference Include="System.Reflection.Emit.Lightweight" />
+    <PackageReference Include="System.Memory" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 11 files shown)


TimothyMakkison:
Hey, glad to see someone's working on this. 

Have you considered using a stackallocated `scratchbuffer` for the inital buffer. This pattern is used in both [`ValueListBuilder`](https://github.com/dotnet/runtime/blob/1e7433e183629a797e771191d3b5327103ec6199/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/ValueListBuilder.cs#L11) and [`SegmentedArrayBuilder`](https://github.com/dotnet/runtime/blob/1e7433e183629a797e771191d3b5327103ec6199/src/libraries/System.Linq/src/System/Linq/SegmentedArrayBuilder.cs#L15), the latter of which is used internally for `Linq` for tasks that are very similar to dappers.

```C#
// similar to ValueListBuilder
var buffer = new Collector<T>([defaut, default, defaut, default, defaut, default, defaut, default]);

// similar to SegmentedArrayBuilder
Collector<TSource>.ScratchBuffer scratch = default;
Collector<TSource> builder = new Collector<T>(scratch);
```

mgravell:
Interesting suggestion. I'll find a moment to look; obviously only suitable for sync path, though.

mgravell:
Obviously it would also only apply in limited scenarios, i.e. the same rules as `unmanaged`

TimothyMakkison:
> Interesting suggestion. I'll find a moment to look; obviously only suitable for sync path, though.

Yeah, I thought that it might be suitable for functions that used `AsList`. I did think that there were areas where `ValueListBuilder` could be used in `Dapper` iirc relating to `DynamicParameters`. I can't imagine the performance gain being large.

PR #2174, opened 2025-07-06
performance: remove unneeded `ToArray` calls

TimothyMakkison:
Removes unnecessary `ToArray` calls

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -397,10 +397,10 @@ namespace Dapper
             }
             while (diving is not null);
 
-            var dynamicParamName = string.Concat(names.ToArray());
+            var dynamicParamName = string.Concat(names);
 
             // Before we get all emitty...
-            var lookup = string.Join("|", names.ToArray());
+            var lookup = string.Join("|", names);
 
             var cache = CachedOutputSetters<T>.Cache;
             var setter = (Action<object, DynamicParameters>?)cache[lookup];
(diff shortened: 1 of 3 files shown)


mgravell (approved):
Looks good; I'll need to double check impact on the full TFM range (i.e. the older overloads), but: 👍

Dapper/DynamicParameters.cs
```diff
@@ -397,10 +397,10 @@ static void ThrowInvalidChain()
             }
             while (diving is not null);
 
-            var dynamicParamName = string.Concat(names.ToArray());
+            var dynamicParamName = string.Concat(names);
 
             // Before we get all emitty...
-            var lookup = string.Join("|", names.ToArray());
+            var lookup = string.Join("|", names);
```
mkorsukov, line 403: A "char" vs "string" can be used in `Join()` call.

TimothyMakkison: I'm scared of getting nerd sniped here, but I'm pretty sure they do the same thing. `string.Join(char` and `string.Join(string`. both call `JoinCore`. The only difference is it converts the `char` into a `Span` via `new ReadOnlySpan<char>(in separator)` before calling `JoinCore`. [Code here](https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Private.CoreLib/src/System/String.Manipulation.cs#L721C13-L726C101)

I would love to know if the generated code is different between `char` and `string`

mkorsukov: You are right, they are similar. I was thinking of semantic use of "char" with the explicit visibility purpose of a-single-char-operation :)

mkorsukov: @mgravell With lots of pull requests in the queue, are we going to have a new Dapper release any time soon? And, if you or your team needs assistance, how can it be possible?

PR #2175, opened 2025-07-06
performance: skip `ToString` with interpolated string

TimothyMakkison:
Interpolated string skips the `Guid.ToString` allocation by calling `ISpanFormattable.TryFormat`

### DIFF
--- a/Dapper/DynamicParameters.cs
+++ b/Dapper/DynamicParameters.cs
@@ -407,7 +407,7 @@ namespace Dapper
             if (setter is not null) goto MAKECALLBACK;
 
             // Come on let's build a method, let's build it, let's build it now!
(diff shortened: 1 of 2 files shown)


mgravell (approved):
Yep, I'm very familiar with the joys of interpolated string internals (I was very recently proposing changes to /runtime in that area) - LGTM

TimothyMakkison:
> Yep, I'm very familiar with the joys of interpolated string internals (I was very recently proposing changes to /runtime in that area)

I wish the compiler were smart enough to do this by default for common types. I've done this exact same change so many times 😞 


OOC how much focus does Dapper place on performance? Looking through the repository I can definitely see a couple of changes to remove redundant strings ie `Database.Async`. IIRC there were some places where `ValueListBuilder` could be used, I doubt it would be a massive boost tho.

PR #2178, opened 2025-08-14
Add support for optional CancellationToken parameters

0xced:
This adds 28 new async methods with optional CancellationToken parameters support.

The methods without CancellationToken have been preserved in the `SqlMapper.Async.Legacy.cs` file for binary backward compatibility. The `this` modifier has been removed from those methods in order to also preserve source compatibility.

This pull request supersedes #2138


DontEatRice:
big YES!

PR #2180, opened 2025-09-26
Fix GridReader wiring: use IParameterCallbacks instead of DynamicParameters in QueryMultiple (sync + async)

### LINKED ISSUE #2179: Possible bug: `GridReader` constructor uses `DynamicParameters` cast instead of `IParameterCallbacks`
In the `GridReader` constructor, the parameter type is `IParameterCallbacks? callbacks`.  
But when creating a `GridReader`, the code does:
`command.Parameters as DynamicParameters` instead of: `command.Parameters as IParameterCallbacks`
This makes custom implementations of `IParameterCallbacks` unusable, even though the constructor was designed to support them.
Should this be changed to use `IParameterCallbacks` instead of `DynamicParameters`?

NandanDevHub:
This change fixes an issue where `QueryMultiple` and `QueryMultipleAsync` only invoked OnCompleted when parameters were of type DynamicParameters.
If a caller provided a custom parameter type that implemented `IParameterCallbacks`, the callbacks were ignored.

**Changes Below:**

- Updated both sync and async code paths to cast command.bParameters to `IParameterCallbacks` instead of `DynamicParameters`.

- Added unit tests to confirm that OnCompleted is now called for custom parameter objects in both sync and async scenarios.

**What i found locally is :**

- Existing behavior with `DynamicParameters` is unchanged.

- Custom implementations of `IParameterCallbacks` now work correctly.

Fixes #2179.

### DIFF
--- a/Dapper/SqlMapper.Async.cs
+++ b/Dapper/SqlMapper.Async.cs
@@ -1042,7 +1042,7 @@ namespace Dapper
                 cmd = command.TrySetupAsyncCommand(cnn, info.ParamReader);
                 reader = await ExecuteReaderWithFlagsFallbackAsync(cmd, wasClosed, CommandBehavior.SequentialAccess, command.CancellationToken).ConfigureAwait(false);
 
-                var result = new GridReader(cmd, reader, identity, command.Parameters as DynamicParameters, command.AddToCache, command.CancellationToken);
+                var result = new GridReader(cmd, reader, identity, command.Parameters as IParameterCallbacks, command.AddToCache, command.CancellationToken);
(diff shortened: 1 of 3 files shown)


NandanDevHub:
I added another small commit to adjust the new tests, reason being that AppVeyor CI doesn’t have SQL Server available, so the first version of the tests failed there.

With this update, the tests still run and pass locally when a database is present, but they no-op cleanly in CI without causing failures.

NandanDevHub:
Hi, just checking in to see if there’s any feedback or change you’d like here. I had eariler checked locally with SQL Server the fix works for both sync and async paths, i would love to make any small changes if needed.

PR #2194, opened 2026-02-24
Add user buffer for QueryAsync

Dimoner:
I suggest adding the ability to pass a custom buffer from user code. This would make it possible, in flows where we sequentially iterate over a large table and process records in chunks, to avoid reallocating the buffer. In my case, I need to traverse 5 million rows in batches of 100k when retrieving them from the database. I want to do this using a single array instead of allocating a new one each time.

Why is this not equivalent to a buffering flag?
Because the method returns IEnumerable rather than IAsyncEnumerable. As a result, the non-buffered mode relies on synchronous overloads. We should either change the design accordingly or provide an IAsyncEnumerable counterpart for the method without the buffering flag.

### DIFF
--- a/Dapper/CommandDefinition.cs
+++ b/Dapper/CommandDefinition.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Data;
 using System.Reflection;
 using System.Reflection.Emit;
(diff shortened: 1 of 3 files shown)


mgravell:
There already exists a non-bufferer pure async version: `QueryUnbufferedAsync` - does this make this redundant?

PR #2200, opened 2026-03-23
Fix/prefer typehandlers for enums

andreasblueher:
Hello @mgravell,

I know there have been already 2 pull requests (#458 & #1711) for the issue #259 which haven't been merged because users might rely on the current way dapper works with enums.

Since I'm using tons of enums, I created a suggestions to introduce a opt-in flag, which allows current users to stay on their behavior and others like me to get an updated version. Since 3.0 seems far, maybe this is something you could consider merging.

Best regards,
Andreas

### DIFF
--- a/Dapper/PublicAPI.Unshipped.txt
+++ b/Dapper/PublicAPI.Unshipped.txt
@@ -1,1 +1,3 @@
 ﻿#nullable enable
+static Dapper.SqlMapper.Settings.PreferTypeHandlersForEnums.get -> bool
+static Dapper.SqlMapper.Settings.PreferTypeHandlersForEnums.set -> void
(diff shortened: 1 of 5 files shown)


mgravell:
Yep, I need to find a chunk of time to chew through a ton of issues and PRs

alaricd:
The AppVeyor CI failure is caused by `global.json` requiring SDK `10.0.102`, but the AppVeyor image only has .NET SDKs up to 9.0.

To fix, update `global.json` to:
```json
{
  "sdk": {
    "version": "8.0.100",
    "rollForward": "latestMajor"
  }
}
```

The `rollForward: latestMajor` setting means it will pick up whatever is the highest installed SDK (9.0.306 on AppVeyor), so the build will work without needing an exact SDK match. This is a one-line change to the `version` field.

PR #2206, opened 2026-05-17
fix: dispose command when DbWrappedReader.DisposeAsync is called

### LINKED ISSUE #2185: Command not disposed with async using and ExecuteReaderAsync
Command not disposed in `DbWrappedReader `when using `DisposeAsync`.
=>
when using   `async using` with `ExecuteReaderAsync` command is not disposed

wyf027:
## Summary

- `DbWrappedReader.DisposeAsync` now disposes the wrapped `IDbCommand` after the reader, matching synchronous `Dispose` behavior.

## Motivation

Fixes #2185 — `async using` with `ExecuteReaderAsync` left the command undisposed because only the reader was disposed asynchronously.

## Test plan

- [x] `DbWrappedReader_DisposeAsync_DisposesCommand` unit test added
- [ ] `dotnet test tests/Dapper.Tests/Dapper.Tests.csproj --filter WrappedReaderTests`

### DIFF
--- a/Dapper/WrappedReader.cs
+++ b/Dapper/WrappedReader.cs
@@ -203,7 +203,24 @@ namespace Dapper
 #if NET5_0_OR_GREATER
         public override Task CloseAsync() => _reader.CloseAsync();
 
-        public override ValueTask DisposeAsync() => _reader.DisposeAsync();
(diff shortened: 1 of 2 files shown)


wyf027:

PR #2209, opened 2026-05-17
feat: QueryAsync with CommandDefinition and Type[] multi-map

### LINKED ISSUE #2193: No QueryAsync overload which accepts CommandDefinition, for cancellation, and Type array, for unbounded map
As per the title, there's no `QueryAsync` overload which accepts a `CommandDefinition` and a `Type[]` + `Func<object[], TReturn>` mapper, which means if you have > 7 types you want to map out of your query, and you want the query to be cancellable, you're out of luck. Not sure if this is intentional, but I do see the private method `MultiMapAsync<TReturn>`, which is called into by many of the existing overloads, does appear to support this variation. Any guidance on how to resolve would be greatly appreciated, thanks!

wyf027:
## Summary

Adds `QueryAsync<TReturn>(IDbConnection, CommandDefinition, Type[], Func<object[], TReturn>, splitOn)` so multi-map queries with more than seven types can use cancellation and other `CommandDefinition` options.

The private `MultiMapAsync` implementation already supported this path; the existing `string sql` overload constructed a `CommandDefinition` without exposing cancellation.

## Test plan

- [x] `TestMultiMapArbitraryMapsCommandDefinitionAsync` (10-type map via `CommandDefinition` + `CancellationToken`)
- [ ] CI (no local .NET SDK in contributor environment)

Fixes #2193

### DIFF
--- a/Dapper/PublicAPI.Unshipped.txt
+++ b/Dapper/PublicAPI.Unshipped.txt
@@ -1,3 +1,4 @@
 ﻿#nullable enable
 static Dapper.SqlMapper.Settings.PreferTypeHandlersForEnums.get -> bool
 static Dapper.SqlMapper.Settings.PreferTypeHandlersForEnums.set -> void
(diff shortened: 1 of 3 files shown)


wyf027:

PR #2210, opened 2026-05-17
feat: Query with CommandDefinition and Type[] multi-map

wyf027:
## Summary

Adds `Query<TReturn>(IDbConnection, CommandDefinition, Type[], Func<object[], TReturn>, splitOn)` for synchronous multi-map queries with more than seven types.

Forwards to the existing `MultiMapImpl` implementation (same pattern as the `string sql` overload).

Related to #2193; complements #2209 (async overload).

## Test plan

- [x] `TestMultiMapArbitraryMapsCommandDefinition`
- [ ] CI (no local .NET SDK in contributor environment)


wyf027:

PR #2217, opened 2026-07-15
Avoid double-enumerating IEnumerable<T> in AsTableValuedParameter

### LINKED ISSUE #2064: Multiple enumerations of IEnumerable<SqlDataRecord> passed as TVP
Looks like Dapper twice enumerates IEnumerable<SqlDataRecord> passed as TVP
```
var numbers = GenerateNumbers(3);
BulkInsert(numbers);

static IEnumerable<int> GenerateNumbers(int count)
{
    for (int i = 0; i < count; i++)
    {
        Console.WriteLine(i);
        yield return i;
    }
}

void BulkInsert(IEnumerable<int> numbers)
{
    var records = numbers.Select(x => MapToSqlDataRecord(x));
    var sql = @"INSERT INTO [Test] (Id) SELECT Id FROM @tvp;";

    using (var connection = new SqlConnection(connectionString))
    {
        connection.Execute(sql, new { tvp = records.AsTableValuedParameter("dbo.TestType") });
    }
}

static SqlDataRecord MapToSqlDataRecord(int i)
{
    var record = new SqlDataRecord(
        new SqlMetaData("Id", SqlDbType.Int)
    );
    record.SetInt32(0, i);
    return record;
}
```
this code writes to console 
0 
0 
1 
2

and if I pass enumerable from database it throws exception because can not open second reader on same connection (same enumerable works fine using ADO.Net)

image
I believe this is the first enumeration

hexonal:
`SqlDataRecordListTVPParameter<T>.Set` used `data.Any()` to swap an empty sequence for `null`, which forced an extra enumeration of the source before the caller ever executed the command. For single-pass sources (open readers, streaming iterators, etc.) that meant the real enumeration during command execution either restarted from scratch or failed outright with something like "There is already an open DataReader...".

This now only short-circuits to `null` when the source is a known `IReadOnlyCollection<T>`, checked via `Count` as @mgravell suggested in the issue thread. Everything else flows straight through untouched, so it's enumerated exactly once, by whoever actually executes the command.

Added three tests: one mirrors the original repro end-to-end (needs SQL Server to run), and two run offline against `SqlDataRecordListTVPParameter<T>.Set` directly, checking that a non-collection source is never touched and a known-empty collection still nulls out.

Fixes #2064

### DIFF
--- a/Dapper/SqlDataRecordListTVPParameter.cs
+++ b/Dapper/SqlDataRecordListTVPParameter.cs
@@ -2,7 +2,6 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Data;
-using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
 
@@ -37,7 +36,15 @@ namespace Dapper
 
         internal static void Set(IDbDataParameter parameter, IEnumerable<T>? data, string? typeName)
         {
-            parameter.Value = data is not null && data.Any() ? data : null;
+            // don't enumerate "data" here - it may be a single-pass source (an open reader, a
+            // streaming iterator, etc); only short-circuit to null when we can tell for free
+            // that it is empty, since providers reject an empty TVP enumerable
+            parameter.Value = data switch
+            {
+                null => null,
+                IReadOnlyCollection<T> { Count: 0 } => null,
+                _ => data,
+            };
             StructuredHelper.ConfigureTVP(parameter, typeName);
         }
     }
--- a/tests/Dapper.Tests/ParameterTests.cs
+++ b/tests/Dapper.Tests/ParameterTests.cs
@@ -131,6 +131,25 @@ namespace Dapper.Tests
             return number_list;
         }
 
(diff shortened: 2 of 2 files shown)


hexonal:
@mgravell — a gentle bump on this one; it has had no eyes on it since it was opened a month ago, and it is small.

Short version: `SqlDataRecordListTVPParameter<T>.Set` calls `data.Any()` to swap an empty sequence for `null`, which enumerates the source once before the caller ever executes the command. For single-pass sources that is fatal — an open reader either restarts from scratch or throws "There is already an open DataReader…" (#2064). The fix only short-circuits when the source is a known `IReadOnlyCollection<T>`, checked via `Count`, which is the approach you suggested in that issue thread; anything else flows through untouched and is enumerated exactly once, by whoever executes the command.

Three tests: one end-to-end repro (needs SQL Server), and two offline ones against `Set` directly — a non-collection source is never touched, a known-empty collection still nulls out. Branch is current and CI is green.

Happy to change the shape if you would rather it were done differently.

PR #2233, opened 2026-09-14
Suppress trim and AOT diagnostics in Dapper fallback paths

### LINKED ISSUE #2232: Dapper.AOT leaves trim-unsafe Dapper reflection paths reachable during Native AOT publish
**Check your library version, and try updating**

This reproduces with Dapper `2.1.86` and Dapper.AOT `1.1.0`, which are the versions currently used by the Aspire Native AOT dashboard work.

**Describe the bug**

Publishing a Dapper.AOT-enabled application with Native AOT produces trimming and AOT diagnostics from reachable code in the Dapper assembly. The application uses `[module: DapperAot]`, includes `Dapper.AOT` at runtime for generated interceptors, and uses `DynamicParameters` populated with explicit scalar values. It does not use object parameter templates or SQL Server table-valued parameters.

The publish reports 16 distinct Dapper diagnostics. Here is the complete breakdown by warning ID and code path.

#### `IL2070`: unannotated `Type` parameters used for reflection (5 diagnostics)

1. `DefaultTypeMap.GetSettableProps(Type)` calls `Type.GetProperties` with public and non-public binding flags. Its `t` parameter does not promise that those properties are preserved.
2. `DefaultTypeMap.GetSettableFields(Type)` calls `Type.GetFields` with public and non-public binding flags. Its `t` parameter does not promise that those fields are preserved.
3. `StructuredHelper.CreateFor(Type, string, int)` calls `Type.GetProperty` for two SQL Server data-record properties. Its `type` parameter does not promise that public properties are preserved. These two call sites produce separate diagnostics at `SqlDataRecordListTVPParameter.cs:69` and `:75`.
4. `TypeExtensions.GetPublicInstanceMethod(Type, string, Type[])` calls `Type.GetMethod`. Its `type` parameter does not promise that public methods are preserved.

The common issue is missing `DynamicallyAccessedMembers` data flow. Annotating the incoming `Type` values may be appropriate where these reflection paths are intended to work after trimming. If the paths are unsupported under AOT, they could instead be isolated behind an appropriately annotated public boundary and kept unreachable from Dapper.AOT-generated paths.

```text
/_/Dapper/TypeExtensions.cs(9): Trim analysis warning IL2070: Dapper.TypeExtensions.GetPublicInstanceMethod(Type,String,Type[]): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to System.Type.GetMethod(...). The parameter 'type' does not have matching annotations.
```

#### `IL2075`: reflected types lose member-preservation requirements (4 diagnostics)

1. `DefaultTypeMap.GetPropertySetter(PropertyInfo, Type)` reflects over `PropertyInfo.DeclaringType`. That return value does not guarantee public and non-public properties are preserved.
2. `SqlMapper.CreateParamInfoGenerator(...)` calls `GetProperties()` on `SqlMapper.Identity.ParametersType`, but that property does not preserve public properties.
3. The same method calls `GetConstructors()` on `Identity.ParametersType`, but the value does not preserve public constructors.
4. The same method calls `GetMethod(...)` on `PropertyInfo.PropertyType`, but that value does not preserve public methods.

These warnings show that adding annotations only to local helper parameters may not be enough. The requirements need to flow through `Identity.ParametersType` and the property-type path, or the runtime parameter-object generator needs to be excluded when a generated Dapper.AOT interceptor handles the command.

```text
/_/Dapper/SqlMapper.cs(2598): Trim analysis warning IL2075: Dapper.SqlMapper.CreateParamInfoGenerator(...): the return value of SqlMapper.Identity.ParametersType does not satisfy the PublicProperties requirement in System.Type.GetProperties().
```

#### `IL3050`: runtime code generation and generic construction (7 diagnostics)

1. `StructuredHelper.CreateFor(...)` creates a `DynamicMethod` for SQL Server table-valued parameters.
2. `SqlMapper.CreateParamInfoGenerator(...)` creates a `DynamicMethod` for runtime parameter-object access.
3. `SqlMapper.CreateParamInfoGenerator(...)` also calls `Type.MakeGenericType`.
4. `SqlMapper.AddTypeHandlerCore(...)` calls `Type.MakeGenericType` at three separate sites while constructing nullable/value-type handler adapters.
5. `SqlMapper.LookupDbType(...)` calls `Type.MakeGenericType` while resolving a data-record handler.

`DynamicMethod` cannot work in Native AOT. `MakeGenericType` works only when the required closed generic instantiation was generated ahead of time. In this repro, the application does not use object parameter templates or SQL Server TVPs, so ideally the generated Dapper.AOT path would prevent these fallback implementations from being rooted. Any still-supported runtime path needs an AOT-safe implementation or explicit generic instantiation coverage.

```text
/_/Dapper/SqlMapper.cs(2573): AOT analysis warning IL3050: Dapper.SqlMapper.CreateParamInfoGenerator(...): using DynamicMethod can break functionality when AOT compiling. Creating a DynamicMethod requires dynamic code.
```

**To Reproduce**

1. Check out https://github.com/microsoft/aspire/pull/19565 after its warning-visibility change is included.
2. From the Aspire repository root, run:

```powershell
.\restore.cmd
.\.dotnet\dotnet.exe publish src\Aspire.Dashboard\Aspire.Dashboard.csproj `
  -c Release -r win-x64 --self-contained true `
  -p:ContinuousIntegrationBuild=false
```

3. Search the output for `/_/Dapper/` or `Dapper.SqlMapper`.

The relevant project enables:

```xml
<PublishAot>true</PublishAot>
<TrimmerSingleWarn>false</TrimmerSingleWarn>
<InterceptorsNamespaces>$(InterceptorsNamespaces);Dapper.AOT</InterceptorsNamespaces>
```

The Dapper.AOT opt-in is in `src/Aspire.Dashboard/ServiceClient/DashboardSqliteDatabase.cs`. The dashboard's `DynamicParameters` usages are under `src/Aspire.Dashboard/Otlp/Storage/`.

**Expected and actual behavior**

Expected: supported Dapper.AOT usage publishes without diagnostics from Dapper's reflection/Reflection.Emit fallback implementation. Paths unsupported under Native AOT should either be removed from the reachable graph, annotated so the warnings stop at the appropriate public boundary, or replaced by AOT-safe paths.

Actual: reflection and dynamic-code implementation details remain reachable and produce `IL2070`, `IL2075`, and `IL3050` warnings.

**Additional context**

The application uses SQLite through `Microsoft.Data.Sqlite`. The warnings include SQL Server TVP helper code even though that feature is not used, which suggests that more fallback implementation is retained than this usage requires.

Environment:

- Windows 10.0.26200, `win-x64`
- .NET SDK `11.0.100-rc.1.26425.128` (`3551975be0`)
- Dapper `2.1.86`
- Dapper.AOT `1.1.0`
- Aspire commit `ca26d2c4b5285508e07ee1b218ba0b36582d0e3e`

I searched open and closed issues for Dapper.AOT trimming warnings and these specific members and did not find a matching report.

JamesNK:
Fixes #2232.

## Summary

Add method-level `UnconditionalSuppressMessage` attributes for the reflection, runtime generic construction, and `DynamicMethod` paths reported by the Dapper.AOT-enabled Native AOT dashboard application:

- `DefaultTypeMap` property/field discovery and inherited property setter lookup (`IL2070`, `IL2075`).
- `StructuredHelper.CreateFor` provider-specific structured parameters (`IL2070`, `IL3050`).
- `SqlMapper.AddTypeHandlerCore` and `LookupDbType` runtime generic construction (`IL3050`).
- `SqlMapper.CreateParamInfoGenerator` runtime parameter-object generation (`IL2075`, `IL3050`).
- `TypeExtensions.GetPublicInstanceMethod` framework method lookup (`IL2070`).

Includes an internal compatibility definition of `UnconditionalSuppressMessageAttribute` for the legacy target frameworks. No public signatures or execution paths change.

This is a diagnostic-suppression change, not an AOT-safe rewrite: it does not preserve reflected members, root arbitrary generic instantiations, or remove the fallback implementations from the reachable graph. Unsupported fallback usage remains unsupported.

## Validation

- `dotnet build .\Dapper\Dapper.csproj --no-restore` succeeded for `net461`, `netstandard2.0`, `net8.0`, and `net10.0`.
- The standalone Dapper.AOT/SQLite sample below published for `win-x64` with Native AOT and **zero Dapper IL diagnostics**. The check included all Dapper `IL` warning IDs, not just the three reported in the issue.
- Running the published native executable printed `42`.
- SDK used: `11.0.100-rc.1.26425.128`; sample target framework: `net10.0`.
- The sample restore/publish also reported `NU1903` for transitive `SQLitePCLRaw.lib.e_sqlite3` 2.1.11 ([advisory](https://github.com/advisories/GHSA-2m69-gcr7-jv3q)). This was separate from the Dapper IL diagnostics; these are the exact package versions tested, not a dependency-version recommendation.
- Full test suite attempted with `dotnet test .\tests\Dapper.Tests\Dapper.Tests.csproj --no-restore --framework net10.0`: **66 passed, 30 skipped, 704 failed**. SQL Server connection failures were reported because the required database instance was unavailable, so the full suite is not validated as passing.
- No source diagnostics were reported for the changed files. CRLF-aware `git diff --check` passed.

The standalone sample was created outside the repository and is included below rather than added to the solution. It validates the scalar `DynamicParameters` scenario; the full Aspire dashboard was not republished for this patch.

### DapperIssue2232.csproj

This is the project used for validation. Adjust the absolute `ProjectReference` to point to the Dapper checkout being tested.

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <PublishAot>true</PublishAot>
    <TrimmerSingleWarn>false</TrimmerSingleWarn>
    <InterceptorsNamespaces>$(InterceptorsNamespaces);Dapper.AOT</InterceptorsNamespaces>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="C:\Development\Source\Dapper\Dapper\Dapper.csproj" />
    <PackageReference Include="Dapper.AOT" Version="1.1.0" />
    <PackageReference Include="Microsoft.Data.Sqlite" Version="10.0.8" />
  </ItemGroup>
</Project>
```

### Program.cs

```csharp
using System;
using Dapper;
using Microsoft.Data.Sqlite;

[module: DapperAot]

using var connection = new SqliteConnection("Data Source=:memory:");
connection.Open();

var parameters = new DynamicParameters();
parameters.Add("Value", 42);
Console.WriteLine(connection.QuerySingle<int>("select @Value", parameters));
```

### Publish And Run

Run from the standalone sample directory with the Windows Native AOT build prerequisites installed:

```powershell
dotnet publish .\DapperIssue2232.csproj -c Release -r win-x64 --self-contained true -p:ContinuousIntegrationBuild=false
.\bin\Release\net10.0\win-x64\publish\DapperIssue2232.exe
```

Observed native executable output:

```text
42
```


JamesNK:
Vibe coded. Please double check this is good. I'm not very fimilar about how Dapper and Dapper.AOT interact.

mgravell (approved):

